PR #37, opened 2024-12-10
Enable Pre-Commit CI

srilman:

Enable Pre-Commit CI to auto-check and fix PRs using our pre-commit config. Config options: https://pre-commit.ci/#configuration.

* It also supports updating the pre-commit file, and I set it to once every month. We can also set it to weekly (too often IMO) or quarterly (too slow IMO). 
* Looks like the Java and Kotlin pre-commits require the Java executable installed, which the CI doesn't install automatically. I think there's an alternative we should use instead, but for now will skip.

## Testing strategy

Testing on this PR if the Pre-Commit CI runs.

## User facing changes

N/A, maintenance PR.

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,7 @@
 # Pre-Commit CI Config
 ci:
   autoupdate_schedule: monthly
+  skip: [pretty-format-java, pretty-format-kotlin]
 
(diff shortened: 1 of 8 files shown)


hadia206 (approved):
Nice. Thanks Srinivas

scott-routledge2 (approved):

PR #41, opened 2024-12-10
Update README.md with description

marquisdepolis:
Updated documentation for minor clarifications and description.


srilman (approved):

srilman (changes requested):
Actually could you mention specifically Linux x86 and both Mac x86 and Mac ARM. Just to make sure.

srilman (approved):

PR #46, opened 2024-12-11
[BSE-4203] Use Pixi for Pip Build Deps

srilman:

Replace the conda-lock usage in our Pip pipelines with Pixi.

We use conda-lock to set up a reproducible environment for our C++ dependencies. However, its not really reproducible because we construct the lockfile during the build process. Furthermore, we install more dependencies than we need for building. We do all this because conda-lock doesn't have the features we need.

Pixi has features like multiple environments that we can use instead. Furthermore, with Pixi, we can only install the C++ dependencies during the Pip build.

Some other changes:
- Fix a bug with the versioning and the data folder in the Pip package
- Use the ManyLinux 2.28 image
- Build MPI4Py from Source and Vendor in Package

Note that we unfortunately need to still use the C++ compilers from Pixi / Conda. Ideally, we should use the compiler provided by cibuildwheel for best compatibility, but we can't until we find other sources for our C++ dependencies. Most are already provided (libcurl on the system, AWS SDK from PyArrow, Boost through CMake build), but HDF5 is the hard one. 

## Testing strategy

Running the Pip CIs.

## User facing changes

Ideally none.

## Checklist


pyproject.toml
```diff
@@ -172,22 +176,22 @@ exclude = [
 # Disable building PyPy wheels on all platforms
 # Disable building wheels for musllinux
 skip = ["pp*", "*musllinux*", "*i686*"]
-manylinux-x86_64-image = "427443013497.dkr.ecr.us-east-2.amazonaws.com/cibuildwheel-manylinux:latest"
 # Increase pip debugging output
 build-verbosity = 1
+build-frontend = { name = "pip" }
+#, args = ["--only-binary", "mpi4py", "--extra-index-url", "https://pypi.anaconda.org/mpi4py/simple"] }
```
srilman, line 182: TODO: Remove when addressing comments

.github/workflows/release.yml
```diff
@@ -7,11 +7,11 @@ on:
...
-  pull_request:
-    paths:
-      - 'pixi.lock'
-      - '.github/workflows/release.yml'
-      - '.github/workflows/_build_*.yml'
+  # pull_request:
+  #   paths:
+  #     - 'pixi.lock'
+  #     - '.github/workflows/release.yml'
+  #     - '.github/workflows/_build_*.yml'
```
srilman, line 14: TODO: Uncomment while addressing comments

.github/workflows/build_wheels.yml
```diff
@@ -205,7 +231,7 @@ jobs:
       # TODO: Chain with E2E tests for effective testing
 
   upload-all:
-    needs: [build-iceberg, test-iceberg-import, build_bodo_linux_wheels, build_bodo_macos_wheels, test-bodo-linux, build_bodosql_wheels, test-bodosql]
+    needs: [build-iceberg, test-iceberg-import, build_bodo_linux_wheels, test-bodo-linux, build_bodosql_wheels, test-bodosql]
```
srilman, line 234: TODO: Undo while addressing comments

.github/workflows/build_wheels.yml
```diff
@@ -36,15 +58,16 @@ jobs:
...
     needs: build_bodo_linux_wheels
     steps:
       - uses: actions/download-artifact@v4
         id: download-artifact
         with:
           pattern: cibw-wheels-ubuntu-*
           path: .
+      - uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python_version }}
```
srilman, line 70: I did this because I couldn't add `set -exo pipefail` inside the docker container. Thus, the job kept passing even though the test failed. This way, the job fails is the test fails

.github/workflows/build_wheels.yml
```diff
@@ -9,24 +9,46 @@ on:
...
+          # we need to fetch the entire history
+          fetch-depth: 0
+      - name: Get Version
+        id: get_version
+        run: |
+          set -exo pipefail
+          pip install setuptools_scm
+          echo "bodo_version=`python -m setuptools_scm`" >> $GITHUB_OUTPUT
+    outputs:
+      bodo_version: ${{ steps.get_version.outputs.bodo_version }}
```
srilman, line 32: We ran into a problem where cibuildwheel was using the wrong version during build. This fixes that by fetching the version and fixing it beforehand

.github/workflows/_build_bodo_pip.yml
```diff
@@ -30,68 +34,97 @@ jobs:
...
+            CONDA_PREFIX=$(pwd)/.pixi/envs/pip-cpp
+            SETUPTOOLS_SCM_PRETEND_VERSION=${{ inputs.bodo_version }}
+          CIBW_ENVIRONMENT_LINUX: >
+            PATH=/project/.pixi/envs/pip-cpp/bin:$HOME/.pixi/bin:$PATH
+            LD_LIBRARY_PATH=/project/.pixi/envs/pip-cpp/lib
+            CFLAGS="-isystem /project/.pixi/envs/pip-cpp/include"
+            CPPFLAGS="-isystem /project/.pixi/envs/pip-cpp/include"
+            CC=/project/.pixi/envs/pip-cpp/bin/x86_64-conda-linux-gnu-gcc
+            CXX=/project/.pixi/envs/pip-cpp/bin/x86_64-conda-linux-gnu-g++
+            DISABLE_CCACHE=1
```
srilman, line 97: CIBW is annoying and wont combine the envs from `CIBW_ENVIRONMENT` and `CIBW_ENVIRONMENT_LINUX` on Linux, so I had to copy them here.

Plus, Pixi shell activation wasn't working inside of the docker container, so I had to manually set envs to find the compiler correctly.

.github/workflows/build_wheels.yml
```diff
@@ -151,6 +176,7 @@ jobs:
           pip install bodo bodosql
       - name: Test
         run: |
+          set -exo pipefail
```
IsaacWarren, line 179:
```suggestion
          set -eo pipefail
```
Had issues with this printing tokens

.github/workflows/build_wheels.yml
```diff
@@ -53,6 +76,7 @@ jobs:
           pip install bodo
       - name: Test
         run: |
+          set -exo pipefail
```
IsaacWarren, line 79:
```suggestion
          set -eo pipefail
```
Had issues with this printing tokens

IsaacWarren (approved):
Thanks Srinivas this is great!

buildscripts/bodo/pip/macos/patch_libs_for_pip.py
```diff
@@ -34,8 +34,10 @@ def patch_lib(fpath):
     #     @rpath/libarrow.500.dylib (compatibility version 500.0.0, current version 500.0.0)
     #     @rpath/libc++.1.dylib (compatibility version 1.0.0, current version 1.0.0)
     #     /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1311.0.0)
+    print("Loaded Libs for", fpath)
```
hadia206, line 37: do we want to keep prints in this file?

hadia206 (approved):
Great job @srilman 
Thanks

buildscripts/bodo/pip/macos/patch_libs_for_pip.py
srilman: Yes for sure, it was useful for debugging

srilman: And it doesn't cause any issues

.github/workflows/build_wheels.yml
srilman: I'm actually fixing that in a separate PR. Its a different issue, so we can keep this

.github/workflows/build_wheels.yml
srilman: See above

PR #47, opened 2024-12-11
BSE-4340: Update spark benchmark

IsaacWarren:
Updates spark benchmark to use pandas on spark to be more in line with other benchmarks.

## Testing strategy

Tested on EMR

## User facing changes

NA

## Checklist
CI is unnecessary for this PR


benchmarks/spark/spark_nyc_taxi_preciptation.py
```diff
@@ -0,0 +1,115 @@
...
+        SparkSession.builder.appName("MonthlyTravelsWeather")
+        .config("spark.jars.packages", "org.apache.hadoop:hadoop-aws:3.3.2")
+        .config("spark.sql.execution.arrow.pyspark.enabled", "true")
+        .config(
+            "fs.s3a.aws.credentials.provider",
+            "org.apache.hadoop.fs.s3a.AnonymousAWSCredentialsProvider",
+        )
+        .getOrCreate()
+    )
+    import pyspark.pandas as ps
```
ehsantn, line 24: Any problem with moving the import to the top?

benchmarks/spark/spark_nyc_taxi_preciptation.py
```diff
@@ -0,0 +1,115 @@
...
+    # Convert datetime columns and create necessary features
+    fhvhv_tripdata = (
+        (
+            fhvhv_tripdata.withColumn("date", to_date(col("pickup_datetime")))
+            .withColumn("month", month(col("pickup_datetime")))
+            .withColumn("hour", hour(col("pickup_datetime")))
+            .withColumn(
+                "weekday", dayofweek(col("pickup_datetime")).isin([2, 3, 4, 5, 6])
+            )
+            # pandas-on-Spark doesn't like these datetime columns which is why we use spark apis for the read and this conversion
```
ehsantn, line 51:
```suggestion
            # pandas-on-Spark doesn't like these datetime columns which is why we use Spark APIs for the read and this conversion
```

ehsantn (approved):
Looks good to me. Thanks @IsaacWarren .

benchmarks/spark/spark_nyc_taxi_preciptation.py
```diff
@@ -0,0 +1,115 @@
...
+    ).rename(columns={"DATE": "date", "PRCP": "precipitation"})
+
+    central_park_weather_observations["date"] = ps.to_datetime(
+        central_park_weather_observations["date"]
+    )
+
+    # Read in trip data using spark
+    fhvhv_tripdata = spark.read.parquet(
+        "s3a://bodo-example-data/nyc-taxi/fhvhv_tripdata_rewrite/"
+    ).drop("__index_level_0__")
```
scott-routledge2, line 40: Why do we need to drop the index here, does this have any performance implications?

scott-routledge2 (approved):
LGTM, thanks @IsaacWarren

benchmarks/spark/spark_nyc_taxi_preciptation.py
IsaacWarren: It's because we read from a rewritten data set that has this index included because we wrote it and that's what pandas does. pandas on spark throws an error if there's a column with this name when creating a dataframe

benchmarks/spark/spark_nyc_taxi_preciptation.py
IsaacWarren: Nope, moved

PR #51, opened 2024-12-11
Fix BodoSQL use in Jupyter (without JIT)

ehsantn:

Fixes BodoSQL use inside Jupyter without JIT (spawn mode). Updates the quickstart example to not use a JIT function.

## Testing strategy

Tested it manually since we don't have unit test infrastructure for Jupyter yet.

## Checklist

### DIFF
--- a/BodoSQL/bodosql/context.py
+++ b/BodoSQL/bodosql/context.py
@@ -1104,6 +1104,9 @@ class BodoSQLContext:
             # are not visible to cloudpickle as used by the function. See:
             # test_json_fns.py::test_object_construct_keep_null[no_nested-no_null-with_case]
             if isinstance(dispatcher, SubmitDispatcher):
+                # __builtins__ which is added to glbls by exec causes issues in
+                # Jupyter/IPython pickling.
+                glbls.pop("__builtins__", None)
                 dispatcher.add_extra_globals(glbls)
 
             return dispatcher(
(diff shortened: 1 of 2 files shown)


BodoSQL/bodosql/context.py
```diff
@@ -1104,6 +1104,9 @@ def sql(self, sql, params_dict=None, dynamic_params_list=None, **jit_options):
             # are not visible to cloudpickle as used by the function. See:
             # test_json_fns.py::test_object_construct_keep_null[no_nested-no_null-with_case]
             if isinstance(dispatcher, SubmitDispatcher):
+                # __builtins__ which is added to glbls by exec causes issues in
+                # Jupyter/IPython pickling.
+                glbls.pop("__builtins__", None)
```
IsaacWarren, line 1109: Don't we need this to use builtin functions like len though?

IsaacWarren (approved):
One question otherwise LGTm, thanks Ehsan

BodoSQL/bodosql/context.py
ehsantn: Builtin functions are the same on the worker and are available in regular globals. We don't need to pickle and send them.

IsaacWarren: Makes sense, thanks

scott-routledge2 (approved):

PR #53, opened 2024-12-12
Avoid importing BodoSQL unnecessarily

ehsantn:

This avoids importing BodoSQL unnecessarily which can cause issues issues if BodoSQL is not installed properly (e.g. Java issues). Also reduces compilation time a bit.

## Testing strategy

Tested it manually to make sure bodosql isn't imported.

## Checklist

### DIFF
--- a/bodo/libs/distributed_api.py
+++ b/bodo/libs/distributed_api.py
@@ -74,6 +74,7 @@ from bodo.utils.typing import (
     ExternalFunctionErrorChecked,
     MetaType,
     decode_if_dict_array,
+    is_bodosql_context_type,
     is_overload_false,
     is_overload_none,
     is_str_arr_type,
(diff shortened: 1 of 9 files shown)


IsaacWarren (approved):
Thanks Ehsan, can we add a simple test that compiles a simple function and asserts bodosql isn't imported just to make sure this doesn't regress

ehsantn:
My concern is that our testing infrastructure itself may import BodoSQL and we don't have control over it.

IsaacWarren:
> My concern is that our testing infrastructure itself may import BodoSQL and we don't have control over it.

We could do it in a subprocess, this seems important and like it could easily be readded by mistake

scott-routledge2 (approved):
Thanks @ehsantn

ehsantn:
> > My concern is that our testing infrastructure itself may import BodoSQL and we don't have control over it.
> 
> We could do it in a subprocess, this seems important and like it could easily be re-added by mistake

Looks like only one test was importing BodoSQL so I just updated it.

PR #55, opened 2024-12-12
DockerHub Image Workflow

hadia206:

Add workflow to create and upload Bodo docker image to DockerHub with each new release.

## Testing strategy
https://github.com/bodo-ai/Bodo/actions/runs/12420811976
Tested uploaded image locally 
`docker pull bodoai/bodoai_image:latest`
`docker run -p 8888:8888 --cpus=8 bodoai/bodoai_image`

## Checklist


README.md
```diff
@@ -59,6 +59,8 @@ See Bodo documentation to learn more: https://docs.bodo.ai/
 
 ## Installation
 
+Note: Bodo requires Python 3.10, 3.11, or 3.12.
```
hadia206, line 62: a note in main repo README

buildscripts/docker/docker_release/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+FROM ubuntu:latest
+
+USER root
+# Copy the conda environment from the builder stage
+COPY --chown=root --from=builder /opt/conda /opt/conda
+
+# Install necessary packages
+RUN apt-get update && \
+    apt-get install -y bzip2 git make python3 python3-pip vim wget && \
```
IsaacWarren, line 21: Why do we need these packages? Particularly python, shoudn't we be using conda's python?

buildscripts/docker/docker_release/pi.ipynb
IsaacWarren, line 1: We already have these in examples, can we just copy from there instead of duplicating

IsaacWarren (commented):
Thanks Hadia

.github/workflows/docker_build_and_publish.yml
```diff
@@ -0,0 +1,93 @@
...
+
+        # Step 4: Stop and remove the container after the test
+        docker rm -f test-container
+
+
+    - name: Upload Notebook Outputs
+      uses: actions/upload-artifact@v4
+      with:
+        name: notebook-outputs
+        path: ./notebooks/
```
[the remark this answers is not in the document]
hadia206: What do you mean by automated test?

buildscripts/docker/docker_release/Dockerfile
hadia206: Not sure, This was in the PR I took over. 
I only added `vim` as I needed that when I was testing.

.github/workflows/docker_build_and_publish.yml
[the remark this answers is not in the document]
IsaacWarren: Checking the output, We're never going to look at the notebook-outpus to make sure they're correct

buildscripts/docker/docker_release/pi.ipynb
hadia206: I thought about that but it complicated the README, build image locally step. So felt this in the same folder make it easier to the user

buildscripts/docker/docker_release/Dockerfile
IsaacWarren: Ok I think we need to at least try removing them, we want our images to be as minimal as possible

buildscripts/docker/docker_release/pi.ipynb
IsaacWarren: Symlink then?

.github/workflows/docker_build_and_publish.yml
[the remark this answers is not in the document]
hadia206: Oh. I'll add a comment in the file.
I only test that the notebook can run, not really care about output data itself since this should be caught by other unit or e2e tests.
This is mainly to have access to the notebook-output in case the actual notebook run failed.
I can also remove this step if it's confusing.

buildscripts/docker/docker_release/env.yaml
```diff
@@ -0,0 +1,11 @@
+name: base
+channels:
+  - conda-forge
+  - bodo.ai
+dependencies:
+  - python=3.12
+  - bodo
+  - bodosql
```
srilman, line 8: Should be include everything, include the Iceberg connector and the Azure SAS token

buildscripts/docker/docker_release/README.md
```diff
@@ -0,0 +1,75 @@
...
+## Prerequisites
+Before you begin, ensure the following:
+
+1. The latest version of Docker is installed on your machine.
+2. The Docker daemon is running (e.g., via Docker Desktop).
+
+## Download and Run the Container
+
+The Bodo image is available on DockerHub:
+https://hub.docker.com/r/bodoai/bodoai_image
```
srilman, line 22: Can we discuss the name a bit? I think a more descriptive name might be more benefitial, like `bodoai/jupyterlab`. Cause the name `bodoai_image` is a bit repetitive and unnecessary. Plus, we may want to have more image types in the future, like a minimal image for example without jupyter or a perf image with better MPI, etc.

buildscripts/docker/docker_release/env.yaml
srilman: Also, should we fix the version somehow to be safe. Like lets say we accidentally publish one release after another, then the Docker build will get a later version of bodo

.github/workflows/release.yml
```diff
@@ -70,4 +70,11 @@ jobs:
       is-release: ${{ github.event_name == 'release' }}
     secrets: inherit
 
+  docker-img:
+    needs: bodo-conda-linux
```
srilman, line 74: I think we should have it need bodosql, iceberg, and azure as well

srilman (commented):
Looks good so far, just a couple of notes

buildscripts/docker/docker_release/env.yaml
hadia206: so this is part of the release process so it should always get water was released.
I don't think we should fix version.

buildscripts/docker/docker_release/README.md
hadia206: I'm ok. with changing name. Let me know what it should be.
I don't think it should be `jupyterlab` though as it can be used without it.

buildscripts/docker/docker_release/pi.ipynb
hadia206: couldn't get that working.

hadia206: I know having duplicates is not convenient but we can address that in a later PR

buildscripts/docker/docker_release/README.md
IsaacWarren: I agree that jupyterlab is a good name, but if you don't like it maybe bodo_example? If we go with that I think we need to setup a minimal image too

IsaacWarren (approved):
Thanks Hadia

buildscripts/docker/docker_release/README.md
srilman: Generally I've seen images named by what features or special software is installed. For example, I've seen like `spark-sql`, spark-jupyter` etc based on what additional functionality is supported. So even though you dont need to use `jupyterlab` with the image, because its installed I think it should be named that.

buildscripts/docker/docker_release/env.yaml
srilman: No thats not necessarily true. Take my example.

We release 2025.1.1.
Linux Conda 2025.1.1 is built.
We notice a problem and then release 2025.1.2
Linux Conda 2025.1.2 is built.
The docker image for 2025.1.1 runs, it will pick up Bodo 2025.1.2.

Now ofc if we're good at keeping up with the CIs and make sure to cancel 2025.1.1 before its too late, then this is fine. But I don't know if we always are. This can be a followup fix though

srilman:
Also @hadia206 what do we do about the missing Codecov main report?

hadia206:
> Also @hadia206 what do we do about the missing Codecov main report?

For that, The problem is main need to run with codecov and generate a coverage file for it.
For now, I'll try with changing base to be Hadia/codecov and see if that works for now.

PR #68, opened 2024-12-15
Add files via upload

marquisdepolis:
Adding the benchmark image to Readme


ehsantn (approved):
Thanks @marquisdepolis . I think we can improve this graph a bit later too (move PySpark 20x note up, move Bodo bar left etc).

ehsantn:
Merging since urgent for launch.

PR #77, opened 2024-12-19
[BSE-4404] Add local notebooks, update README

scott-routledge2:

Adds local notebooks and results for a slightly larger dataset. Also includes a pandas notebook for comparison which is about on the same as Dask (3-4x slower than bodo).

## Testing strategy

Ran notebooks locally

## User facing changes
No

## Checklist


marquisdepolis (approved):

benchmarks/README.md
```diff
@@ -35,39 +35,38 @@ The graph below summarizes the total execution time of each system (averaged ove
...
 pip install bodo==2024.12.1
 pip install "dask[dataframe]"==2024.12.0
 pip install "modin[all]"==0.32.0
 pip install pyspark==3.5.3  
+pip install pandas==2.2.3
 pip install boto3 # for S3 download
 ```
 
-To run the entire benchmarks as a script
+We use a smaller subset of the [For Hire Vehicle High Volume dataset](https://www.nyc.gov/site/tlc/about/tlc-trip-record-data.page) to allow the workload to run locally on an Apple M2 Macbook Pro with 10 cores and 16 GB memory. We have also included an implementation using Pandas, which is often on par with or better than other distributed frameworks when data size is smaller due to the overhead from coordinating parallel workers. Even at this smaller scale, Bodo shows a roughly 3x improvement over the next best system (Dask/Pandas).
```
ehsantn, line 49:
```suggestion
We use a smaller subset of the [For Hire Vehicle High Volume dataset](https://www.nyc.gov/site/tlc/about/tlc-trip-record-data.page) to allow the workload to run locally (example setup: an Apple M2 Macbook Pro with 10 cores and 16 GB memory). We have also included an implementation using Pandas, which is often on par with or better than other distributed frameworks when data size is smaller due to the overhead from coordinating parallel workers. Even at this smaller scale, Bodo shows a roughly 3x improvement over the next best system (Dask/Pandas). 
```

ehsantn (approved):

PR #81, opened 2024-12-19
Handle nested objmode functions in UDF inlining

ehsantn:
## Changes included in this PR

Makes sure UDF inlining doesn't inline nested functions that have objmode in them since this causes compilation failures. Noticed the issue when trying an AI example.

## Testing strategy

Added a unit test.

## Checklist
- [ ] Pipelines passed before requesting review. To run CI you must include `[run CI]` in your commit message.
- [x] I am familiar with the [Contributing Guide](https://github.com/bodo-ai/Bodo/blob/main/CONTRIBUTING.md) 
- [x] I have installed + ran pre-commit hooks.

PR #80, opened 2024-12-19
[BSE-4374] Patch numba for object mode globals in spawn mode

scott-routledge2:
## Changes included in this PR

Attach function global dictionary to `global_dict` of function descriptor when running in object mode with block. Otherwise keep the old behavior which avoids setting global_dict for caching.

## Testing strategy

Ran CI and Nightly

## Checklist
- [ ] Pipelines passed before requesting review. To run CI you must include `[run CI]` in your commit message.
- [x] I am familiar with the [Contributing Guide](https://github.com/bodo-ai/Bodo/blob/main/CONTRIBUTING.md) 
- [x] I have installed + ran pre-commit hooks.

PR #82, opened 2024-12-19
Add preprocess the pile AI example

ehsantn:

Adds an example demonstrating how to do data preprocessing for AI use cases using Bodo. It preprocesses "the pile" data from Hugging Face.

## Testing strategy

Tested it manually with sample data locally. Need to test on larger AWS instances with full data.

## User facing changes

A new example.

## Checklist


ehsantn:
Waiting for #80 and #81 to test at scale.

ehsantn:
@scott-routledge2 could you test with full data on the platform and collect performance data? You can install the Hugging Face dependencies using `pixi add datasets=3.2.0 transformers=4.47.1`.

scott-routledge2:
> @scott-routledge2 could you test with full data on the platform and collect performance data? You can install the Hugging Face dependencies using `pixi add datasets=3.2.0 transformers=4.47.1`.

What is meant by "full data" is there a different dataset name or split?

ehsantn:
The "monology/pile-uncopyrighted" is the full data which is already in the code. Maybe remove the "revision" argument if it doesn't have all the data.

ehsantn:
Removed.

examples/preprocess_thepile_bodo.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+@bodo.jit
+def preprocess_pile(df, out_file):
+    df["text"] = df["text"].map(clean_text)
+    df["text_hash"] = df["text"].map(hash_text)
+    df = df.drop_duplicates(subset=["text_hash"])
+    df = df.drop("text_hash", axis=1)
+    processed_data = df.apply(tokenize_data, axis=1)
+    processed_data.to_json(out_file, orient="records", lines=True)
```
scott-routledge2, line 68: For the full dataset it might make sense to write to parquet files as opposed to a single jsonl

scott-routledge2 (approved):
LGTM!

examples/preprocess_thepile_bodo.py
ehsantn: This is from the original example (I assumed the downstream system expects JSONL). Do we write separate JSONL files per core?

scott-routledge2: Makes sense, and no it looks like there is a single jsonl at the end.

marquisdepolis:
LGTM. Could we add a version to test this on Dask too to compare?

marquisdepolis (approved):

ehsantn:
> LGTM. Could we add a version to test this on Dask too to compare?

I think so. Please open an issue. I'd compare to Dask after Hugging Face data load is implemented.

PR #84, opened 2024-12-20
[BSE-4389] Add Bodo Platform SDK Guide in the quick start.

mbojanowski:
Add Bodo Platform SDK Guide in the quick start. Where we have:

* create cluster
* run simple job
* execute query 
* execute query using connector

also added missing reference in Bodo SDK Guide

image

## Checklist


mbojanowski:
We can create additional guide which is a little more simple but, then we will need to maintain two guides, which is not ideal.

ehsantn (changes requested):
The intention of the issue wasn't moving the guide. The intention is adding a very simple SDK example to the cloud quick start.

docs/mkdocs.yml
```diff
@@ -164,10 +165,10 @@ nav:
           - Managing Packages Manually: guides/using_bodo_platform/packages.md
           - Running Shell Commands on a Cluster: guides/using_bodo_platform/shell.md
           - Connecting to a Cluster: guides/using_bodo_platform/connect.md
+          - Bodo Platform SDK: guides/using_bodo_platform/bodo_platform_sdk_guide.md
           - Configuring Customer Managed VPC: guides/using_bodo_platform/customer_managed_vpc.md
           - AWS PrivateLink: guides/using_bodo_platform/aws_private_link.md
           - Troubleshooting: guides/using_bodo_platform/troubleshooting.md
-          - Bodo Platform SDK guide : guides/using_bodo_platform/bodo_platform_sdk_guide.md
```
ehsantn, line 170: Let's undo this change.

mbojanowski: Done

mbojanowski:
I got few improvements from @marquisdepolis:

* Rename similar sounding Bodo Platform SDK pages to explain what they are!
  * now we have `Platform SDK Quick Start` and `Platform SDK Guide`
  * we also agreed on removing `docs/docs/installation_and_setup/platform_sdk.md` as this is already in both `Platform SDK Quick Start` and `Platform SDK Guide`

* Put bodo platform sdk left hand side as second suggestion"

image
not sure about that as first I think user should also understood Bodo Platform in general.

* How to run a piece of code on a piece of data on one's laptop
   * this will require feature in Bodo Platform, right now not supported

* We should support reading in files from local environment
   * this also will require feature in Bodo Platform, right now not supported

* Add reference on source to be github, s3, workspace etc -- show how. Change code to read from a public s3 bucket and does some work on it - in the example
   * looks like we don't have public S3/Github with examples right now, 

* Change the search too to hide older pages:
  * search is not showing older pages, when you type `Bodo SDK` it will first show all pages that have `Bodo` and `SDK` but also it will show all pages with only Bodo or SDK thats why we got so many results

mbojanowski:
I also changed some left menu list:

from:
image

to:
image


from:
image

to:
image

mbojanowski:
Form this:
image

to:

image

mbojanowski:
Form this:
image

to this:
image

marquisdepolis (approved):

ehsantn (approved):
Thanks @mbojanowski !

PR #86, opened 2024-12-20
[BSE-4415] Support large counts in MPI scatterv

scott-routledge2:

Changes scatterv to use [scatterv_c](https://www.mpich.org/static/docs/v4.0.2/www3/MPI_Scatterv_c.html) to support larger send counts. 

## Testing strategy

PR CI, Nightly, local testing:

```py
import pandas as pd
import bodo

@bodo.jit
def test(df):
    return df.A.str.len().sum()

df = pd.DataFrame({"A": ["abcd"*10000]*1000000})
print(test(df))
```
Now fails with a different error:
`bodo.utils.typing.BodoError: data type string for column A not supported yet`
which is really supposed to be pyarrow memory error `Could not find an empty frame of required size (68719476736)!` from trying to fit an array that is too large (37 gb) into memory, but the exception gets caught in `boxing._infer_series_arr_type` and overwritten. I think we should ideally fix the error handling here so that the error is more helpful.

But if we decrease the size of the outer array to 100000 (~4 gb) then it runs on this branch but not on main.

## Checklist

### DIFF
--- a/bodo/libs/_distributed.h
+++ b/bodo/libs/_distributed.h
@@ -123,9 +123,9 @@ static void c_gatherv(void* send_data, int sendcount, void* recv_data,
                       int* recv_counts, int* displs, int typ_enum,
                       bool allgather, int root,
                       int64_t comm_ptr = 0) __UNUSED__;
-static void c_scatterv(void* send_data, int* sendcounts, int* displs,
-                       void* recv_data, int recv_count, int typ_enum, int root,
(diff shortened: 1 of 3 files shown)


scott-routledge2:
Looks like there are some seg faults on Nightly. Looking into them...

ehsantn (approved):
Looks good to me. Thanks @scott-routledge2 !

IsaacWarren (approved):
Thanks Scott

PR #88, opened 2024-12-22
Create a simple JIT wrapper for regular Python functions

ehsantn:

Adds a `wrap_python` decorator that is both easier to use and also much faster than objmode (doesn't require compilation during runtime). This interface fits the common UDF case for calling regular Python especially well.

## Testing strategy

Added unit test and also tested the AI "the pile" data processing example (will be added in a separate PR).

## User facing changes

Adds `wrap_python` decorator.

## Checklist

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -104,8 +104,6 @@ snowflake-connector-python = "*"
 xlrd = "*"
 xlsxwriter = "*"
 openpyxl = "*"
-datasets = "3.2.0.*"
-transformers = "4.47.1.*"
 [target.linux-64.dependencies]
 cx_oracle = "*"
 pyarrow = { version = "18.1.0", channel = "bodo.ai" }
--- a/pixi.toml
+++ b/pixi.toml
@@ -104,6 +104,8 @@ snowflake-connector-python = "*"
 xlrd = "*"
 xlsxwriter = "*"
 openpyxl = "*"
+datasets = "3.2.0.*"
+transformers = "4.47.1.*"
 [target.linux-64.dependencies]
 cx_oracle = "*"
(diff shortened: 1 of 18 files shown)


njriasan:
@ehsantn Can you provide me some context on this change? Was this directly requested or has this been a known TODO for some time?

ehsantn:
This came out of my work on AI examples and seeing the issues of objmode (interface, performance, reliability). This will have more feedback and iterations soon probably.

njriasan (approved):
LGTM! I don't have enough context on the reasoning for this change, but the implementation of this change seems reasonable.

The biggest possible concern is that Numba is looking to decrease the number of JIT decorators they maintain to reduce difficulties with their API. If there is a bug fix to say objmode that improves the quality of life I want to ensure that it seems likely this new API will pick it up.

ehsantn:
I haven't seen activity on the objmode side but will monitor.

pixi.toml
```diff
@@ -104,6 +104,8 @@ snowflake-connector-python = "*"
 xlrd = "*"
 xlsxwriter = "*"
 openpyxl = "*"
+datasets = "3.2.0.*"
+transformers = "4.47.1.*"
```
scott-routledge2, line 108: I'm assuming these were for the AI examples? Are they needed at this point?

scott-routledge2 (approved):
LGTM, thanks @ehsantn !

pixi.toml
ehsantn: Got added by mistake. Will remove now.

ehsantn: Done.

PR #89, opened 2024-12-23
[BSE-4419] Support large counts in MPI gatherv

scott-routledge2:

Changes `MPI_Gengatherv` and `c_gatherv` to use `MPI_Gatherv_c`/ `MPI_Allgatherv_c`. For spawn mode, most cases will take the `MPI_Gengatherv` path via `gather_array` but also including `c_gatherv` here for robustness. 

## Testing strategy

PR CI, Nightly, locally with:
```
import pandas as pd
import bodo

bodo.set_verbose_level(2)

@bodo.jit
def test(df):
    return df

df = pd.Series(["abcd"*10000]*100000)
df = test(df)
print(df.head(15))
```
`BODO_NUM_WORKERS=2 python -u test.py`

## Checklist


bodo/hiframes/boxing.py
```diff
@@ -936,6 +936,9 @@ def _infer_series_arr_type(S: pd.Series, array_metadata=None):
             arr_type = types.Array(arr_type.dtype, 1, "C")
 
         return arr_type
+    except pa.lib.ArrowMemoryError:  # pragma: no cover
+        # OOM
+        raise
```
scott-routledge2, line 941: I noticed this when scattering Series that are too large to fit into memory `S.array` can fail. Before the error message would be:
```
bodo.utils.typing.BodoError: data type string for column A not supported yet
``` Now the error looks like:
```
pa.lib.ArrowMemoryError: Could not find an empty frame of required size (68719476736)! 
```

File: bodo/libs/_distributed.h [hunk 2]
```diff
@@ -993,21 +993,24 @@ static void c_gather_scalar(void* send_data, void* recv_data, int typ_enum,
     return;
 }
 
-static void c_gatherv(void* send_data, int sendcount, void* recv_data,
-                      int* recv_counts, int* displs, int typ_enum,
+static_assert(sizeof(MPI_Count) == sizeof(int64_t));
+static_assert(sizeof(MPI_Aint) == sizeof(int64_t));
```
scott-routledge2, line 997: The documentation says that `MPI_Aint`  it is a "C type that holds any valid address." On my machine it is typedef'd as `long int`. We'd need to check that it is still 64 bits on windows.

scott-routledge2:
A couple failures on Nightly:
`test_iceberg_ddl.py::test_alter_table_drop_column[tabular] ` and
`test_snowflake_catalog.py::test_filter_pushdown_row_count_caching[test_db_snowflake_catalog0]`
These tests appear to be flakey, both pass for me locally and I think they should be safe to ignore.

njriasan (approved):
LGTM! Thanks @scott-routledge2

ehsantn (approved):
Looks good to me. Thanks @scott-routledge2 !

PR #94, opened 2024-12-27
Add LLM query example with Ollama

ehsantn:

Adds an example that demonstrates preprocessing data and querying an LLM using Bodo (with Ollama).

## Testing strategy

Tested it locally.

## User facing changes

A new example.

## Checklist

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2024 Bodo.ai
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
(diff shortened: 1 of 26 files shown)


scott-routledge2 (approved):
LGTM, is there a larger dataset of queries we could use to show performance benefits over python?

ehsantn:
Running LLMs locally without GPUs is very slow so larger data would make it not usable.

marquisdepolis:
Ollama is meant for local runs, smaller scale w/o GPUs. Next we can use
OpenAI compatible APIs, via https://github.com/andrewyng/aisuite or
LiteLLM, at larger scale, and see how that helps.

marquisdepolis (approved):
I'd edit to model="llama3", or even make that an outside variable, to keep it cleaner. And add a comment next to raw_prompts*10 that this just runs the examples 10x times.

ehsantn:
Done.

PR #107, opened 2025-01-03
Move Relational Algebra Python APIs to an explicit Python Entry Point API

njriasan:

One of the challenges with understanding the BodoSQL Java code is understanding how it interacts with Python. This is particularly challenging because most IDEs come with functionality for detecting dead code, but this doesn't work properly for anything exposed to Python because the function is essentially an entry point.

This PR moves the Python entry points for the RelationalAlgebraGenerator file to its own file. I believe this makes the code more understandable for 3 reasons:

1. You can explicitly tell which APIs are used from Python. This is especially important if we want to refine or port the Python APIs.
2. It allows dead code to be properly detected. For example one of our `getPandasString` APIs was exposed just for Python, but it became unused after our named parameter changes, which I could not properly detect.
3. It standardizes the entry from Python, making "cost" or the transition more transparent. If we wanted to expand a compilation service or decouple the planner from code generation for example, this should make it more feasible to define the required API boundary.

In a followup PR I will attempt to update the remaining Python entry points for BodoSQL, including constructors.

## Testing strategy

I am depending on PR CI since I cannot feasibly test all of the BodoSQL Python code. Since this refactors the APIs it should be "all or nothing" for API breakage.

## Checklist


BodoSQL/bodosql/context.py
```diff
@@ -765,19 +766,19 @@ def _compile(self, sql, params_dict=None, dynamic_params_list=None):
             False,  # We need to execute the code so don't hide credentials.
         )
         if bodo.get_rank() == 0:
-            is_dll = generator.isDDLProcessedQuery()
+            is_ddl = JavaEntryPoint.isDDLProcessedQuery(generator)
```
njriasan, line 769: My hope is the `JavaEntryPoint` can be a standard for understanding an API moves to Java, at least for low level operations.

BodoSQL/bodosql/imported_java_classes.py
```diff
@@ -49,6 +49,10 @@
         # Initialize logging. Must be done after importing all classes to ensure
         # PythonLoggersClass is available.
         configure_java_logging(bodo.user_logging.get_verbose_level())
+        # Note: We call this JavaEntryPoint so its clear the Python code enters java
+        # and the class is named PythonEntryPoint to make it clear the Java code
+        # is being entered from Python.
+        JavaEntryPoint = gateway.jvm.com.bodosql.calcite.application.PythonEntryPoint
```
njriasan, line 52: My hope is that a lot of these other entry points can be removed soon, at least for things like constructors that aren't tested/used in Java.

BodoSQL/bodosql/tests/test_bodo_and_bodosql.py
```diff
@@ -39,9 +40,9 @@ def test_planner_reset(memory_leak_check):
     queries = ["SELECT 1", "SELECT COUNT(*) FROM TABLE1", "Select MAX(A) FROM TABLE1"]
     final_query = "DESCRIBE TABLE MYTABLE"
     for query in queries:
-        generator.parseQuery(query)
-        generator.resetPlanner()
-    generator.parseQuery(final_query)
+        JavaEntryPoint.parseQuery(generator, query)
```
njriasan, line 43: This is an example of what I hope will serve as the example pattern. This makes it clear we cross into Java, which would otherwise require an understanding of when a method implicitly crosses into Java.

BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/application/RelationalAlgebraGenerator.java
```diff
@@ -424,7 +423,7 @@ public void resetPlanner() {
    * @param sql Query to parse Sets parseNode to the generated SQLNode
    * @throws SqlSyntaxException if the SQL syntax is incorrect.
    */
-  public void parseQuery(String sql) throws SqlSyntaxException {
+  void parseQuery(String sql) throws SqlSyntaxException {
```
njriasan, line 426: These APIs can be moved to package private. While any code can use the Python APIs, this helps clarified that they are not the desired API for the Java code.

IsaacWarren (approved):
Thanks Nick

njriasan:
All failures are unrelated to this PR.

knassre-bodo (approved):
Makes perfect sense to me

PR #114, opened 2025-01-03
[BSE-4454] Tweak Buffer Pool Memory Allocation Default to Avoid OS OOMs outside of Platform

srilman:

Right now, it's pretty easy for users to cause an OS-level OOM with Bodo. On remote systems like the platform and VMs, its not a big deal because they are usually not configured with swap and there are very few external services. The only important one is SSH. But on local devices like a laptop, this is really bad because the OS will be hammered by the extreme amount of memory and will avoid killing the process.

Thus, this PR modifies the buffer pool defaults to use memory conservatively, while introducing an environment variable to unlock support for remote platforms.

## Testing strategy

Mainly local testing, testing benchmarks while monitoring memory usage.

## User facing changes

Some workloads that would fit nearly perfectly in memory (so more than 80% of physical RAM but less than double) will OOM without setting the env. We need to introduce a page in our docs about the buffer pool in general.

## Checklist


benchmarks/nyc_taxi/bodo/nyc_taxi_precipitation.py
```diff
@@ -101,5 +101,5 @@ def get_time_bucket(t):
 
 if __name__ == "__main__":
     weather_dataset = "s3://bodo-example-data/nyc-taxi/central_park_weather.csv"
-    hvfhv_dataset = "s3://bodo-example-data/nyc-taxi/fhvhv/"
+    hvfhv_dataset = "s3://bodo-example-data/nyc-taxi/fhvhv_tripdata/"
```
srilman, line 104: I think this path was wrong, I modified it to the correct path

benchmarks/nyc_taxi/pandas/nyc_taxi_precipitation.py
```diff
@@ -0,0 +1,102 @@
...
+Similar to:
+https://github.com/toddwschneider/nyc-taxi-data/blob/c65ad8332a44f49770644b11576c0529b40bbc76/citibike_comparison/analysis/analysis_queries.sql#L1
+"""
+
+import time
+
+import pandas as pd
+
+
+def get_monthly_travels_weather(weather_dataset, hvfhv_dataset):
```
srilman, line 13: All I did was remove the bodo decorator

bodo/libs/_memory.cpp
```diff
@@ -703,10 +703,27 @@ BufferPoolOptions BufferPoolOptions::Defaults() {
         options._allocate_extra_frames = true;
     } else {
         // Fraction of total memory we should actually assign to the buffer
-        // pool. We will read this from an environment variable if it's set, but
-        // will default to 95% if spilling is available and enabled, and to 500%
-        // otherwise.
-        double mem_fraction = options.storage_options.empty() ? 5. : .95;
+        // pool.
```
srilman, line 706: Main changed

bodo/libs/_memory.cpp
```diff
@@ -703,10 +703,27 @@ BufferPoolOptions BufferPoolOptions::Defaults() {
...
+        // - 0.8 is the default value.
+        // - If we're running on a remote system, we use 0.95 with spilling
+        //   enabled, 5.0 without spilling enabled.
+        // - If users set an environment variable, we use that fraction.
+        double mem_fraction = 0.8;
```
ehsantn, line 711: Isn't this way too low? We usually encountered significant underutilization of the system's physical memory. That's why we set 5 as the default I believe.

One factor is streaming since buffers double in size repeatedly but a lot of the memory isn't actually touched. We will use streaming in Python later too though.

srilman: So my concern is that on local systems like a laptop, people are not going to run Bodo by itself. They usually will have a web browser open with a couple of tabs, maybe VSCode or JupyterLab, and some other stuff. That already incurs a lot of memory utilization and swap space. For example, on my laptop having VSCode open and 4 Firefox tabs (2 of which are just static pages) had about 4GB / 16GB in memory and 4GB swap. Now I do have more background things than others, like I used a fancier LSP in VSCode in all that. But I don't think 80% is unreasonable.

srilman: When we start integrating streaming with Python, we can check back on this, but we cant just assume that Bodo will be the only CPU and memory intensive process. We may need to invest time working on reducing the amount of underutilized memory, like via segmented data structures (https://github.com/martinus/unordered_dense?tab=readme-ov-file#4-segmented_map-and-segmented_set) or array packing in buffers.

srilman: Oh plus by that point we need to figure out the spilling config anyways, so it might be a moot point

ehsantn: 80% of physical memory is not unreasonable. My concern is that our logical size may end up 20% of physical memory for example.

ehsantn: What do others do?

srilman: Maybe I should clarify, I'm opening up a PR on the platform to set the env `BODO_BUFFER_POOL_REMOTE_MODE=1` on the AMI, so the platform is unaffected. Plus, on benchmarking, we can use that env.

srilman: DuckDB does the exact same thing: https://duckdb.org/docs/operations_manual/limits.html. 80% of RAM. But I know that they've put in the effort to improve the memory packing.

IsaacWarren:
One additional thought, Can we mention the environment variable in the OOM message?

benchmarks/nyc_taxi/bodo/nyc_taxi_precipitation.py
scott-routledge2: Good catch, thanks!

scott-routledge2 (approved):
LGTM! We could also stress test this further by adding more parquet files to the benchmark (Locally I was able to run up to 6 files although I'm not sure if I had other apps open though):
```
hvfhv_dataset = [
    f"s3://bodo-example-data/nyc-taxi/fhvhv_tripdata/fhvhv_tripdata_2019-{i:02}.parquet"
    for i in range(2, 8)
]
```

srilman:
@IsaacWarren I'm going to open a follow-up task to improve the error messaging of the Buffer Pool in general, because its unclear for random people if its a OOM

srilman:
Looks like the failed tests are previously fixed, known for other issues so merging

PR #83, opened 2024-12-20
[BSE-4386] Add Linux ARM Support

srilman:
## Changes included in this PR

Add support for Linux on the ARM / Aarch64 architecture. All dependencies already support it, its just a matter of setting up the build correctly

NOTE: I'm a bit blocked on the Azure testing portion, but I want to get the packaging parts to review before going OOO, so opening it up now.

## Checklist
- [ ] Pipelines passed before requesting review. To run CI you must include `[run CI]` in your commit message.
- [ ] I am familiar with the [Contributing Guide](https://github.com/bodo-ai/Bodo/blob/main/CONTRIBUTING.md) 
- [ ] I have installed + ran pre-commit hooks.

PR #118, opened 2025-01-06
Use warning instead of value error when tests are not being run in `check_query`

scott-routledge2:

As titled. This is to enable a green nightly, we can later go back and look at the tests that failed on an individual basis to see if they need to be removed/reworked.

## Testing strategy
Locally/on CI

## User facing changes
No

## Checklist

### DIFF
--- a/BodoSQL/bodosql/tests/utils.py
+++ b/BodoSQL/bodosql/tests/utils.py
@@ -322,7 +322,8 @@ def check_query(
             return
 
     if not (run_jit_seq or run_jit_1D or run_jit_1DVar or run_python):
-        raise ValueError("check_query: No tests are being run.")
+        warnings.warn("check_query: No tests are being run.")
+        return
 
     # If a user sets BODOSQL_TESTING_DEBUG, we print the
     # unoptimized plan, optimized plan, and the Pandas code


IsaacWarren (commented):
How many tests was this an issue for? Thanks Scott

scott-routledge2:
@IsaacWarren It's a lot, maybe around 300? Not sure how many separate tests there are though.

IsaacWarren:
> @IsaacWarren It's a lot, maybe around 300? Not sure how many separate tests there are though.

Do we know when this started? Seems concerning, should we try to fix them instead of merging this? That seems like too many broken tests to do a release IMO

ehsantn:
Looks like this is reverting a change from a recent PR: https://github.com/bodo-ai/Bodo/pull/83/files#diff-371bc62229377d93fed7d1bb28fbb123437b4f9d6bf9c1675de84f2ef9d14da7L325

IsaacWarren:
> Looks like this is reverting a change from a recent PR: https://github.com/bodo-ai/Bodo/pull/83/files#diff-371bc62229377d93fed7d1bb28fbb123437b4f9d6bf9c1675de84f2ef9d14da7L325

Ah ok, thanks for the context, approving

scott-routledge2:
@IsaacWarren It is sort of by design that the tests aren't being run in certain situations on azure to save time since we know another runner will run the test. So for example if `check_query` is passed as `only_1D_jit=True` it is skipped on one rank but the test still gets executed in the multi rank case.

The warning was added here because we had a situation where we were testing something with `only_1D_jit` and inspecting the logs which failed because the test wasn't being run on one rank. Not sure why it was changed to a ValueError recently but I assumed we wanted to make the skipping more explicit or double check we are skipping where it doesn't makes sense.

srilman (approved):

PR #120, opened 2025-01-06
Move intervalToNanos out of the Calcite files and into our parser utils functions

njriasan:

Changes to `SqlParserUtil.java` should only be those breaking changes that we need to back port into importing Calcite (and should only be done when we do not actively include all uses of the file already in our source code). Since `intervalToNanos`, which is meant to replace the `intervalToMilli` in Calcite requires a new function signature, every usage must exist inside our source code, so it should existing our extension files.

Ideally in the future we could also remove `SqlParserUtil` in general, but right now we need the remaining changes to ensure we get some default precisions because of its reuse in the validator. 

## Checklist


IsaacWarren (approved):

scott-routledge2 (approved):
LGTM! Thanks @njriasan

PR #124, opened 2025-01-07
Replace stream operator state typing to use Numba type refinement

ehsantn:

Updates typing for streaming state types to use Numba's type refinement and avoid repeated IR transformation. This can improve compilation time in streaming case (currently SQL but later Python) significantly. Also, eliminates the transformation code which is very hard to develop and maintain.

## Testing strategy

Ran all CI/nightly tests.

## User facing changes

None.

## Checklist


bodo/libs/streaming/groupby.py
```diff
@@ -695,76 +710,13 @@ def init_groupby_state(
     mrnf_sort_col_na=None,
     mrnf_col_inds_keep=None,
     op_pool_size_bytes=-1,
-    expected_state_type=None,
     parallel=False,
 ):
     pass
 
 
-def _get_init_groupby_state_type(
```
ehsantn, line 704: Just separated validation from initial type creation.

ehsantn:
All CIs passed except Azure nightly, which had existing errors that are fixed on main. Merged with main and reran just in case but this should be good to go.
https://dev.azure.com/bodo-inc/Bodo/_build/results?buildId=25410&view=results

bodo/hiframes/table.py
```diff
@@ -158,6 +158,24 @@ def to_pandas(self, index=None):
         return df
 
 
+def _unify_array_types(typingctx, t1, t2):
```
njriasan, line 161: So one issue with just using this function is that the name implies its a "general purpose" array unification function for arrays (which maybe that is the case because if you have conditionals forcing two different integer bit widths this is a reasonable behavior.

However, the use of `_derive_common_key_type` suggests to me that we could have a major issue if this is used outside of a context where key unification is expected (e.g. the data columns of the join). I don't have any objection with this behavior because I think it matches the internal expected behavior, but let's add a TODO to make the `_derive_common_key_type` more generic and follow more "general Bodo rules" for legal type unification.

bodo/hiframes/table.py
```diff
@@ -223,6 +241,23 @@ def as_array(self):
     def key(self):
         return self.arr_types, self.has_runtime_cols
 
+    def unify(self, typingctx, other):
+        """Unify two TableType instances (required for runtime join filter typing)."""
+        if (
+            isinstance(other, TableType)
+            and (len(self.arr_types) == len(other.arr_types))
+            and (self.has_runtime_cols == other.has_runtime_cols)
```
njriasan, line 249: QQ: Does the provided implementation work properly for runtime array columns (e.g. unify DataFrame types after a pivot was performed). If not it might be safe to replace this with `and not  self.has_runtime_cols and not other.has_runtime_cols` and include a TODO to do a deeper dive for this edge case later.

bodo/libs/streaming/groupby.py
```diff
@@ -164,6 +164,19 @@ def ftypes(self):
             func_types.append(supported_agg_funcs.index(fname))
         return func_types
 
+    def is_precise(self):
+        return self.build_table_type != types.unknown
+
+    def unify(self, typingctx, other):
```
njriasan, line 170: Is it possible to move this functionality into the parent class. Something like defining `is_precise` as an abstract method (actually I assume its already defined by the type interface) and just making the common implementation:

```python
if type(self) is type(other):
    if not other.is_precise() and self.is_precise():
        return self
    return other
```

bodo/libs/streaming/union.py
```diff
@@ -58,6 +55,28 @@ def __init__(
     def key(self):
         return (self.all, self.in_table_types)
 
+    def is_precise(self):
+        return not (
+            (len(self.in_table_types) == 0)
+            or any(t == types.unknown for t in self.in_table_types)
+        )
+
+    def unify(self, typingctx, other):
```
njriasan, line 64: It looks like this would be the only subclass that is required to implement its own unify if you move it to the parent class.

bodo/numba_compat.py
```diff
@@ -2975,6 +3005,9 @@ def CallConstraint_resolve(self, typeinfer, typevars, fnty):
     for a in itertools.chain(pos_args, kw_args.values()):
         # Forbids imprecise type except array of undefined dtype
         if not a.is_precise() and not isinstance(a, types.Array):
+            # Bodo change: allow streaming state type to be imprecise
```
njriasan, line 3008: For sustainability it might be safer to wrap this in an API or a property of the type. Not super critical right now, but this will make this easier to extend in anything for streaming Python requires it.

bodo/transforms/typing_pass.py
njriasan, line 1: Really great to see you were able to remove so much code.

bodo/utils/transform.py
```diff
@@ -1698,13 +1698,18 @@ def set_ith_arg_to_omitted_value(
     assert call_type.args[pos_idx] == types.Omitted(
         expected_existing_value
     ), f"Omitted({expected_existing_value}) {pos_idx}th argument expected"
-    pass_info.calltypes[rhs] = pass_info.typemap[rhs.func.name].get_call_type(
```
njriasan, line 1701: I see several more cases of this in other APIs that are perhaps not interacting with streaming. For example `_recompile_func` in distributed analysis. Do these need to be updated as well? If so we should probably encapsulate the return type replacement into a a common API for replacement.

bodo/libs/streaming/join.py
```diff
@@ -141,6 +140,38 @@ def key(self):
...
+        return (
+            self.build_table_type != types.unknown
+            and self.probe_table_type != types.unknown
+        )
+
+    def unify(self, typingctx, other):
+        """Unify two JoinStateType instances when build_table_type/probe_table_type
+        are not resolved.
+        """
+        if isinstance(other, JoinStateType):
```
njriasan, line 153: This is the other unify that would need to persist. However if we add a StreamingStateType `canUnify()` that allows us to just capture the additions to the if check we can simplify the changes per function (including having it default to True).

ehsantn:
This PR makes this query ~15% faster to compile.

```sql
select count(*)
from
    customer,
    orders,
    lineitem,
    supplier,
    nation,
    region
where
    c_custkey = o_custkey
    and l_orderkey = o_orderkey
    and l_suppkey = s_suppkey
    and c_nationkey = s_nationkey
    and s_nationkey = n_nationkey
    and n_regionkey = r_regionkey
```

njriasan (approved):
Thanks @ehsantn! Overall this looks great. I have a few questions/suggestions to cleanup the code.

One question I have though is if you have inspected the IR for any of the streaming examples. I remember that sometimes `unify` was generating casts in the IR for correctness (this was most commonly with if statements so I think its unlikely for streaming). However, can you confirm that you have inspected the IR and don't see any unexpected casts. Thanks!

bodo/io/snowflake_write.py
```diff
@@ -50,13 +50,27 @@
...
+        """Unify two SnowflakeWriterType instances when one doesn't have a resolved
+        input_table_type.
+        """
+        if isinstance(other, SnowflakeWriterType):
+            if not other.is_precise() and self.is_precise():
+                return self
+
+            # Prefer the new type in case write append changed its table type
+            return other
+
```
DrTodd13, line 73: Duplicate code as below?  Move to StreamingStateType?

bodo/io/stream_parquet_write.py
```diff
@@ -46,13 +46,27 @@
...
+        """Unify two ParquetWriterType instances when one doesn't have a resolved
+        input_table_type.
+        """
+        if isinstance(other, ParquetWriterType):
+            if not other.is_precise() and self.is_precise():
+                return self
+
+            # Prefer the new type in case append table changed its table type
+            return other
+
```
DrTodd13, line 69: Same duplication comment as above.  Just type is differing so abstract it?

bodo/io/stream_parquet_write.py
```diff
@@ -539,6 +531,10 @@ class ParquetWriterAppendInfer(AbstractTemplate):
     def generic(self, args, kws):
         pysig = numba.core.utils.pysignature(parquet_writer_append_table)
         folded_args = bodo.utils.transform.fold_argument_types(pysig, args, kws)
+        # Update state type in signature to include build table type from input
+        input_table_type = folded_args[1]
+        new_state_type = ParquetWriterType(input_table_type)
+        folded_args = (new_state_type, *folded_args[1:])
```
DrTodd13, line 537: This pattern also seems to be duplicated?  Should be its own function in StreamingStateType?

bodo/hiframes/table.py
ehsantn: Not sure if it would work. Disabled for now as suggested.

bodo/libs/streaming/groupby.py
ehsantn: I feel like it's safer to have it explicitly here and a few lines of code duplication is worth it for this critical codes.

bodo/libs/streaming/union.py
ehsantn: See above.

bodo/utils/transform.py
ehsantn: I don't think anything else needs to be changed since the streaming change is limited to these cases.

bodo/libs/streaming/join.py
ehsantn: See above.

bodo/io/snowflake_write.py
ehsantn: I feel like it's safer to have it explicitly here and a few lines of code duplication is worth it for this critical codes.

bodo/io/stream_parquet_write.py
ehsantn: Same as above.

bodo/io/stream_parquet_write.py
ehsantn: Same as above.

ehsantn:
> Thanks @ehsantn! Overall this looks great. I have a few questions/suggestions to cleanup the code.
> 
> One question I have though is if you have inspected the IR for any of the streaming examples. I remember that sometimes `unify` was generating casts in the IR for correctness (this was most commonly with if statements so I think its unlikely for streaming). However, can you confirm that you have inspected the IR and don't see any unexpected casts. Thanks!

Unify is called here inside `add_type` of type inference when data type is updated, so it doesn't create actual casts. We don't even have a cast function for tables looks like. I double checked the runtime join filter case which seems riskiest and it was fine (calltype's return type matches target variable type).

PR #135, opened 2025-01-12
Refactor function handling if checks in distributed_analysis.py

ehsantn:
## Checklist

### DIFF
--- a/bodo/transforms/distributed_analysis.py
+++ b/bodo/transforms/distributed_analysis.py
@@ -457,7 +457,7 @@ class DistributedAnalysis:
                     new_dist.append(None)
                     continue
                 new_d = _min_dist(d, _get_var_dist(v.name, array_dists, self.typemap))
-                _set_var_dist(v.name, array_dists, new_d)
+                _set_var_dist(self.typemap, v.name, array_dists, new_d)
                 new_dist.append(new_d)
 
             array_dists[lhs] = new_dist
@@ -678,8 +678,8 @@ class DistributedAnalysis:
             new_dist = _min_dist(l_dist[0], m_dist)
             for d in l_dist:
                 new_dist = _min_dist(new_dist, d)
-            _set_var_dist(lhs, array_dists, new_dist)
-            _set_var_dist(rhs.value.name, array_dists, new_dist)
+            _set_var_dist(self.typemap, lhs, array_dists, new_dist)
+            _set_var_dist(self.typemap, rhs.value.name, array_dists, new_dist)
             return
         elif isinstance(rhs_typ, CategoricalArrayType) and attr == "codes":
             # categorical array and its underlying codes array have same distributions
@@ -896,7 +896,7 @@ class DistributedAnalysis:
                 func_def.value, numba.core.dispatcher.ObjModeLiftedWith
             ):
                 if lhs not in array_dists:
-                    _set_var_dist(lhs, array_dists, Distribution.OneD_Var)
+                    _set_var_dist(self.typemap, lhs, array_dists, Distribution.OneD_Var)
                 return
             # some functions like overload_bool_arr_op_nin_1 may generate const ufuncs
             if isinstance(func_def, ir.Const) and isinstance(func_def.value, np.ufunc):
@@ -923,7 +923,7 @@ class DistributedAnalysis:
         # Similar to ObjModeLiftedWith, we assume out data is distributed (1D_Var)
         if isinstance(self.typemap[rhs.func.name], WrapPythonDispatcherType):
             if lhs not in array_dists:
-                _set_var_dist(lhs, array_dists, Distribution.OneD_Var)
+                _set_var_dist(self.typemap, lhs, array_dists, Distribution.OneD_Var)
 
         # Check distributed analysis call registry for handler first
         ctx = DistributedAnalysisContext(self.typemap, array_dists, equiv_set)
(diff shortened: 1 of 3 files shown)


bodo/transforms/distributed_analysis_call_registry.py
```diff
@@ -0,0 +1,634 @@
...
+        self.func_name = func_name
+        self.metadata = metadata
+        self.diag_info = diag_info
+
+
+class DistributedAnalysisCallRegistry:
+    """Registry of function call handlers for distributed analysis"""
+
+    def __init__(self):
+        self.call_map = {
```
DrTodd13, line 34: Please add description of the expected contents of each tuple.

DrTodd13 (approved):

bodo/transforms/distributed_analysis_call_registry.py
```diff
@@ -0,0 +1,634 @@
...
+        # XXX: Make this more scalable by recalling the distributed
+        # analysis already in this pass for each of the provided
+        # func names.
+        func_name = guard(get_overload_const_str, func_name_typ)
+        # We support mappable prefixes that don't need to be separate functions.
+        if func_name[0] == "~":
+            func_name = func_name[1:]
+        # Note: This isn't an elif because the ~ may modify the name
+        if func_name in (
+            "bodo.libs.array_ops.array_op_isna",
```
njriasan, line 352: We should find a way to move this to some more generic name registration, but this is not urgent at this time.

bodo/transforms/distributed_analysis_call_registry.py
```diff
@@ -0,0 +1,634 @@
...
+            # Not currently in the code because it is otherwise inlined.
+            # This should be included somewhere.
+            _meet_array_dists(ctx.typemap, lhs, rhs.args[0].name, ctx.array_dists)
+            return True
+    else:
+        # If we don't have a func, this is a shallow copy.
+        _meet_array_dists(ctx.typemap, lhs, rhs.args[0].name, ctx.array_dists)
+        return True
+
+    return False
```
njriasan, line 365: These warnings demonstrate that it would be really useful to have explicit distributed analysis only unit tests to check all of these edge cases. For example, having a single "instruction" and then have a the unit tests pass different input (or even output) distribution could test the expected behavior in a scalable way.

This seems like it could be a good first issue or initial intern project in the future. I would open this issue on github issues.

bodo/transforms/distributed_analysis_call_registry.py
```diff
@@ -0,0 +1,634 @@
...
+                metadata,
+                diag_info,
+                labels_arg_name,
+                array_dists,
+                "labels when provided are assumed to be REP",
+                rhs.loc,
+            )
+
+
+def _analyze_sklearn_score_err_ytrue_ypred_optional_sample_weight(
```
njriasan, line 574: Most of these are fairly specific copied code, which is differently reasonable for a first step. Am I correct in saying that step two will attempt to move all of these generic functions to more generic handling or will these remain in the code long-term?

bodo/transforms/distributed_analysis.py
```diff
@@ -881,6 +913,11 @@ def _analyze_call(
         self, inst, lhs, rhs, func_var, args, kws, equiv_set, array_dists
     ):
         """analyze array distributions in function calls"""
+        from bodo.transforms.distributed_analysis_call_registry import (
+            DistributedAnalysisContext,
```
njriasan, line 917: Do you know what's the source of this circular import pattern? It would be great if we could refactor a common building block to its own file so we can avoid this.

ehsantn: `distributed_analysis_call_registry` imports helpers from `distributed_analysis.py`. Doesn't seem worth refactoring right now but will do in the future.

bodo/transforms/distributed_analysis_call_registry.py
ehsantn: Done.

ehsantn:
> One things to consider is if there is/should be an interface to extend the registry. The simplest current example would be removing BodoSQL code, but this could easily apply to anyone trying to add support for a new package more easily.

I'm thinking of adding an interface when the code grows and needs to be split up into different files anyways (e.g. separate for scikit-learn).

bodo/transforms/distributed_analysis_call_registry.py
ehsantn: I think we will refactor over time but some functions are just very specific. We can refactor further into pieces though. For example, the scikit-learn portions can be in their own separate file beside the other scikit-learn code for example.

bodo/transforms/distributed_analysis_call_registry.py
```diff
@@ -0,0 +1,634 @@
...
+    _set_var_dist,
+)
+from bodo.utils.typing import BodoError, get_overload_const_str, is_overload_none
+from bodo.utils.utils import is_distributable_typ
+
+
+class DistributedAnalysisContext:
+    """Distributed analysis context data needed for handling calls"""
+
+    def __init__(self, typemap, array_dists, equiv_set, func_name, metadata, diag_info):
```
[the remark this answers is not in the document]
ehsantn: Added type annotations.

bodo/transforms/distributed_analysis_call_registry.py
```diff
@@ -0,0 +1,634 @@
...
+
+
+def meet_first_2_args_analysis(ctx, inst):
+    """Handler that meets distributions of the first two call arguments"""
+    _meet_array_dists(
+        ctx.typemap, inst.value.args[0].name, inst.value.args[1].name, ctx.array_dists
+    )
+
+
+def meet_first_2_args_analysis_xgb_fit(ctx, inst):  # pragma: no cover
```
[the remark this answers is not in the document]
ehsantn: The error is fairly specific to `xgboost.fit` since typically libraries can be called with replicated inputs too.

bodo/transforms/distributed_analysis_call_registry.py
ehsantn: I agree. Opened an issue: https://github.com/bodo-ai/Bodo/issues/141

PR #137, opened 2025-01-13
Cleaned up the optimization pipeline

njriasan:

Cleans up the optimization pipeline to remove `optimize=False`. This is because this is the only available option on production. It also removes unused code to simplify the pipeline and fixes some naming.

## Testing strategy

This relies on CI. It may require refactoring the customer tests but those should be simple.

## Checklist


BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/prepare/BodoPrograms.kt
```diff
@@ -656,46 +621,4 @@ object BodoPrograms {
...
-            materializations: List<RelOptMaterialization>,
-            lattices: List<RelOptLattice>,
-        ): RelNode = rel
-    }
-
-    /**
-     * Simple program that does nothing but dump the output to stdout.
-     * Should only be used for debugging
-     */
-    class PrintDebugProgram(
```
njriasan, line 674: We could leave this to help with debugging, but I'm not sure its currently useful. It exists in this PR history if we ever need to recover it.

IsaacWarren (approved):
Thanks Nick

BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/prepare/BodoPrograms.kt
knassre-bodo: I rec we leave it, then leave a commented-out invocation of it in the pipeline with instructions of how to use it when uncommenting.

BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/prepare/ProgramCollection.kt
```diff
@@ -2,7 +2,7 @@ package com.bodosql.calcite.prepare
 
 import org.apache.calcite.tools.Program
 
-typealias ProgramFactory = (optimize: Boolean) -> Program
+typealias ProgramFactory = () -> Program
```
knassre-bodo, line 5: I think we can go a lot further with simplifying this:
- Get rid of `ProgramCollection`
- Get rid of `PlannerType` class & all the integer choices of it
- Fill `.programs` in `PlannerImpl.kt` directly with the list of programs to be invoked

Removes at least 2 files and 3 layers of unnecessary redirection for what is now always a constant setup.

BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/prepare/BodoPrograms.kt
njriasan: I don't like the idea of introducing/leaving in commented out code, but I'll leave the implementation here.

BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/prepare/ProgramCollection.kt
njriasan: Sure happy to do this.

knassre-bodo (approved):
LGTM

PR #140, opened 2025-01-13
[BSE-4391] Rename BodoSQL pip package

hadia206:

Renamed BodoSQL package name to match upload step that looks for packages starting with `cibw-`.
[See here](https://github.com/bodo-ai/Bodo/blob/main/.github/workflows/build_wheels.yml#L204)

## Testing strategy
N/A

## User facing changes

N/A
## Checklist

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -105,7 +105,7 @@ jobs:
           pip wheel --no-deps --no-build-isolation -v .
       - uses: actions/upload-artifact@v4
         with:
-          name: bodosql-wheel-no-arch
+          name: cibw-bodosql
           path: ./BodoSQL/*.whl
 
   test-bodosql:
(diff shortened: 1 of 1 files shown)


scott-routledge2 (approved):
LGTM, thanks @hadia206 !

.github/workflows/build_wheels.yml
```diff
@@ -105,7 +105,7 @@ jobs:
           pip wheel --no-deps --no-build-isolation -v .
       - uses: actions/upload-artifact@v4
         with:
-          name: bodosql-wheel-no-arch
+          name: cibw-bodosql
```
IsaacWarren, line 108: I think it makes more sense to leave the name as it was and change the filter path for the download of the pypi upload since these wheels aren't built with CI build wheel

IsaacWarren (approved):
One comment but up to you, thanks Hadia

.github/workflows/build_wheels.yml
hadia206: oh, I see. I followed what was done for Iceberg, thought both were done the same way.
In this will keep original name but what should be the pattern ?

PR #147, opened 2025-01-15
Unbox dataframes eagerly to reduce compilation time

ehsantn:

Adds a flag to unbox dataframe columns eagerly and disable lazy unboxing calls (set to True by default). This reduces compilation time by ~10% for the NYC Taxi benchmark code. 

Lazy unboxing was developed previously since passing string columns around was very expensive. However, Arrow support brings zero-copy boxing/unboxing for strings and other arrays, which eliminates the need practically. Some uncommon types like IntervalArray are not supported in Arrow but they are not as important. We need to use Arrow for our datetime array type though.

In addition, spawn mode scatters input dataframes to workers, which makes sure the input dataframe to JIT is created by Bodo and allows zero-copy.

## Testing strategy

CI and nightly tests.

## User facing changes

None.

## Checklist


ehsantn:
Nightly has errors but they seem to match main.

bodo/hiframes/boxing.py
```diff
@@ -80,6 +80,12 @@
 # the number of dataframe columns above which we use table format in unboxing
 TABLE_FORMAT_THRESHOLD = 0
 
+
+# Unbox dataframe columns eagerly, which improves compilation time by disabling lazy
+# unboxing calls (only for table format since non-table format is deprecated).
+UNBOX_DATAFRAME_EAGERLY = True
```
IsaacWarren, line 86: If this is enabled by default can we remove the old lazy unboxing code? It'll still be in the history if we need to turn it back on again

IsaacWarren (approved):
Thanks Ehsan

DrTodd13 (approved):

PR #173, opened 2025-01-23
BSE-4502: Fix default str args

IsaacWarren:

Scipy upgrade added string default arg to factorial which broke object mode overload
## Testing strategy
PR CI

## User facing changes
Scipy factorial usable in jit functions

## Checklist

### DIFF
--- a/bodo/utils/typing.py
+++ b/bodo/utils/typing.py
@@ -2,6 +2,7 @@
 Helper functions to enable typing.
 """
 
+import builtins
 import copy
 import itertools
 import operator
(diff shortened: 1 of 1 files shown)


ehsantn (approved):
Looks good to me.

bodo/utils/typing.py
```diff
@@ -2273,14 +2274,22 @@ def _gen_objmode_overload(
 
         # Matplotlib specifies some arguments as `<deprecated parameter>`.
         # We can't support them, and it breaks our infrastructure, so omit them.
+        #
+        def get_default(default_val):
+            default_val_type = type(default_val)
```
srilman, line 2279: Why do it like this instead of isinstance? Not that its wrong or anything, just seems odd since accessing `type` is generally not recommended

srilman (approved):

bodo/utils/typing.py
IsaacWarren: Switched to use a class pattern match, didn't know about it before

PR #175, opened 2025-01-24
Support reading CSV and Parquet from HuggingFace

ehsantn:

As title. Uses `huggingface_hub` which is fsspec compatible.

## Testing strategy

Added unit tests.

## User facing changes

`pd.read_csv("hf://...")` and `pd.read_parquet("hf://...")` work now.

## Checklist


ehsantn:
Parquet support was simple so I decided to include it in the same PR.

bodo/io/_fs_io.cpp
```diff
@@ -18,6 +18,11 @@
 #include "_fs_io.h"
 #include "arrow_compat.h"
 
+// Forward declaration from pyfs.pyx
+struct c_PyFileSystemBodo;
```
IsaacWarren, line 22: Do we do this where we use get_cpp_fs in other files?

IsaacWarren (approved):
Thanks Ehsan, 2 questions but LGTM

pixi.toml
```diff
@@ -95,6 +95,7 @@ botocore = "*"
 s3fs = ">=2022.1"
 adlfs = ">=2022.1"
 gcsfs = ">=2022.1"
+huggingface_hub = "*"
```
srilman, line 98: Should we add this to the pyproject.toml as an optional, so `pip install bodo[huggingface]` works? Or something like that

bodo/io/_fs_io.cpp
ehsantn: We only use `get_cpp_fs` in one other file, which imports `pyfs.cpp`. However, importing it again here (or even `pyfs.h`) causes compilation issues with duplicate symbols.
https://github.com/bodo-ai/Bodo/blob/0bf0f5f1a2c7574e0666b7fb4ab225c8fa5af2d1/bodo/io/_fsspec_reader.cpp#L17

IsaacWarren: I think we would fix it by moving the inclusion of the cpp file to a header and wrapping it in pragma once and then both files can import the header, that is the cleaner solution IMO

docs/docs/file_io.md
```diff
@@ -654,6 +654,20 @@ docs [here](https://cloud.google.com/docs/authentication/getting-started#setting
 Bodo uses the fsspec-based [gcsfs](https://gcsfs.readthedocs.io/en/latest/){target="blank"} library internally for
 read and write of data on GCS.
 
+
+### Hugging Face Datasets
+
+Bodo supports reading CSV, JSON, and Parquet data from Hugging Face datasets using the [huggingface_hub](https://huggingface.co/docs/huggingface_hub) library
```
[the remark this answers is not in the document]
ehsantn: Removed JSON since looks like we have issues. The file system code path is the same so could be an existing problem. Will open an issue.

pixi.toml
ehsantn: Done, thanks for the suggestion.

bodo/io/_fs_io.cpp
ehsantn: This still gives duplicate symbol error since macros are only for macro expansion step of compilation and don't affect linking. We will refactor most of this code and remove this file hopefully soon anyways.

PR #178, opened 2025-01-24
Fix JSON lines read in parallel

ehsantn:

Fixes a JSON read bug that was exposed during HuggingFace effort. Using `}` separator for this dataset is problematic since the data itself has `}`. We should discuss not accepting JSON altogether if `lines=False` because of potential similar issues.

Looks like this bug was introduced 5 years ago! Previous version checked lines when setting the separator.
https://github.com/bodo-ai/Bodo/commit/b3d9e7182beab0d893ccd9985ee54f1907312318

## Testing strategy

Added a unit test.

## User facing changes

Bug fix and minor doc change.

## Checklist

### DIFF
--- a/bodo/io/_csv_json_reader.cpp
+++ b/bodo/io/_csv_json_reader.cpp
@@ -2192,7 +2192,7 @@ extern "C" PyObject *file_chunk_reader(
         // seem like something that should worry us right now
 
         char row_separator = '\n';
-        if (strcmp(suffix, "json") == 0) {
+        if ((strcmp(suffix, "json") == 0) && !json_lines) {
             row_separator = '}';
         }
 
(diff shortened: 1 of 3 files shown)


IsaacWarren (approved):
Thanks Ehsan, to make sure I understand, we split a file on the row separator to distribute chunks to different ranks? That seems very precarious for json with possiblity of things like nested data

ehsantn:
> Thanks Ehsan, to make sure I understand, we split a file on the row separator to distribute chunks to different ranks? That seems very precarious for json with possiblity of things like nested data

Yes. This is why I'm suggesting limiting our parallel JSON support to JSONL potentially.

srilman (approved):

PR #189, opened 2025-01-29
[BSE-4471] Move SF Prefetch Support to Python

srilman:

Move the special prefetching code from the Iceberg connector and Java to Python, simplifying it greatly in the process.

## Testing strategy

Tested the Snowflake-managed Iceberg tests locally, triggering on Nightly.

## User facing changes

None.

## Checklist


iceberg/bodo_iceberg_connector/iceberg-java/src/main/java/com/bodo/iceberg/catalog/NessieBuilder.java
```diff
@@ -1,29 +0,0 @@
-package com.bodo.iceberg.catalog;
```
srilman, line 1: Just removed it since we don't support Nessie anyways

iceberg/bodo_iceberg_connector/iceberg-java/src/main/java/com/bodo/iceberg/BodoIcebergHandler.java
```diff
@@ -38,44 +35,6 @@ public BodoIcebergHandler(String connStr, String catalogType, String coreSitePat
...
-  /**
-   * Start the Snowflake query to get the metadata paths for a list of Snowflake-managed Iceberg
-   * tables. NOTE: This API is exposed to Python.
-   *
-   * <p>The query is not expected to finish until the table is needed for initialization / read.
-   *
-   * @param tablePathsStr A JSON string of a list of tablePaths Py4J can't pass the direct list of
-   *     strings in
-   */
-  public static BodoIcebergHandler buildPrefetcher(
```
srilman, line 50: This only applied to Snowflake prefetch

IsaacWarren:
@srilman So after this PR we won't prefetch for anything that falls back to the java iceberg connector?

srilman:
@IsaacWarren to be more specific, prefetching is only implemented for Snowflake-managed Iceberg tables (maybe it could be implemented for others if they have an async API. I can only really imagine REST though). And I removed all support for Snowflake from the connector because of how bad it is and how much trouble it causes us. Even if someone installs the connector, it will still use Pythons catalog.

ehsantn:
> @IsaacWarren to be more specific, prefetching is only implemented for Snowflake-managed Iceberg tables (maybe it could be implemented for others if they have an async API. I can only really imagine REST though). And I removed all support for Snowflake from the connector because of how bad it is and how much trouble it causes us. Even if someone installs the connector, it will still use Pythons catalog.

@srilman could you explain "how bad it is and how much trouble it causes us"? I thought Snowflake develops the Iceberg support in the Java library, but the Python one we have is just monkey-patched from an abondaned PR.

srilman:
@ehsantn So really the main problem is prefetching code, but the base catalog is not great either. For starters, the whole catalog is built ontop of `snowflake-jdbc` which had various problems (over-logging to stdout / stderr that we can't disable for some reason, 75MB - 100MB contributing to most of the connectors size and installation issues), but its also hard to monkey-patch our special handling for things like create table and the prefetch code because its Java. Furthermore, the catalog itself is nothing more than a set of SQL calls with the Iceberg interfaces, which is why we had the confidence to wrap it for prefetching purposes. So the Python catalog is not significantly worse in that regard.

For prefetching itself, aside from the logging and testing issues we've had in the past, it doesn't really mesh well with Bodo's caching due to the internal caching done on the Iceberg Java side to avoid Py4J issues, which is why we've had a hard time testing and ensure prefetching is actually working in the past. Its much easier now with just Python, since I can use Bodo logging tools.

IsaacWarren (approved):
Thanks Srinivas, LGTM

bodo/io/iceberg/catalog/snowflake.py
```diff
@@ -320,3 +364,25 @@ def list_views(self, namespace: str | Identifier) -> list[Identifier]:
...
+            sf_identifier = SnowflakeCatalog._SnowflakeIdentifier.table_from_id(
+                table_id
+            )
+
+            with self.connection.cursor(DictCursor) as cursor:
+                cursor.execute_async(METADATA_QUERY, (sf_identifier.table_name,))
+                query_id = cursor.sfqid
+                assert query_id is not None
+
+            self.open_metadata_requests[sf_identifier.table_name] = query_id
```
ehsantn, line 388: Do we need the whole table id to avoid name conflicts in the dictionary?

iceberg/bodo_iceberg_connector/iceberg-java/src/main/java/com/bodo/iceberg/catalog/NessieBuilder.java
ehsantn: We did at some point. Any reason to remove it? Does it hurt something?

ehsantn (approved):
Looks good to me. Thanks @srilman .

bodo/io/iceberg/catalog/snowflake.py
srilman: `sf_identifier.table_name` is actually the whole table_id unescaped. Was part of the original code, dont love the naming. Something we should change if/when we open up a PR on PyIcebergs side

iceberg/bodo_iceberg_connector/iceberg-java/src/main/java/com/bodo/iceberg/catalog/NessieBuilder.java
srilman: Mainly to avoid confusion for anyone looking at the code and seeing this file. We can always get it back from Git history.

PR #98, opened 2024-12-31
Upgrade to Numba 0.61 

ehsantn:
## Changes included in this PR

As title. Numba 0.61 supports Python 3.13 and Numpy 2 and fixes some issues. Python 3.13 package isn't available on conda-forge yet though. In progress PR: https://github.com/conda-forge/numba-feedstock/pull/152

## Testing strategy

CI and nightly tests.

## User facing changes

Newer Numpy versions can be installed.

## Checklist
- [x] Pipelines passed before requesting review. To run CI you must include `[run CI]` in your commit message.
- [x] I am familiar with the [Contributing Guide](https://github.com/bodo-ai/Bodo/blob/main/CONTRIBUTING.md) 
- [x] I have installed + ran pre-commit hooks.

PR #201, opened 2025-02-03
Update Numba and Numpy versions in other places after Numba 0.61.0 upgrade

ehsantn:
Follow up on Numba 0.61 upgrade: https://github.com/bodo-ai/Bodo/pull/98


IsaacWarren (approved):
Thanks Ehsan

scott-routledge2 (approved):

PR #202, opened 2025-02-03
[BSE-4473] PyIceberg Write Support

srilman:

Rework the Iceberg write code to use PyIceberg under the hood for metadata management. Note that a lot of the write code was helper code for correctly formatting things like Iceberg Schema strings, PyArrow schemas for Parquet write, etc. PyIceberg already has helpers for all of these, so we can just delete

## Testing strategy

Ran all of the Iceberg tests locally and on CI. Will run Nightly on the main branch.

## User facing changes

Ideally none, except for better error messages and stability.

## Checklist


BodoSQL/bodosql/bodosql_types/glue_catalog.py
```diff
@@ -119,7 +119,7 @@ def unbox_glue_catalog(typ, val, c):
 @numba.jit
 def get_conn_str(warehouse):
     """Get the connection string for a Glue Iceberg catalog."""
-    return f"glue?warehouse={warehouse}"
+    return f"iceberg+glue?warehouse={warehouse}"
```
srilman, line 122: Basically by this point there is no difference between the runtime and compile time connection string. That was an annoying and unnecessary requirement anyways

bodo/__init__.py
```diff
@@ -131,7 +131,7 @@ def _global_except_hook(exctype, value, traceback):
     os.environ.get("BODO_ENABLE_STREAMING_SORT_LIMIT_OFFSET", "1") != "0"
 )
 # Flag used to enable creating theta sketches for columns when writing with Iceberg
-enable_theta_sketches = os.environ.get("BODO_ENABLE_THETA_SKETCHES", "1") != "0"
+enable_theta_sketches = os.environ.get("BODO_ENABLE_THETA_SKETCHES", "0") != "0"
```
srilman, line 134: Will undo in the next PR that adds connector support

bodo/io/iceberg/catalog/s3_tables.py
```diff
@@ -221,6 +222,14 @@ def _validate_database_and_table_identifier(
 
         return namespace, table_name
 
+    @staticmethod
+    def _write_metadata(
+        metadata, io: FileIO, metadata_path: str, overwrite: bool = False
+    ) -> None:
+        ToOutputFile.table_metadata(
+            metadata, io.new_output(metadata_path), overwrite=overwrite
+        )
```
srilman, line 231: A patched fix from the PyIceberg PR

bodo/io/iceberg/merge_into.py
```diff
@@ -270,3 +269,11 @@ def impl(
         impl = locals["impl"]
 
     return impl
+
+
+def format_iceberg_conn_njit():
+    pass
+
+
+def get_table_details_before_write():
```
srilman, line 278: To fix import issues, MERGE INTO is still broken by this PR.

bodo/io/iceberg/theta.py
IsaacWarren, line 1: I'm assuming the plan is a followup PR for this?

bodo/tests/test_iceberg/test_iceberg_write.py
```diff
@@ -0,0 +1,2194 @@
...
+
+    def impl(df, table_name, conn, db_schema):
+        df.to_sql(table_name, conn, db_schema, if_exists="fail")
+
+    orig_use_dict_str_type = bodo.hiframes.boxing._use_dict_str_type
+    if base_name == "DICT_ENCODED_STRING_TABLE":
+        bodo.hiframes.boxing._use_dict_str_type = True
+
+    try:
+        # TODO Uncomment after adding replicated write support.
```
IsaacWarren, line 82: I thought we had replicated write now

bodo/tests/test_iceberg/test_iceberg_write.py
IsaacWarren, line 1: Not reviewing since this was just split from test_iceberg.py

IsaacWarren (approved):
A couple small comments but overall LGTM, thanks @srilman!

BodoSQL/bodosql/tests/test_types/test_tabular_iceberg_catalog.py
```diff
@@ -31,6 +35,52 @@
...
+        table_written = True
+
+        yield iceberg_table_name
+    finally:
+        if table_written:
+            catalog = RestCatalog(
+                "tabular_catalog",
+                uri=f"http://api.tabular.io/ws?credential={credential}&warehouse={warehouse}",
+            )
+            catalog.purge_table(f"{schema}.{iceberg_table_name}")
```
ehsantn, line 81: Is it ok for all ranks to call this?

BodoSQL/bodosql/tests/test_types/test_tabular_iceberg_catalog.py
```diff
@@ -371,12 +420,9 @@ def read_results(con_str, schema, table_name):
...
         raise e
     finally:
+        catalog = RestCatalog("tabular_catalog", uri=rest_uri)
         try:
-            run_rank0(bic.delete_table)(
-                bodo.io.iceberg.format_iceberg_conn(con_str),
-                schema,
-                table_name,
-            )
+            catalog.purge_table(f"{schema}.{table_name}")
```
ehsantn, line 425: Same.

bodo/utils/py_objs.py
```diff
@@ -122,3 +127,73 @@ def create_struct_from_pyobject(typ, obj, context, builder, pyapi):
     )
     struct_proxy.pyobj = obj
     return struct_proxy._getvalue()
+
+
+def install_opaque_class(
```
ehsantn, line 132: This is potentially dangerous because it doesn't delete the object after use. Can you confirm that memory management is done manually for all the uses properly?

ehsantn (approved):
I'm not familiar with the details but looks good to me overall.

bodo/tests/test_iceberg/test_iceberg_write.py
srilman: Not for non-streaming. Looks like we do for streaming, but Sahil included an important comment:
```
        # Since streaming write is used only for SQL, replicated in this
        # context means actually replicated data (instead of independent sequential
        # functions with different data).
```
Not very familiar with replicated in Python, but it sounds like its possible for ranks to have different data in replicated Python mode. Not sure how that would work with write in general, since they all have to do an independent write to the table?

bodo/io/iceberg/theta.py
srilman: Yep, the next one with connector support. PyIceberg 0.8 doesn't support Puffin, 0.9 does but they're trying to get a merge into PR merged before starting the release it looks like.

BodoSQL/bodosql/tests/test_types/test_s3_tables_iceberg_catalog.py
```diff
@@ -101,18 +99,13 @@ def read_results(con_str, schema, table_name):
...
-                    bodo.io.iceberg.format_iceberg_conn(con_str),
-                    "write_namespace",
-                    table_name,
-                )
-            except Exception:
+        catalog = S3TablesCatalog(
+            "s3_tables_catalog", **{"s3tables.warehouse": s3_tables_catalog.warehouse}
+        )
+        try:
+            catalog.drop_table(f"write_namespace.{table_name}")
```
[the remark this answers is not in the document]
srilman: I think something about S3 tables made this work, maybe multiple drops at around the same time are ok? Will wrap with a run_rank0 though

BodoSQL/bodosql/tests/test_types/test_tabular_iceberg_catalog.py
srilman: See above

bodo/utils/py_objs.py
srilman: So I think i reworked it multiple times to the point where only `python_list_of_heterogeneous_tuples_type` needs to be opaque now, and I'm certain we deallocate it correctly in C++. I modified the others to use `install_py_obj_class`.

srilman: Oh my bad, also `pyarrow_fs_type` uses it, but the `catalog` object created and owns it. Since the catalog object is still around as long as the transaction is, it should be fine to not delete it in C++.

PR #205, opened 2025-02-04
[BSE-4538] Improve Buffer Pool Public Error Message for OOM

srilman:

Modify the Buffer Pool error message when not in debug mode (used for testing) to provide a better description of the issue and options to address.

```
RuntimeError: Error in Arrow Reader: Cast Out of memory: Bodo detected an out-of-memory error. Rank 9/10 requires 5.3GiB of memory (5.2GiB in use + 8.0MiB requested) but only has 1.27GiB available. To address the issue:
  - If you're using BodoSQL, consider enabling spilling to disk or object store (S3 / ABFS).
  - Consider using a larger machine or cluster with at least 50.34GiB of memory. Note, this is only an estimate based on the current state.
  - If you're running Bodo on a remote cluster, set `BODO_BUFFER_POOL_REMOTE_MODE=1` so it can use more resources without concern. Note that this may lead to OS OOMs if the cluster is under-provisioned.
```

Note that the default error message won't make much sense because the limit checker is disabled by default and the error is caused by another underlying Buffer Pool error. But I think its specific to the test I was running (a Python read test).

## Testing strategy

Mainly manual examples since we want debug mode enabled in CI.

## User facing changes

Better error message on OOMs.

## Checklist


IsaacWarren (approved):
Thanks Srinivas this looks great, one question, in the example why does it say it's allocating 8MB that fails even though there's 1.27GB available?

srilman:
@IsaacWarren actually the wording is bad. There is 1.27GB in total on the rank, not 1.27GB available.

srilman:
Will improve

hadia206 (approved):
Thanks for the improving the messages.
This is great!

srilman:
The job with the pool tests passed, so merging

PR #220, opened 2025-02-06
[BSE-4562 WINDOWS] Windows PR CI

scott-routledge2:

Add a workflow for running unit tests on windows runners. For now I was planning on making it available via a manual workflow trigger. 

## Testing strategy
Run the workflow, PR CI:
https://github.com/bodo-ai/Bodo/actions/runs/13245415394/job/36970679758?pr=220
Note that these tests will fail due to existing gaps/bugs in Windows. Merging this workflow will hopefully help us to find + fix these issues faster.

## User facing changes
No

## Checklist


.github/workflows/windows_unittests.yml
```diff
@@ -0,0 +1,56 @@
+name: Run PR CI on Windows
+on:
+  pull_request:
```
scott-routledge2, line 3: Based on my understanding of GA this is needed to test the workflow on my pull-request. 
I was thinking that this line could be removed before merging (so it doesn't run every time someone submits a PR) rather, we could use the manual trigger below. However, I wasn't seeing it on the actions tab, so does it need to be merged in order to trigger the workflow manually?

.github/workflows/_test_python_source.yml
```diff
@@ -34,16 +39,19 @@ jobs:
           with:
             build-all: true
         - name: Load Hadoop from Cache
+          if: runner.os != 'Windows'
           uses: actions/cache/restore@v4
           with:
             path: hadoop.tar.gz
             key: hadoop-3.3.2-${{ runner.os }}
         - name: Install SAS and Clean Cache
+          if: runner.os != 'Windows'
```
srilman, line 48: Cleaning cache is kind of important, we can't run that on windows?

.github/workflows/windows_unittests.yml
srilman: Why not reuse the PR CI workflow, but add a parameter to the workflow dispatch to choose between linux and windows. We can even add Linux-ARM and so on in the future that way

srilman (commented):
Looks fine, but I think we can make it much simpler

.github/workflows/_test_python_source.yml
scott-routledge2: We can't install SAS token provider on Windows. So we could in theory just break this into two steps or add a check if  windows inside the script.

.github/workflows/windows_unittests.yml
scott-routledge2: I think this is a good idea. We could even have a way of triggering it through the commit method (like [run ci win] or something.

scott-routledge2:
https://github.com/bodo-ai/Bodo/actions/runs/13271882794

.github/workflows/pr_ci.yml
```diff
@@ -2,6 +2,12 @@ name: PR CI
 on:
   pull_request:
   workflow_dispatch:
+    inputs:
+      runner_os:
+        description: 'The OS of the runners that will build and test bodo. Must be one of {"ubuntu-latest", "windows-latest"}'
+        required: false
+        type: string
+        default: 'ubuntu-latest'
```
srilman, line 10: Optional, but you can define specific options that are allowed. When you trigger the run, there will be a dropdown of all allowed options. Easier than explaining in the description. See https://github.blog/changelog/2021-11-10-github-actions-input-types-for-manual-workflows/.

```
on:
  workflow_dispatch:
    inputs:
      name:
        type: choice
        description: Who to greet
        options: 
        - monalisa
        - cschleiden
```

srilman (approved):
Looks great, just 1 small improvement. Thanks @scott-routledge2, this will help with ARM and others in the future!

.github/actions/build-source/action.yml
```diff
@@ -29,13 +29,16 @@ runs:
           role-skip-session-tagging: true
 
       - name: Build Bodo
-        run: pixi run build-bodo -Cbuild.verbose=true
+        run: |
+          if [[ $RUNNER_OS != 'Windows' ]]; then
+            # Build Bodo with our fork of Arrow
```
IsaacWarren, line 34: Shouldn't we only do this for linux x64, just in case we add linux arm/OSX

IsaacWarren (approved):
Thanks Scott

PR #221, opened 2025-02-06
Fix Pandas PR CI Trigger

srilman:

Fix a problem with PR CI where it is still running Pandas tests even when no compiler or Pandas files are modified. Turns out it was a dumb mistake mixing up the if and else parts of an if expression.

In addition, I rewrote and disabled a couple more tests from PR CI, in particular a lot of duplicate Iceberg tests that would take ~20 minutes in total and the table column deletion pass tests that would take ~15 minutes.

Finally, I updated some of the test times and added any new tests to the log. In total, PR CI seems to take 30 minutes in total with all of these changes. I plan to make some more changes, but probably during my oncall next week.

## Testing strategy

Running PR CI.

## User facing changes

None

## Checklist


bodo/tests/utils.py
```diff
@@ -3223,15 +3223,15 @@ def pytest_slow_unless_changed(features):
...
     if compiler_files_were_changed
-    else pytest.mark.pandas
+    else compose_decos((pytest.mark.slow, pytest.mark.pandas))
 )
 
 # This is for marking an entire test file
 # (pytestmark = pytest_pandas)
 pytest_pandas = [pytest.mark.pandas] + (
-    [pytest.mark.slow] if compiler_files_were_changed else []
+    [] if compiler_files_were_changed else [pytest.mark.slow]
```
srilman, line 3234: This was the main mistake why Pandas tests would run in PR CI

bodo/tests/test_table_column_del.py
```diff
@@ -24,6 +24,8 @@
 )
 from bodo.utils.utils import is_expr
 
+pytestmark = pytest.mark.slow
```
srilman, line 27: These tests take ~15 minutes in total on the PR CI machines. Considering they are compiler tests that don't change often, I marked as slow.

bodo/tests/test_sql.py
```diff
@@ -24,7 +24,7 @@
 )
 from bodo.utils.testing import ensure_clean_mysql_psql_table
 
-pytestmark = pytest.mark.sql
+pytestmark = [pytest.mark.sql, pytest.mark.slow]
```
srilman, line 27: Similarly, this feature is mostly broken and not often updated so marked as slow. If you have any suggestions for other tests to mark as slow, let me know.

bodo/tests/utils.py
```diff
@@ -3223,15 +3223,15 @@ def pytest_slow_unless_changed(features):
 # This is for use as a decorator for a single test function.
 # (@pytest_mark_pandas)
 pytest_mark_pandas = (
```
IsaacWarren, line 3225: This looks like it works now but what about a skipif? Might be more straightforward

bodo/tests/test_iceberg/test_stream_iceberg_write.py
```diff
@@ -207,14 +207,7 @@ def check_table_comment(
             )
 
 
-@pytest.mark.timeout(1000)
-@pytest.mark.parametrize("column_comments", ["test_col_comments", None])
-@pytest.mark.parametrize("table_comments", ["test_tbl_comments", "", None])
-@pytest.mark.parametrize("table_properties", [True, False])
 def test_iceberg_write_basic(
```
srilman, line 210: So originally, this test would test Iceberg stream write on all types with all combinations of column comments, table comments, and table properties. That's completely unnecessary since comments and properties are not affected by the dataframe type. So I split it into two separate tests. That made these tests 15-20 minutes faster

buildscripts/github/test_dur_bodo.json
IsaacWarren, line 1: Do you have any ideas for automating these changes?

IsaacWarren (approved):
Thanks Srinivas

bodo/tests/utils.py
srilman: Yeah probably would. My hope was to use a pytest extension like testmon that looks at coverage data to figure out what tests are affected and auto-enable / auto-skip

ehsantn (approved):
Thanks for the fixes!

PR #225, opened 2025-02-07
BSE-4481: Spark test fixture refactor

IsaacWarren:

Refactors spark test fixture to more easily allow multiple catalogs.
## Testing strategy
Running tests locally

## User facing changes
NA

## Checklist


IsaacWarren:
> This looks fine, but I wonder, do we need to use Spark for everything? Since we now have PyIceberg, we can always use that and Arrow / Pandas to do the read or write. Thus, we only really need Spark for the Filesystem tests / HadoopCatalog. Just a thought, whatever works

It's true and I'm using pyiceberg in new tests/tests I need to rewrite anyways but I don't want to have to rewrite every test that uses spark right now

BodoSQL/bodosql/tests/test_types/filesystem_test_harness.py
```diff
@@ -5,6 +5,7 @@
 import bodosql
```
scott-routledge2, line 5: Change file name?

BodoSQL/bodosql/tests/test_types/test_tabular_iceberg_catalog_ddl.py
```diff
@@ -13,7 +13,7 @@
```
scott-routledge2, line 13: Change the file name here too?

scott-routledge2 (approved):
LGTM thanks @IsaacWarren

PR #228, opened 2025-02-11
Create an int128 type for Windows

ehsantn:
## Checklist

### DIFF
--- a/bodo/libs/_bodo_common.h
+++ b/bodo/libs/_bodo_common.h
@@ -80,7 +80,62 @@ void Bodo_PyErr_SetString(PyObject* type, const char* message);
 
 // --------- Windows Compatibility ------------ //
 #if defined(_WIN32)
-typedef boost::multiprecision::int128_t __int128_t;
+#include <__msvc_int128.hpp>
+
+// Subclass std::_Signed128 to add missing C++ operators
+// such as casting and conversion.
+// Avoiding error checking and exceptions to behave like a native
+// type as much as possible.
+struct __int128_t : std::_Signed128 {
+
+    __int128_t() : std::_Signed128() {}
+
+    __int128_t(std::_Signed128 in_val) : std::_Signed128(in_val) {}
+
+    template<std::integral T>
+    constexpr __int128_t(T in_val) noexcept : std::_Signed128(in_val) {}
+
+    __int128_t(float in_val) {
+        _Word[0] = in_val;
+        _Word[1] = 0;
+    }
+
+    __int128_t(double in_val) {
+        _Word[0] = in_val;
+        _Word[1] = 0;
+    }
+
+    operator float() const
+    {
+        return _Word[0];
+    }
+
+    operator double() const
+    {
+        return _Word[0];
+    }
+
+    template<std::integral T>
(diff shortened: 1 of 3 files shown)


bodo/libs/_bodo_common.cpp
```diff
@@ -541,6 +541,104 @@ std::unique_ptr<Schema> Schema::Project(
...
+            _Word[0] = 0ULL;
+            _Word[1] = static_cast<int64_t>(0x8000000000000000ULL);
+        }
+        return;
+    }
+
+    // Divide the long double by 2^64 to get the "high" part.
+    // floorl(...) ensures we only keep the integer part.
+    static const double TWO_POW_64 = std::ldexpl((double)1.0, 64);
+    ;
```
IsaacWarren, line 583:
```suggestion

```

bodo/libs/_bodo_common.cpp
```diff
@@ -541,6 +541,104 @@ std::unique_ptr<Schema> Schema::Project(
 
 // ---------------------------------------------------------------------------
 
+#if defined(_WIN32)
+
+// Constructor for float/double to int128 conversion
+template <FloatOrDouble T>
+__int128_t::__int128_t(T in_val) {
```
IsaacWarren, line 548: Do you think we could add a couple simple unittests to compare this and the native implementation on linux/mac?

IsaacWarren (approved):
Thanks Ehsan

bodo/libs/_bodo_common.cpp
ehsantn: Done.

ehsantn:
@IsaacWarren could you look at this commit too? I think this could be a critical change.
https://github.com/bodo-ai/Bodo/pull/228/commits/444bb084d7a4de47eb89b0d7e2275b8887723661

bodo/libs/_bodo_common.cpp
ehsantn: Tested my changes manually (in regular Visual Studio). Windows is so slow that adding C++ is not worth the effort right now honestly.

IsaacWarren:
> @IsaacWarren could you look at this commit too? I think this could be a critical change. [444bb08](https://github.com/bodo-ai/Bodo/commit/444bb084d7a4de47eb89b0d7e2275b8887723661)

Took, a look, it makes sense as far as I can tell, not sure why wouldn't do it that way

scott-routledge2 (approved):
LGTM, but we should either run nightly or make sure we run a test that takes the decimal MPI reduce codepath since that may not be part of PR CI.

scott-routledge2:
For instance we can try `mpiexec -n 2 pytest -svW ignore bodo/tests/test_series_part2.py::test_series_argmin_max` and make sure it works for the decimal case

ehsantn:
> For instance we can try `mpiexec -n 2 pytest -svW ignore bodo/tests/test_series_part2.py::test_series_argmin_max` and make sure it works for the decimal case

`test_series_part2.py::test_series_max` works but test_series_argmin_max doesn't. I'll investigate.

PR #235, opened 2025-02-12
[BSE-4458] Refactor Iceberg IO to Use PyIceberg Under the Hood

srilman:

Refactor our Iceberg IO to support PyIceberg as a backend. Replaces a lot of metadata-related operations to use PyIceberg instead, delete any helper code that PyIceberg already does, and limit the use of the Iceberg connector.

## Testing strategy

PR Ci and Nightly both

## User facing changes

Users no longer need to install the Iceberg connector for Iceberg support. Instead, they need to either install `bodo[iceberg]` or `bodo` and `pyiceberg`.

## Checklist


srilman:
Nightly: https://dev.azure.com/bodo-inc/Bodo/_build?definitionId=5&_a=summary

srilman:
Nightly Run that worked: https://dev.azure.com/bodo-inc/Bodo/_build/results?buildId=25556&view=results

iceberg/bodo_iceberg_connector/iceberg-java/src/main/java/com/bodo/iceberg/catalog/CatalogCreator.java
```diff
@@ -83,24 +79,14 @@ public static Catalog create(String connStr, String catalogType, String coreSite
...
     }
 
-    // CachingCatalog is a simple map between table names and their `Table` object
-    // Does not modify how the Table object works in any way
-    // Can be invalidated by CREATE / REPLACE ops or a timer (if time passed in as
-    // arg)
-    // Benefits: Potentially some speed up in collecting repeated metadata like
-    // schema
-    // Downsides: Does not refresh if another program modifies the Catalog
-    return CachingCatalog.wrap(catalog);
```
srilman, line 103: Removed the CachingCatalog change because it was causing issues when jumping between PyIceberg and Iceberg-Java. This shouldn't be as much as a problem anymore because we use some caching on the PyIceberg side, which is still the predominant caller for getting metadata. Iceberg Java is only used for more heavy handed stuff like commits.

iceberg/bodo_iceberg_connector/iceberg-java/src/main/java/com/bodo/iceberg/catalog/CatalogCreator.java
```diff
@@ -54,9 +53,9 @@ public static Catalog create(String connStr, String catalogType, String coreSite
     final Catalog catalog;
 
     // S3Tables doesn't use a URI
-    if (connStr.startsWith("arn:aws:s3tables") && catalogType.equals("s3tables")) {
-      catalog = S3TablesBuilder.create(connStr);
-      return CachingCatalog.wrap(catalog);
+    if (connStr.startsWith("iceberg+arn:aws:s3tables") && catalogType.equals("s3tables")) {
+      catalog = S3TablesBuilder.create(connStr.replaceFirst("^iceberg\\+", ""));
```
srilman, line 57: Was a bug, I removed the part where we strip the prefix during compilation.

bodo/io/iceberg/catalog/dir.py
```diff
@@ -0,0 +1,291 @@
...
+        if WAREHOUSE_LOCATION not in self.properties:
+            raise ValueError(f"Missing {WAREHOUSE_LOCATION} property")
+
+    @property
+    def warehouse_path(self) -> str:
+        return self.properties[WAREHOUSE_LOCATION]
+
+    def _table_path(self, identifier: Identifier) -> str:
+        wh_path = self.warehouse_path.removesuffix("/")
+        return f"{wh_path}/{'/'.join(identifier)}"
```
srilman, line 47: Bug here to remove cases where `{warehouse_path}//...`

bodo/io/iceberg/file_io.py
```diff
@@ -0,0 +1,83 @@
+import os
```
srilman, line 1: Added a FileIO wrapper to call into PyArrow's AzureFileSystem

bodo/io/iceberg/common.py
```diff
@@ -196,93 +172,30 @@ class IcebergConnectionType(types.Type):
...
-        session_params = conn_contents.pop("session_parameters", {})
-        conn_contents.update(session_params)
-        # Remove Snowflake Specific Parameters
-        conn_contents.pop("warehouse", None)
-        conn_contents.pop("database", None)
-        conn_contents.pop("schema", None)
-        conn_str = (
-            f"snowflake://{account}.snowflakecomputing.com/?{urlencode(conn_contents)}"
-        )
+def _fs_from_file_path(file_path: str, io: FileIO) -> FileSystem:
```
srilman, line 181: Mostly a duplicate of a internal function from PyIceberg, but slightly modified to take use of our monkey-patch

bodo/io/iceberg/monkey_patch.py
```diff
@@ -0,0 +1,79 @@
+import pyiceberg.io
```
srilman, line 1: Sadly I needed to monkey patch a few things from PyIceberg to fix various issues. See below

bodo/io/iceberg/monkey_patch.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def build_position_accessors(
+    schema_or_type: Schema | IcebergType,
+) -> dict[int, Accessor]:
+    """
+    Monkey-patched version of pyiceberg.schema.build_position_accessors to
+    use the new _BuildPositionAccessors class.
+    """
+    return visit(schema_or_type, _BuildPositionAccessors())
```
srilman, line 72: Fix to apply filters to top-level struct fields. Right now, PyIceberg only works on inner struct fields

bodo/io/iceberg/monkey_patch.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+# Monkey-patch the original function
+pyiceberg.schema.build_position_accessors = build_position_accessors
+
+# Moneky-patch inferred IO to support Bodo's
+pyiceberg.io.pyarrow.PyArrowFileIO = BodoPyArrowFileIO
```
srilman, line 79: Hook to apply our AzureFileSystem changes to PyIceberg. I need to do this here because there are places where PyIceberg just calls `PyArrowFileIO` directly instead of using their own hooks for custom FileIO.

bodo/io/iceberg/read_compilation.py
```diff
@@ -236,3 +209,106 @@ def determine_str_as_dict_columns(
...
+    detect_dict_cols: bool = False,
+    is_merge_into_cow: bool = False,
+) -> tuple[
+    # Original Schemas
+    list[str],
+    list[types.ArrayCompatible],
+    pa.Schema,
+    # Runtime Schema
+    list[str],
+    list[types.ArrayCompatible],
```
srilman, line 251: I messed up this part initially which caused some schema evolution tests to fail. Basically, this function should return the original schema info (names, types, PyArrow schema) and schema after apply BodoSQL column pruning.

bodo/ir/iceberg_ext.py
```diff
@@ -875,16 +861,35 @@ def get_filters_pyobject(filter_str, vars):  # pragma: no cover
...
+    from pyiceberg.expressions.literals import DateLiteral, TimestampLiteral
+    from pyiceberg.expressions.literals import literal as inner_literal
+
+    # PyIceberg literal doesn't support all data types
+    # See https://github.com/apache/iceberg-python/issues/1456
+    if isinstance(val, datetime.date):
+        return DateLiteral((val - datetime.date(1970, 1, 1)).days)
+    if isinstance(val, datetime.datetime):
+        return TimestampLiteral((val - datetime.datetime(1970, 1, 1)).microseconds)
+    if isinstance(val, (list, pd.core.arrays.ExtensionArray)):
```
srilman, line 883: Had to add `pd.core.arrays.ExtensionArray`

ehsantn (approved):
Thanks for the huge effort @srilman. I just skimmed through to see if anything stands out since most of the code is already reviewed.

bodo/io/iceberg/catalog/dir.py
srilman: To clarify it was a bug in my own code in the base read PR that was discovered when running nightly.

bodo/io/iceberg/monkey_patch.py
srilman: Makes sense, I'm going to leave it like this for the time being because I will have to make some large changes when PyIceberg 0.9 comes out anyways

bodo/ir/iceberg_ext.py
srilman: I'm not sure exactly why, just that sometimes Pandas arrays are used instead of lists especially when coming from BodoSQL

bodo/io/fs_io.py
```diff
@@ -901,54 +899,10 @@ def overload_get_storage_options_pyobject(storage_options):
...
-@overload(arrow_filesystem_del, jit_options={"cache": True})
-def overload_arrow_filesystem_del(fs_instance):
-    """Delete ArrowFs instance"""
-
-    def impl(fs_instance):
-        return _arrow_filesystem_del(fs_instance)
-
-    return impl
+this_module = sys.modules[__name__]
+PyArrowFSType, pyarrow_fs_type = install_opaque_class(
```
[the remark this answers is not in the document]
srilman: Yes I believe so, because it is something that we don't explicitly create a reference to, and the transaction that we use later on in the function hold a reference to it

bodo/io/parquet_pio.py
```diff
@@ -470,6 +470,51 @@ def norm_path(p: str) -> str:
...
+
+    Returns:
+        tuple[str | list[str], str]: Filepath(s) without the prefix
+            and the prefix itself.
+    """
+    parsed_url = urlparse(fpath)
+    protocol = parsed_url.scheme
+
+    if (
+        protocol in {"abfs", "abfss", "wasb", "wasbs"} and bodo.enable_azure_fs
```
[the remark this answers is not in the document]
srilman: Yes I think I need to merge with the base branch

bodo/io/iceberg/write.py
```diff
@@ -1127,109 +848,77 @@ def iceberg_write(
...
         iceberg_schema_str,
         is_parallel,
-        output_pyarrow_schema,
+        output_schema,
         fs,
         dummy_theta_sketch,
         bucket_region,
+        properties,
     )
-    arrow_filesystem_del(fs)
```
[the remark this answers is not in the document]
srilman: No because before it used to be a C++ Arrow filesystem pointer that we carried around in numba requiring an explicit delete, now its a Python object

bodo/ir/iceberg_ext.py
ehsantn: The list/array difference is fine, but my point is that we may need other nested types (struct/map) here too. A TODO may be enough for now.

BodoSQL/bodosql/tests/test_iceberg_merge_into/test_iceberg_merge_into_bodo_datatypes.py
```diff
@@ -13,7 +13,12 @@
 )
 from bodo.tests.utils import check_func
 
-pytestmark = pytest.mark.iceberg
+pytestmark = [
+    pytest.mark.iceberg,
+    pytest.mark.skip(
+        reason="[BSE-4569] MERGE INTO with PyIceberg is not supported yet"
+    ),
```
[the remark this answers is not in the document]
srilman: Yes for the time being

bodo/tests/test_iceberg/test_iceberg_metrics.py
```diff
@@ -24,7 +24,7 @@
 )
 from bodo.tests.utils import pytest_mark_one_rank
 
-pytestmark = pytest.mark.iceberg
+pytestmark = [pytest.mark.iceberg, pytest.mark.skip]
```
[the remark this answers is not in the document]
srilman: Oh just missed that, undid it

PR #239, opened 2025-02-13
Move dist_reduce to cpp file

ehsantn:
Move the function to cpp file to avoid recompiling a bunch of files in each development iteration.


IsaacWarren (approved):

scott-routledge2 (approved):
Thanks @ehsantn

PR #249, opened 2025-02-14
[Windows Int128 Compat] Decimal to String, Float to Decimal Casts

scott-routledge2:
## Checklist

### DIFF
--- a/bodo/libs/decimal_arr_ext.py
+++ b/bodo/libs/decimal_arr_ext.py
@@ -318,58 +318,6 @@ def decimal_max(lhs, rhs):
         return impl
 
 
-def decimal_to_str_codegen(context, builder, signature, args, scale):
-    (val,) = args
-    scale = context.get_constant(types.int32, scale)
-    remove_trailing_zeros = context.get_constant(types.bool_, 1)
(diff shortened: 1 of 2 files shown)


bodo/libs/decimal_arr_ext.py
```diff
@@ -318,58 +318,6 @@ def impl(lhs, rhs):  # pragma: no cover
         return impl
 
 
-def decimal_to_str_codegen(context, builder, signature, args, scale):
```
scott-routledge2, line 321: This seems to just be copied in `_decimal_scalar_to_str`. I don't see any reason why we need to keep both versions?

ehsantn (approved):
Looks good to me.

bodo/libs/decimal_arr_ext.py
IsaacWarren: Is one for arrays and one for scalars?

IsaacWarren (approved):
Thanks Scott

bodo/libs/decimal_arr_ext.py
scott-routledge2: They were both for scalars AFAICT. They both call the C++ impl `decimal_to_str`

PR #250, opened 2025-02-14
Enable HDF5 on Windows

ehsantn:

Fixes HDF5 support on Windows.

## Testing strategy

HDF5 tests are now passing on Windows.

## User facing changes

None.

## Checklist

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -127,6 +127,9 @@ hdf5 = { version = ">=1.14.3,<1.14.4.0a0", build = "*mpi_mpich_*" }
 [feature.rcpp.target.linux.dependencies]
 hdf5 = { version = ">=1.14.3,<1.14.4.0a0", build = "*mpi_mpich_*" }
 
+[feature.rcpp.target.win.dependencies]
+hdf5 = { version = ">=1.14.3,<1.14.4.0a0", build = "*mpi_impi_*" }
+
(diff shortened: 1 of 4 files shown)


pixi.toml
```diff
@@ -244,7 +247,7 @@ mpich = { version = "4.1.*", build = "h*" }
 mpich = { version = "4.1.*", build = "h*" }
 
 [feature.local-mpi.target.win.dependencies]
-impi-devel = { version = "=2021.13.1" }
+impi-devel = "*"
```
ehsantn, line 250: HDF5 needs newer impi-devel. Seems like this pinning isn't necessary.

IsaacWarren: We pinned it to match the platform and be the right MPI version for mpi4py if I remember correctly

ehsantn: Looks like our vendored mpi4py works with this newer version so I assume there is no problem. Packages can be too conservative sometimes. MPI APIs are very stable generally.

scott-routledge2 (approved):
LGTM, thanks @ehsantn

pixi.toml
IsaacWarren: Ok, as long as it's dynamically linking it should be fine, I think that's where we had the issue

IsaacWarren (approved):

PR #255, opened 2025-02-17
Fix decimal __str__

scott-routledge2:
Change our  `__str__` decimal overload to remove trailing zeros while keeping the Snowflake behavior for `decimal_scalar_to_str`

## Testing strategy
Running: 
`bodo/tests/test_series_part1.py::test_series_astype_str[decimal]`

## Checklist

### DIFF
--- a/bodo/libs/decimal_arr_ext.py
+++ b/bodo/libs/decimal_arr_ext.py
@@ -520,7 +520,6 @@ def _decimal_scalar_to_str(typingctx, arr_t, remove_trailing_zeros_t):
     def codegen(context, builder, signature, args):
         (val, remove_trailing_zeros) = args
         scale = context.get_constant(types.int32, arr_t.scale)
-        # remove_trailing_zeros = context.get_constant(types.bool_, 0)
 
         uni_str = cgutils.create_struct_proxy(types.unicode_type)(context, builder)
         in_low, in_high = _ll_get_int128_low_high(builder, val)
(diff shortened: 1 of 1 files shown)


srilman (approved):

ehsantn (approved):

bodo/libs/decimal_arr_ext.py
```diff
@@ -319,7 +319,9 @@ def impl(lhs, rhs):  # pragma: no cover
 
 
 @intrinsic(prefer_literal=True)
-def _str_to_decimal_scalar(typingctx, val, precision_tp, scale_tp):
+def _str_to_decimal_scalar(
+    typingctx, val, precision_tp, scale_tp, remove_trailing_zeros=False
+):
```
ehsantn, line 324: Setting default values for intrinsics doesn't work (unlike overloads). All calls to this has to be updated to pass the the new parameter.

ehsantn (changes requested):

bodo/libs/decimal_arr_ext.py
ehsantn: Unless if Numba supports it now.

scott-routledge2: I actually did update everywhere we use the intrinsic to specify the correct value of remove_trailing_zeroes so I will just remove the default value here just to be safe.

PR #269, opened 2025-02-18
Fix test_reduce_budget hang on Windows

ehsantn:
As title. The `cerr` print in this test causes a hang with MPI on Git Bash for some reason.

### DIFF
--- a/bodo/tests/test_memory_budget.cpp
+++ b/bodo/tests/test_memory_budget.cpp
@@ -101,9 +101,13 @@ bodo::tests::suite memory_budget_tests([] {
         comptroller->ReduceOperatorBudget(0, 50);
         bodo::tests::check(comptroller->GetOperatorBudget(0) == 50);
 
+// Avoid running this test on Windows because the cerr print causes a hang with
+// MPI on Git Bash for some reason.
+#ifndef _WIN32
         // Check that increasing the budget via ReduceOperatorBudget is illegal
         comptroller->ReduceOperatorBudget(0, 75);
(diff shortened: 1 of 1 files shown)


IsaacWarren (approved):

bodo/tests/test_memory_budget.cpp
```diff
@@ -101,9 +101,13 @@ bodo::tests::suite memory_budget_tests([] {
         comptroller->ReduceOperatorBudget(0, 50);
         bodo::tests::check(comptroller->GetOperatorBudget(0) == 50);
 
+// Avoid running this test on Windows because the cerr print causes a hang with
+// MPI on Git Bash for some reason.
+#ifndef _WIN32
         // Check that increasing the budget via ReduceOperatorBudget is illegal
         comptroller->ReduceOperatorBudget(0, 75);
         bodo::tests::check(comptroller->GetOperatorBudget(0) == 50);
+#endif
```
scott-routledge2, line 110: So any check failed in windows causes a hang? I think we should just change the check function to not write to std err on windows? Could we just pass std out to [check](https://github.com/bodo-ai/Bodo/blob/e94d4f8fc1c4a4e42aff82fc2e4eb7a536dec542/bodo/tests/test_framework.cpp#L214)?

ehsantn: This cerr print is inside `ReduceOperatorBudget` so I just turned this section of the test off.

scott-routledge2 (approved):

PR #271, opened 2025-02-18
Avoid non-portable shuffle checks on Windows

ehsantn:
Removes the assumption that certain number of inflight send/recv operations which platform/implementation dependent and doesn't pass on Windows with Intel MPI.


IsaacWarren (approved):

scott-routledge2 (approved):
Thanks @ehsantn

PR #279, opened 2025-02-19
Fix caching test runners on Windows

ehsantn:
Moves the "loose" code in caching test runners inside `if __name__ == "__main__":` since Windows subprocesses run the main module which results in recursive process creation.


IsaacWarren (approved):

scott-routledge2 (approved):
Thanks @ehsantn

PR #282, opened 2025-02-20
[BSE-4582] Bodo package for Windows Pip

scott-routledge2:

Adds pip package for windows to our build_wheels workflow and various minor fixes/patches that enable the pip package to work successfully.

Follow-ups:
- Use Ninja everywhere on Windows https://bodo.atlassian.net/browse/BSE-4599
- Use consistent version of mpi4py on all platforms https://bodo.atlassian.net/browse/BSE-4600

## Testing strategy

Running the workflow with a basic test that is used in other pip. More thorough testing of the wheel can be done in a follow up once the other issues with Windows stabilize a bit (e.g. pip install bodo and run all examples, testing libs are included are only strictly necessary, etc.).

Relevant workflow run: https://github.com/bodo-ai/Bodo/actions/runs/13531555253/job/37814727994

## User facing changes
`pip install bodo` will work on Windows

## Checklist


pyproject.toml
```diff
@@ -236,6 +240,12 @@ for file in $DYLIBS_DIR/*; do
 done
 """
 
+[tool.cibuildwheel.windows]
+# Can't use pytest because we don't include tests in the wheel
+test-command = """
+set BODO_NUM_WORKERS=2 && python {project}\\buildscripts\\bodo\\pip\\windows\\test.py
```
scott-routledge2, line 246: multiline commands are kind of weird on command prompt so opted to move to its own test file.

pyproject.toml
```diff
@@ -7,7 +7,8 @@ requires = [
     # Build Libraries
     "numpy>=1.24,<2.2",
     "pyarrow==18.1.0",
-    "mpi4py>=3.1,<3.2",
+    "mpi4py~=4.0; sys_platform == 'win32'",
```
scott-routledge2, line 10: using mpi4py ~=4.0 since there were compilation issues with 3.1. 4.0 is also nice since it locates impi DLLs and mpiexec.exe for us

.github/workflows/_build_bodo_pip.yml
```diff
@@ -80,7 +80,13 @@ jobs:
               tar -xJf boost_1_85_0.tar.xz -C .
               (cd boost-1.85.0; ./bootstrap.sh --with-libraries=container,json,stacktrace; ./b2 install)
             fi
-          CIBW_BEFORE_BUILD: |
+          # delvewheel is still relatively new and not installed by default
+          # https://cibuildwheel.pypa.io/en/stable/options/#repair-wheel-command
+          CIBW_BEFORE_BUILD_WINDOWS: >
```
scott-routledge2, line 85: For some reason CIBW doesn't support multiline commands

.github/workflows/_build_bodo_pip.yml
```diff
@@ -121,6 +135,11 @@ jobs:
             AWS_SECRET_ACCESS_KEY
             AWS_SESSION_TOKEN
             AWS_DEFAULT_REGION
+          CIBW_REPAIR_WHEEL_COMMAND_WINDOWS: >
+            sccache --show-stats &&
```
scott-routledge2, line 139: Sccache doesn't appear to be working properly, there is some discussion about this here https://gitlab.kitware.com/cmake/cmake/-/issues/22529 However we are not using /Zi or /FS in our release build AFAICT.

bodo/libs/_bodo_common.h
```diff
@@ -8,7 +8,6 @@
 #endif
 
 #include <Python.h>
-#include <boost/multiprecision/cpp_int.hpp>
```
scott-routledge2, line 11: We don't need this header since we are not using boost for int128 anywhere

.github/workflows/_build_bodo_pip.yml
```diff
@@ -93,7 +99,15 @@ jobs:
...
             MPICH_VERSION=$(/opt/python/cp312-cp312/bin/python -c 'import tomllib; print(tomllib.load(open("pyproject.toml", "rb"))["tool"]["scikit-build"]["cmake"]["define"]["MPICH_VERSION"])')
             pip install mpich=="$MPICH_VERSION" -i https://pypi.anaconda.org/mpi4py/simple --default-timeout=100
-          CIBW_ENVIRONMENT: >-
+          CIBW_ENVIRONMENT_WINDOWS: >
+            DISABLE_CCACHE=1
+            SCCACHE_BUCKET=engine-codebuild-cache
+            SCCACHE_REGION=us-east-2
+            SCCACHE_S3_USE_SSL=true
+            SCCACHE_S3_SERVER_SIDE_ENCRYPTION=true
+            BODO_WINDOWS_BUILD_TYPE="Release"
```
srilman, line 108: Why? If you're running `pip wheel` it should default to release by the config in `pyproject.toml`

CMakeLists.txt
```diff
@@ -353,6 +361,13 @@ if(NOT EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/bodo/mpi4py")
     OUTPUT_STRIP_TRAILING_WHITESPACE
   )
   file(COPY ${MPI4PY_PACKAGE_DIR} DESTINATION "${CMAKE_CURRENT_SOURCE_DIR}/bodo/")
+  
+  # Replaces module path mpi4py.MPI with bodo.mpi4py.MPI
+  execute_process(
```
srilman, line 366: It looks like we're running this for all systems, not just Windows? Also why are we running it inside of the CMake? This seems kind of dangerous.

srilman (commented):
Overall looks fine, but I have a couple of concerns.

.github/workflows/_build_bodo_pip.yml
scott-routledge2: It does but this doesn't set CMAKE_BUILD_TYPE, which we use in a bunch of places. https://cmake.org/cmake/help/v3.0/variable/CMAKE_BUILD_TYPE.html

CMakeLists.txt
scott-routledge2: Good point, this patch is specific to the Windows pip environment so it doesn't do anything for Mac/Linux or even Windows dev but I guess it's not good to rely on this behavior. I was thinking we could move it to the repairing the wheel step?

.github/workflows/_build_bodo_pip.yml
```diff
@@ -80,7 +80,13 @@ jobs:
               tar -xJf boost_1_85_0.tar.xz -C .
               (cd boost-1.85.0; ./bootstrap.sh --with-libraries=container,json,stacktrace; ./b2 install)
             fi
-          CIBW_BEFORE_BUILD: |
+          # delvewheel is still relatively new and not installed by default
+          # https://cibuildwheel.pypa.io/en/stable/options/#repair-wheel-command
+          CIBW_BEFORE_BUILD_WINDOWS: >
+            pip install delvewheel &&
```
IsaacWarren, line 86: Make this a build dependency on windows?

.github/workflows/_build_bodo_pip.yml
IsaacWarren: It should? Is there a different variable we should be checking? Is this a windows only issue?

CMakeLists.txt
IsaacWarren: Repair would be the ideal place for this IMO

pyproject.toml
IsaacWarren: Is there a reason not to use mpi4py 4 everywhere? This being an inconsistent version is worrying IMO, if APIs change or we use a feature only in one version but not the other

pyproject.toml
IsaacWarren: Have you manually checked the libraries in the output wheel?

IsaacWarren (commented):
Are we OK with publishing this wheel? Right now I think it will be published on next release, if not we need to change the release step to exclude the windows wheel.

Overall LGTM but have a couple comments, great work Scott!

.github/workflows/_build_bodo_pip.yml
scott-routledge2: It can't be a build dependency because the repair command happens outside of the build environment.

.github/workflows/_build_bodo_pip.yml
scott-routledge2: Based on my initial investigation there is quite a bit of work that needs to go into configuring the environment. Even locally using VS dev command prompt (which sets all of the paths for compiler and related programs correctly), there are some linking issues and the build type is not being properly passed. Probably worth a day or two of investigation but out of the scope of this PR for now since we have been working around it just fine so far.

IsaacWarren (approved):
Thanks Ehsan

srilman (approved):

PR #288, opened 2025-02-21
[Windows compat] Support Iceberg on Windows

ehsantn:
As title. Just a small fix in path parsing. Fixes a bunch of tests locally but there may be more PRs necessary after running nightly.

### DIFF
--- a/bodo/io/iceberg/file_io.py
+++ b/bodo/io/iceberg/file_io.py
@@ -1,5 +1,6 @@
 import os
 import typing as pt
+from pathlib import PureWindowsPath
 from urllib.parse import urlparse
 
 from pyarrow.fs import FileSystem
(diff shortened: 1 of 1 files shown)


bodo/io/iceberg/file_io.py
```diff
@@ -32,6 +50,10 @@ class BodoPyArrowFileIO(PyArrowFileIO):
     @pt.override
     def parse_location(location: str) -> tuple[str, str, str]:
         """Return the path without the scheme."""
+
+        if _is_windows_path(location):
+            return "file", "", os.path.abspath(location)
```
IsaacWarren, line 55: Maybe move this into not uri.scheme to keep "file" locations together

IsaacWarren (approved):
Small suggestion otherwise LGTM

bodo/io/iceberg/file_io.py
ehsantn: The actual problem is that "C:" will be recognized as scheme so this has to be separate.

ehsantn:
> LGTM but can you create an issue to get this fix into PyIceberg itself

Created an issue but not sure if they want to handle local files in general?

PR #290, opened 2025-02-22
[Windows compat] Fixes Matlplotlib xcorr, IO write, and some other issues on Windows

ehsantn:
As title.

### DIFF
--- a/bodo/io/_fs_io.cpp
+++ b/bodo/io/_fs_io.cpp
@@ -389,9 +389,9 @@ void open_outstream(Bodo_Fs::FsEnum fs_option, bool is_parallel,
             if (is_parallel) {
                 std::filesystem::path out_path(dirname);
                 out_path /= fname;  // append file name to output path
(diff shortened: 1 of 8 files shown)


IsaacWarren (approved):
Could we use https://en.cppreference.com/w/cpp/filesystem/path for the path conversions instead? Otherwise LGTM

IsaacWarren:
> > Could we use https://en.cppreference.com/w/cpp/filesystem/path for the path conversions instead? Otherwise LGTM
> 
> I looked at it but seems like it looks at the local machine's OS to decide on separators which is not what we want here. https://en.cppreference.com/w/cpp/filesystem/path/make_preferred

I meant something like https://en.cppreference.com/w/cpp/filesystem/path/generic_string

ehsantn:
Yes, makes sense. Added the change.

scott-routledge2 (approved):
LGTM, thanks @ehsantn

PR #293, opened 2025-02-24
[BSE-4165] Cleanup Imports and CMake for Conda-Build

srilman:

Removed the unnecessary requirements for Curl and Boost (dylib not headers) that Conda-Forge brought up. Consequently, I removed the unnecessary REST Catalog FS code. Lastly, I simplified some of the config in the CMake for conda-forge.

## Testing strategy

Conda Build CI: https://github.com/bodo-ai/Bodo/actions/runs/13662611570/job/38196988495
Pip Build CI: https://github.com/bodo-ai/Bodo/actions/runs/13660473866

## User facing changes

Ideally none.

## Checklist


CMakeLists.txt
```diff
@@ -748,9 +747,7 @@ target_link_libraries(ext PRIVATE
 )
 
 if (NOT WIN32)
-  # On Windows we do not link Boost_LIBRARIES here to avoid having multiple defnitions of everything.
-  # We use the header-only version of boost::json: https://www.boost.org/doc/libs/1_87_0/libs/json/doc/html/json/overview.html
-  target_link_libraries(ext PRIVATE pthread ${Boost_LIBRARIES})
```
srilman, line 753: If we are only using the header version of Boost, we don't need to link. Conda-Forge is complaining that it doesn't see us use the boost shared library anywhere, but CMake complains if I only use the Boost headers. Thus, I fixed it here

bodo/io/_s3_reader.cpp
```diff
@@ -109,214 +105,6 @@ void _sleep_exponential_backoff(unsigned int n, unsigned int max = 10) {
 static std::shared_ptr<arrow::fs::S3FileSystem> s3_fs;
 bool is_fs_anonymous_mode = false;  // only valid when is_fs_initialized is true
 
-std::pair<const std::string, const std::string>
-IcebergRestAwsCredentialsProvider::get_warehouse_config() {
```
srilman, line 113: Removing the old REST Catalog IO code to remove CURL as a build dependency. Again, Conda-Forge was complaining about it

bodo/libs/_bodo_common.cpp
```diff
@@ -1,7 +1,6 @@
 #include "_bodo_common.h"
 
 #include <arrow/array.h>
-#include <zstd.h>
```
srilman, line 4: Just moving zstd closer to its use

bodo/libs/_bodo_common.h
```diff
@@ -377,30 +377,32 @@ std::unique_ptr<BodoBuffer> AllocateBodoBuffer(
         bodo::default_buffer_memory_manager());
 
 constexpr inline bool is_unsigned_integer(Bodo_CTypes::CTypeEnum typ) {
-    if (typ == Bodo_CTypes::UINT8)
```
srilman, line 380: Fixed some warnings from Clang-Tidy

CMakeLists.txt
```diff
@@ -209,15 +210,13 @@ if (WIN32)
...
+else()
   add_compile_options(
     "-Wno-c99-designator"         # Check if still necessary?
     "-Wno-return-type-c-linkage"  # Check if still necessary?
     "-Wno-macro-redefined"        # Check if still necessary?
     "-Wno-implicit-const-int-float-conversion" # Necessary for DataSketches
-    "-fwrapv"                     # Conda used to force-add. Remove in follow up PR
-    "-fstack-protector-strong"    # Conda used to force-add. Remove in follow up PR
-    "-D_FORTIFY_SOURCE=2"         # Conda used to force-add. Remove in follow up PR
+    "-fwrapv"                     # Conda used to force-add. Its an optimization for integer overflow
```
srilman, line 219: The conda-forge team wants us to just use the flags that they set in their environment, instead of overriding it

CMakeLists.txt
```diff
@@ -104,13 +100,12 @@ endif()
 
 # Find additional import libraries needed to compile on Windows
 if(WIN32)
-  find_package(CURL REQUIRED)
```
srilman, line 107: We shouldn't need curl directly anymore with PyIceberg handling the REST stuff

CMakeLists.txt
```diff
@@ -104,13 +100,12 @@ endif()
...
   find_package(OpenSSL REQUIRED)
 endif()
 
 # Add boost as a dependency
+# We use the header-only version of boost::json: https://www.boost.org/doc/libs/1_87_0/libs/json/doc/html/json/overview.html
 find_package(Boost 1.85 REQUIRED COMPONENTS json)
```
srilman, line 108: We don't need to overlink against Boost this way.

srilman:
I believe the failing S3 tests are unrelated and generally broken on CI.

bodo/io/_s3_reader.cpp
IsaacWarren: Would this break write since it still goes through the bodo iceberg connector? Or does it still get the fs from the pyiceberg catalog?

IsaacWarren (approved):
Thanks @srilman, one question

bodo/io/_s3_reader.cpp
srilman: Its still from PyIceberg, the write + Iceberg-Java is used for metadata so we only need an FS in Java

scott-routledge2 (approved):
LGTM thanks @srilman !  Can we also quickly check that this compiles on windows (by running any of pipelines e.g. PR CI on windows) ?

pyproject.toml
```diff
@@ -194,8 +194,8 @@ set -exo pipefail
 BODO_NUM_WORKERS=2 python {project}/buildscripts/bodo/pip/test.py
 
 # Check that only expected libs are included in the wheel
-LIB_WHITELIST="^(libcom_err|libcrypto|libcurl|libfftw3_mpi|libfftw3f_mpi|libfftw3|libfftw3f|libgssapi_krb5|libhdf5|libk5crypto|libkeyutils|libkrb5support|libkrb5|libnghttp2|libs2n|libssh2|libssl|libsz|libzstd|libboost_container|libboost_json|libbrotli.*|libcrypt|libidn2|liblber-2|libpcre2|libpsl|libsasl2|libselinux|libssh|libunistring|libldap-2)-.*\\.so(\\.[0-9]+)*"
-LIBS_DIR="$(pip show  bodo | grep Location | awk -F 'Location: ' '{print $2}')/bodo.libs" && \
+LIB_WHITELIST="^(libcom_err|libcrypto|libcurl|libgssapi_krb5|libhdf5|libk5crypto|libkeyutils|libkrb5support|libkrb5|libnghttp2|libs2n|libssh2|libssl|libsz|libzstd|libbrotli.*|libcrypt|libidn2|liblber-2|libpcre2|libpsl|libsasl2|libselinux|libssh|libunistring|libldap-2)-.*\\.so(\\.[0-9]+)*"
```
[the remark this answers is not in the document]
srilman: I did before, but its still linked because of AWS

CMakeLists.txt
srilman: I think because conda-forge changes stuff on their end, so we can let them decide. Plus, they generally make Bodo slower, so we should avoid if possible

PR #301, opened 2025-02-25
BSE-4592: Get universe size without stdin

IsaacWarren:

As title, I'm assuming this fixes it because getting the universe size launches hydra.

## Testing strategy
New unit test that fails without this change and passes after

## User facing changes
Interactive spawn mode won't crash

## Checklist

### DIFF
--- a/bodo/tests/test_spawn/test_spawn_mode.py
+++ b/bodo/tests/test_spawn/test_spawn_mode.py
@@ -417,7 +417,7 @@ def test_spawn_globals_objmode():
 
 def test_spawn_input():
     """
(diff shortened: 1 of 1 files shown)


bodo/tests/test_spawn/test_spawn_mode.py
```diff
@@ -413,3 +413,17 @@ def f():
         return val
 
     assert f() == VALUE
+
+
+def test_spawn_input():
+    """
+    Tests that using input after spawn mode doens't fail
```
scott-routledge2, line 420:
```suggestion
    Tests that using input after spawn mode doesn't fail
```

PR #302, opened 2025-02-26
[Windows compat] Fix hanging tests and some other Windows issues

ehsantn:
As title.

### DIFF
--- a/bodo/io/snowflake.py
+++ b/bodo/io/snowflake.py
@@ -2338,6 +2338,10 @@ def do_upload_and_cleanup(
     else:
         stage_name_with_dir = f'@"{stage_name}"/{stage_dir}/'
 
+    # Windows "\" should be replaced with "/" for Snowflake PUT command when using quotes:
+    # https://docs.snowflake.com/en/sql-reference/sql/put
+    chunk_path = chunk_path.replace("\\", "/")
+
     upload_sql = (
         f"PUT 'file://{chunk_path}' {stage_name_with_dir} AUTO_COMPRESS=FALSE "
         f"/* io.snowflake.do_upload_and_cleanup() */"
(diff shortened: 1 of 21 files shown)


bodo/io/snowflake.py
```diff
@@ -2338,6 +2338,10 @@ def do_upload_and_cleanup(
     else:
         stage_name_with_dir = f'@"{stage_name}"/{stage_dir}/'
 
+    # Windows "\" should be replaced with "/" for Snowflake PUT command when using quotes:
+    # https://docs.snowflake.com/en/sql-reference/sql/put
+    chunk_path = chunk_path.replace("\\", "/")
```
IsaacWarren, line 2343: Can we use pathlib?

bodo/tests/test_snowflake_write.py
```diff
@@ -266,7 +266,7 @@ def test_impl_do_upload_and_cleanup(cursor, chunk_path, stage_name):
         # Use GET to fetch all uploaded files
         with TemporaryDirectory() as tmp_folder:
             get_stage_sql = (
-                f"GET @\"{stage_name}\" 'file://{tmp_folder}' "
+                f"GET @\"{stage_name}\" 'file://{tmp_folder.replace('\\', '/')}' "
```
IsaacWarren, line 269: Shouldn't we be able to handle windows slashes here?

IsaacWarren (approved):

scott-routledge2 (approved):

bodo/io/snowflake.py
ehsantn: Not sure since it is a URI and has "file://". I think replace is more clear here.

bodo/tests/test_snowflake_write.py
ehsantn: Snowflake docs says single quotes shouldn't have back slashes:
https://docs.snowflake.com/en/sql-reference/sql/get

PR #304, opened 2025-02-26
BSE-4590: Use disassembler in _replace_load_deref_code

IsaacWarren:

Use the builtin disassembler to modify the applied functions bytecode. Adds a function to reverse the process as well.

## Testing strategy
Ran existing freevar tests locally and added unittest for helper function.

## User facing changes
More safe applied functions with non const free variables.

## Checklist

### DIFF
--- a/bodo/transforms/typing_pass.py
+++ b/bodo/transforms/typing_pass.py
@@ -6383,7 +6383,7 @@ def _bc_stream_to_bytecode(bc_stream, out_size):
                 arg >>= 8
 
         else:
-            assert arg == None
+            assert arg is None
     return out
 
 
--- a/bodo/transforms/typing_pass.py
+++ b/bodo/transforms/typing_pass.py
@@ -14,7 +14,6 @@ import numba
 import numpy as np
 import pandas as pd
 from numba.core import event, ir, ir_utils, types
-from numba.core.bytecode import _unpack_opargs
 from numba.core.compiler_machinery import register_pass
 from numba.core.extending import register_jitable
 from numba.core.inline_closurecall import inline_closure_call
@@ -4370,7 +4369,8 @@ class TypingTransforms:
             f_ind: code.co_argcount + i for i, f_ind in enumerate(freevar_inds)
         }
         new_code = _replace_load_deref_code(
-            code.co_code, freevar_arg_map, code.co_argcount, code.co_nlocals
+            code,
+            freevar_arg_map,
         )
 
         # we can now change the IR/code since all checks are done (including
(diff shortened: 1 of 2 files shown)


IsaacWarren:
> LGTM although I am a little confused what the change that broke/exposed this was. Where some of values in the byte code greater than 256?

Yes some are greater than 256, you can look `dis.opmap` to see them all, also it was an int8 array so from -128 to 127.

bodo/transforms/typing_pass.py
```diff
@@ -6373,51 +6397,52 @@ def _replace_load_deref_code(code, freevar_arg_map, prev_argcount, prev_n_locals
     """
     import dis
 
-    from numba.core.bytecode import ARG_LEN, CODE_LEN, NO_ARG_LEN
-
-    # assuming these constants are 1 to simplify the code, very unlikely to change
-    assert CODE_LEN == 1 and ARG_LEN == 1 and NO_ARG_LEN == 1, (
```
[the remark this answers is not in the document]
IsaacWarren: They're still 1 this is future proofing though

bodo/transforms/typing_pass.py
```diff
@@ -6363,7 +6364,30 @@ def _set_updated_container(varname, update_func, updated_containers, equiv_vars)
...
+    out = bytearray(out_size)
+    for (
+        offset,
+        op,
+        arg,
+    ) in bc_stream:
+        out[offset] = op
+        if op >= dis.HAVE_ARGUMENT:
+            for i in range(ARG_LEN):
+                out[offset + CODE_LEN + i] = arg & 0xFF
```
ehsantn, line 6382: Do we need to handle `_inline_cache_entries` like Numba does? Maybe we can copy the initial bytecode in `out` since we are skipping over some things potentially?

https://github.com/numba/numba/blob/bced43d44b405a9db443b70171eb0125216770b6/numba/core/bytecode.py#L199

bodo/transforms/typing_pass.py
```diff
@@ -6363,7 +6364,30 @@ def _set_updated_container(varname, update_func, updated_containers, equiv_vars)
...
+                arg >>= 8
+
+        else:
+            assert arg == None
```
ehsantn, line 6386:
```suggestion
            assert arg is None
```

ehsantn (commented):
See comment. Also, did this fix the issue in the original example code?

bodo/transforms/typing_pass.py
```diff
@@ -6373,51 +6397,52 @@ def _replace_load_deref_code(code, freevar_arg_map, prev_argcount, prev_n_locals
...
-    banned_ops = (dis.opmap["STORE_DEREF"], dis.opmap["LOAD_CLOSURE"])
-    # local variable access to be adjusted
-    local_varname_ops = (
-        dis.opmap["LOAD_FAST"],
-        dis.opmap["STORE_FAST"],
-        dis.opmap["DELETE_FAST"],
+    prev_argcount = code.co_argcount
+    prev_n_locals = code.co_nlocals
+
+    def _patch_opargs(code, freevar_arg_map, prev_argcount, prev_n_locals):
```
DrTodd13, line 6403: Docstring?

DrTodd13 (changes requested):

bodo/transforms/typing_pass.py
IsaacWarren: I don't think we need to do anything special since we set based on the next offset but it's probably a good idea to initialize out to the initial bytecode

ehsantn (approved):
Thanks for the updates.

DrTodd13 (approved):

IsaacWarren:
> See comment. Also, did this fix the issue in the original example code?

Yeah, confirmed it works now with numpy > 2

PR #317, opened 2025-03-03
Ensure platform enviornment is compatible with conda package.

scott-routledge2:
Relaxes the matplotlib dependency from platform. Since our package that we are building for the platform constrains numpy >2.0 (not sure why) it is incompatible with the matplotlib requirement <=3.8.2. (Needs to be updated anyways)
Also removes the constraint from our meta.yaml. This is needed because if someone wants to install the bodo conda package from an environment that has matplotlib installed, they can't (either need matplotib 3.8.2 and numpy 1.2.* or numpy 2.0+ but a newer version of matplotlib which is not allowed by bodo).

## Testing strategy
On platform:
```
pixi project export conda-environment -p linux-64 -e platform conda-linux-64.lock.yml

conda env update --name test --file conda-linux-64.lock.yml

conda install --yes -c https://USERNAME:PASSWORD@bodo.jfrog.io/artifactory/api/conda/bodo.ai-platform-dev bodo==2025.3.* --no-update-deps 
```

## Checklist

### DIFF
--- a/.github/workflows/_build_bodo_conda_native.yml
+++ b/.github/workflows/_build_bodo_conda_native.yml
@@ -67,6 +67,8 @@ jobs:
           set -exo pipefail
           cd buildscripts/bodo/conda-recipe
 
+          conda --version
+
           # --no-verify because it doesn't work for Conda v2 packages
           # --no-anaconda-upload because we'll upload ourselves
           # Sccache works when the files are located in the same path.
--- a/.github/workflows/_build_bodo_conda_native.yml
+++ b/.github/workflows/_build_bodo_conda_native.yml
@@ -67,8 +67,6 @@ jobs:
           set -exo pipefail
           cd buildscripts/bodo/conda-recipe
 
-          conda --version
-
           # --no-verify because it doesn't work for Conda v2 packages
           # --no-anaconda-upload because we'll upload ourselves
           # Sccache works when the files are located in the same path.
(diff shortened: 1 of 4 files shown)


IsaacWarren (commented):
Thanks Scott, how hard would it be to just disable our matplotlib support and remove the dependency? We haven't worked on it in years. If someone needs it we can update our extension to support a recent version of matplotlib. WDYT @hadia206?

scott-routledge2:
> Thanks Scott, how hard would it be to just disable our matplotlib support and remove the dependency? We haven't worked on it in years. If someone needs it we can update our extension to support a recent version of matplotlib. WDYT @hadia206?

I am not sure about the effort, do you mean remove all references to matplotlib in our code or is there a way we can just turn off support temporarily some how? I don't really see this as a blocker since matplotlib is not emphasized in our examples and we have a warning in our docs that support is experimental. cc @ehsantn . My main concern was that someone can't have Bodo + matplotlib installed in the same conda environment. With this PR they can still use them separately so this is still an improvement IMO.

ehsantn:
Our matplotlib support inside JIT is experimental and less important now that we have spawn mode. Not worth blocking this PR and the release. We can discuss what to do with it later.

hadia206:
> Thanks Scott, how hard would it be to just disable our matplotlib support and remove the dependency? We haven't worked on it in years. If someone needs it we can update our extension to support a recent version of matplotlib. WDYT @hadia206?

I don't think we have it as a dependency on Bodo package. This is for platform so people can use it on notebook (not necessarily with Bodo). 
We don't have any examples of matplotlib. We can just remove the matplotlib user document.

hadia206 (approved):

IsaacWarren (approved):

Issue #223, opened 2025-02-07
Is there any plan to support MultiIndex DataFrames in Parquet I/O in the future?


Roger-Liang:
As the title descripted.

ehsantn:
@Roger-Liang Thanks for the feature request. This is definitely something we can look into. Could you share more about your use case and why this is important to you? Also, example code to support would be appreciated.

Roger-Liang:
Hi @ehsantn ,

Thanks for the quick response!

I frequently work with MultiIndex DataFrames where one of the levels represents dates, and I rely on the pyarrow engine to handle Parquet I/O with partitioning based on the Date column. Partitioning by date is essential for efficiently managing and querying large, time-series datasets.

Currently, my workflow requires resetting the MultiIndex to promote the Date level to a column so that I can partition the data during the write process. When reading the data back, I need to manually reconstruct the MultiIndex. This workaround not only adds extra code but also increases the risk of errors, especially as the complexity and size of the data grow.

Below is an example of my current approach:

```
# Create a sample MultiIndex DataFrame with 'Date' and 'Sensor' as index levels
arrays = [
    ['2021-01-01', '2021-01-01', '2021-01-02', '2021-01-02'],
    ['Sensor1', 'Sensor2', 'Sensor1', 'Sensor2']
]
index = pd.MultiIndex.from_arrays(arrays, names=('Date', 'Sensor'))
df = pd.DataFrame({'Value': np.random.randn(4)}, index=index)

# Reset the 'Date' level to a column for partitioning
df_reset = df.reset_index(level='Date')

# Save the DataFrame using pyarrow engine, partitioned by the 'Date' column
df_reset.to_parquet('multiindex_data', engine='pyarrow', partition_cols=['Date'])

# Read the partitioned Parquet files back into a DataFrame
df_loaded = pd.read_parquet('multiindex_data', engine='pyarrow')

# Manually reconstruct the MultiIndex by setting 'Sensor' back as an index level
df_loaded.set_index('Sensor', append=True, inplace=True)
print(df_loaded)
```
Native support for MultiIndex DataFrames in Parquet I/O would greatly simplify this process by preserving the full hierarchical index automatically—even when partitioning by Date. This enhancement would not only streamline my workflow but also improve data integrity and reduce the overhead of manual index management.

Looking forward to your thoughts on this!

ehsantn:
Thank you @Roger-Liang for the detailed example! We will look into it and prioritize soon.

PR #320, opened 2025-03-04
#223: Support MultiIndex Parquet Write

srilman:

Support writing DataFrames containing a Multi-Index index to a Parquet file. Note, this does not support partitioning by an index column; that requires a follow-up PR. First 1/2 of https://github.com/bodo-ai/Bodo/issues/223.

Also includes some refactoring to make the Index handling code consolidated to one function / file, instead of the original which was distributed between various functions and C++.

## Testing strategy

Constructed some new Parquet write tests, tested locally with 1 and 3 ranks. Will run PR CI and Nightly on IO tests to be safe.

## User facing changes

Users can now write dataframes with an index type to Parquet.

## Checklist


bodo/hiframes/pd_multi_index_ext.py
```diff
@@ -28,13 +29,24 @@
     is_overload_none,
 )
 
+IndexNameType: pt.TypeAlias = (
```
srilman, line 32: Again, for typechecking purposes, so later devs know what the names can be

bodo/hiframes/pd_multi_index_ext.py
```diff
@@ -80,7 +92,10 @@ class MultiIndexModel(models.StructModel):
     def __init__(self, dmm, fe_type):
         members = [
             ("data", types.Tuple(fe_type.array_types)),
-            ("names", types.Tuple(fe_type.names_typ)),
+            (
+                "names",
+                types.Tuple(fe_type.names_typ),
+            ),  # TODO: Use FrozenList like Pandas
```
srilman, line 98: Noticed some inconsistent behavior between Bodo and Pandas on the latest version

bodo/io/arrow.cpp
```diff
@@ -82,16 +82,12 @@ ArrowReader* snowflake_reader_init_py_entry(
...
-                          const int ri_stop, const int ri_step,
-                          const char* idx_name, const char* bucket_region,
-                          int64_t row_group_size, const char* prefix,
-                          bool convert_timedelta_to_int64, const char* tz,
-                          bool downcast_time_ns_to_us, bool create_dir,
-                          bool force_hdfs);
+                          const char* bucket_region, int64_t row_group_size,
+                          const char* prefix, bool convert_timedelta_to_int64,
+                          const char* tz, bool downcast_time_ns_to_us,
+                          bool create_dir, bool force_hdfs);
```
srilman, line 90: Happy to trim a lot of junk about index handling from this C++ function, especially since its reused for Iceberg and Snowflake that both don't care abound indexes at all

bodo/io/parquet_pio.py
```diff
@@ -1551,202 +1549,3 @@ def _get_partition_cat_dtype(dictionary):
...
-            types.int32,
-            types.voidptr,
-            types.boolean,
-            types.voidptr,
-            types.int64,
-            types.voidptr,
-            types.voidptr,  # tz
-        ),
-        codegen,
-    )
```
srilman, line 1752: Moved to `parquet_write.py`. parquet_write_table_cpp was modified

bodo/utils/conversion.py
```diff
@@ -2245,6 +2245,25 @@ def overload_index_to_array(I):
...
+    convert Index object to data array.
+    """
+    from bodo.hiframes.pd_multi_index_ext import MultiIndexType
+
+    if isinstance(I, MultiIndexType):
+        return lambda I: list(
+            bodo.hiframes.pd_index_ext.get_index_data(I)
+        )  # pragma: no cover
+
+    return lambda I: [bodo.utils.conversion.index_to_array(I)]  # pragma: no cover
```
srilman, line 2264: For MultiIndex, `index_to_array` will return a tuple of arrays while the others return an array. So I created this helper function to make the output type consistent across all types.

bodo/hiframes/pd_multi_index_ext.py
ehsantn: Technically, Index name can be any object in Pandas. Bodo's support seems generic too (see tuple example below). But it's fine to restrict it for Parquet write and throw errors there.

```python
In [7]: A = pd.Index([1, 2], name=(3,4))

In [8]: bodo.jit(spawn=False)(lambda A: A.name)(A)
/Users/ehsan/dev/Bodo/bodo/transforms/distributed_analysis.py:305: BodoWarning: No parallelism found for function '<lambda>'. Distributed diagnostics:
Setting distribution of variable 'A' to REP: Distributed analysis replicated argument 'A' (variable 'A'). Set distributed flag for 'A' if distributed partitions are passed (e.g. @bodo.jit(distributed=['A'])).

File "<ipython-input-8-8a06ad0bec01>", line 1:
bodo.jit(spawn=False)(lambda A: A.name)(A)
^
  warnings.warn(
Out[8]: (3, 4)
```

bodo/hiframes/pd_multi_index_ext.py
ehsantn: What was the inconsistent behavior you saw in this instance? I'm just curious.

bodo/io/arrow.cpp
ehsantn: Great to see!

bodo/libs/_array.cpp
```diff
@@ -497,6 +497,13 @@ table_info* arr_info_list_to_table(array_info** arrs, int64_t n_arrs) {
     return new table_info(columns);
 }
 
+// Raw pointers since called from Python
+void append_arr_info_list_to_cpp_table(table_info* table, array_info** arrs,
+                                       int64_t n_arrs) {
+    std::vector<std::shared_ptr<array_info>> columns(arrs, arrs + n_arrs);
+    table->columns.insert(table->columns.end(), columns.begin(), columns.end());
+}
```
ehsantn, line 505: The memory management of all these array pointers seems scary, but seems correct to me.

bodo/utils/conversion.py
ehsantn: But array types can be different in MultiIndex so list wouldn't work. Am I missing something? Why not tuple?

bodo/tests/test_parquet_write.py
```diff
@@ -1038,6 +1078,42 @@ def impl(df, path):
         bodo.barrier()
 
 
+def test_to_pq_multiIdx(check_write_func, memory_leak_check):
+    """Test to_parquet with MultiIndexType"""
+    arrays = [
+        ["bar", "bar", "baz", "baz", "foo", "foo", "qux", "qux"],
+        ["one", "two", "one", "two", "one", "two", "one", "two"],
```
ehsantn, line 1085: Let's test arrays with different data types.

ehsantn (commented):
Thanks for the effort @srilman. Looks good to me overall except when MultiIndex array types don't match.

bodo/hiframes/pd_multi_index_ext.py
srilman: Yes I believe it is restricted to string names right now because thats all Parquet can support. But this is just for typing in Bodo

bodo/hiframes/pd_multi_index_ext.py
srilman: The TODO. Pandas uses a FrozenList

bodo/utils/conversion.py
srilman: Hmm so the test case does use a MultiIndex with different types so it is working correctly. Looking further, I think whats happening is `bodo.hiframes.pd_index_ext.get_index_data` returns a tuple for MultiIndex so the output is either a tuple or list (with a single elem). It works fine because tuples and lists are similar enough. But you're right, to be safe I will change it to always return a tuple.

bodo/tests/test_parquet_write.py
srilman: Sure did so see above

bodo/utils/conversion.py
srilman: Oh never mind I was looking at something else.

bodo/libs/_array.cpp
srilman: Is there a better approach to this. All I want is to concatenate the dataframe table and index arrays together, ideally in Python

bodo/hiframes/pd_dataframe_ext.py
```diff
@@ -3630,152 +3630,7 @@ def pivot_impl(
...
-
-    pandas_metadata["pandas_version"] = pd.__version__
-
-    return pandas_metadata
-
-
-@overload_method(
-    DataFrameType, "to_parquet", no_unliteral=True, jit_options={"cache": True}
-)
+@overload_method(DataFrameType, "to_parquet", no_unliteral=True)
```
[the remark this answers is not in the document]
ehsantn: Ok.

bodo/hiframes/pd_multi_index_ext.py
srilman: No nothing broke. FrozenList is very similar to Tuple and nothing asserts it type. I just noticed during manual testing

bodo/libs/_array.cpp
ehsantn: I think this is fine at least for now. I just wanted to point out in case someone notices something.

bodo/hiframes/pd_multi_index_ext.py
srilman: Added

bodo/tests/test_parquet_write.py
srilman: Updated the test

bodo/utils/conversion.py
srilman: Updated

ehsantn (approved):
Thanks for the updates.

bodo/hiframes/pd_dataframe_ext.py
```diff
@@ -3887,80 +3748,17 @@ def to_parquet_overload(
...
-            # if the index name is None then we need to write just "null" to the metadata file
-            # without quotation marks(null). But if a name is provided we need to
-            # wrap the name with quotation mark to indicate it is a string
-            pandas_metadata_str = pandas_metadata_str.replace('"%s"', "%s")
-
     func_text = "def bodo_df_to_parquet(df, path, engine='auto', compression='snappy', index=None, partition_cols=None, storage_options=None, row_group_size=-1, _bodo_file_prefix='part-', _bodo_timestamp_tz=None, _is_parallel=False):\n"
 
+    # Get all column names that will be written to parquet
+    # Note, index columns are added later for the non-partition case
+    # TODO: Extend for partitioned writes as well
```
scott-routledge2, line 3755: Just wondering, what is required to write index columns in the partitioned case i.e why can't we just append index columns to the table like we do in the other case? And also what happens if partition_cols are supplied and index is True?

scott-routledge2 (approved):
LGTM, a couple comments mostly for my own understanding.

bodo/hiframes/pd_dataframe_ext.py
```diff
@@ -3835,9 +3692,12 @@ def to_parquet_overload(
         part_col_idxs = []
         for part_col_name in partition_cols:
             try:
+                # TODO: Support index columns as partition columns
                 idx = df.columns.index(part_col_name)
             except ValueError:
-                raise BodoError(f"Partition column {part_col_name} is not in dataframe")
+                raise BodoError(
+                    f"Partition column `{part_col_name}` is not in dataframe"
```
scott-routledge2, line 3699: Also was wondering if this is the error a user will see if they try and use an index as a partition column?  Could we be more explicit about what is not supported in that case?

bodo/hiframes/pd_dataframe_ext.py
```diff
@@ -3835,9 +3692,12 @@ def to_parquet_overload(
         part_col_idxs = []
         for part_col_name in partition_cols:
             try:
+                # TODO: Support index columns as partition columns
```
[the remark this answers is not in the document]
srilman: Thats actually in the followup PR, thats why I didn't bother

bodo/hiframes/pd_dataframe_ext.py
srilman: Yeah I'm not sure as well, maybe its just something we didn't do for time purposes. Again, thats my followup

PR #321, opened 2025-03-04
Update s3 tests for invalid creds

scott-routledge2:

Something changed with our test setup and now the correct credentials are being loaded (since they are being set in the test env, not sure why this worked before but my guess was some fixture somewhere was reseting them?). Adds a temp_env_override to be more explicit about using incorrect creds in the test.

## Testing strategy

Locally I am signed in with AWS so the test fails for me with the same error. I tested again after changes and it passed.

## Checklist

### DIFF
--- a/bodo/tests/test_io_errorchecking.py
+++ b/bodo/tests/test_io_errorchecking.py
@@ -10,7 +10,7 @@ import pandas as pd
 import pytest
 
 import bodo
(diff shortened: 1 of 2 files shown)


IsaacWarren (approved):
Thanks Scott

srilman (approved):

PR #323, opened 2025-03-05
Fix Jupyter output on Windows

ehsantn:

Worker output goes to console instead of Jupyter cells on Windows, since child processes don't inherit stdout/stderr file descriptors on Windows. This PR add a mechanism for sending worker output to the spawner using ZeroMQ so the prints appear in cells properly.

## Testing strategy

Tested it manually on Windows.

## User facing changes

Jupyter output works on Windows.

## Checklist

### DIFF
--- a/bodo/spawn/worker.py
+++ b/bodo/spawn/worker.py
@@ -467,6 +467,28 @@ def exec_func_handler(
     os.environ = original_env_var
 
 
+class StdoutQueue:
+    """Replacement for stdout/stderr that sends output to the spawner via a ZeroMQ socket."""
+
+    def __init__(self, out_socket, is_stderr):
+        self.out_socket = out_socket
+        self.is_stderr = is_stderr
+        self.closed = False
+
+    def write(self, msg):
+        # TODO(Ehsan): buffer output similar to ipykernel
+        self.out_socket.send_string(f"{int(self.is_stderr)}{msg}")
+
+    def flush(self):
+        if self.is_stderr:
+            sys.__stderr__.flush()
+        else:
+            sys.__stdout__.flush()
+
+    def close(self):
+        self.closed = True
+
+
 def worker_loop(
     comm_world: MPI.Intracomm, spawner_intercomm: MPI.Intercomm, logger: logging.Logger
 ):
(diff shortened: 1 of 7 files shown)


bodo/spawn/worker.py
```diff
@@ -482,6 +504,23 @@ def worker_loop(
             "Spawner and worker 0 must be on the same machine"
         )
 
+    # Send output to spawner manually if we are in Jupyter on Windows since child processes
+    # don't inherit file descriptors from the parent process.
+    out_socket = None
+    if sys.platform == "win32" and (
```
scott-routledge2, line 510: Maybe we could define this as a Variable/helper function `IS_WINDOWS_JUPYTER` for consistency, since I believe this is different than the check we are doing in other places for Windows+Jupyter (see `BodoSQL/bodosql/py4j_gateway.py:87`)

bodo/spawn/worker.py
```diff
@@ -482,6 +504,23 @@ def worker_loop(
...
+        "JPY_SESSION_NAME" in os.environ
+        or "PYDEVD_IPYTHON_COMPATIBLE_DEBUGGING" in os.environ
+    ):
+        port = spawner_intercomm.bcast(None, 0)
+
+        import zmq
+
+        context = zmq.Context()
```
scott-routledge2, line 518: Would there be a way to unit test this? Not a blocker for this PR but maybe we could set up a test where we simulate jupyter environment (setting those environment variables) and then spawning a worker, writing some messages and ensuring it is captured via capfd. Just a thought.

scott-routledge2 (approved):
LGTM, I had a couple of comments for clean up/testing considerations but they are not blockers. Thanks @ehsantn

bodo/spawn/worker.py
IsaacWarren: It would be nice if we at least get a test that makes sure the worker's output ends up in the correct io stream so we can catch any regressions.

IsaacWarren (commented):
Thanks Ehsan, looks like we can extend this to fix the platform issue too. One observation, I think this might not work in the multinode case since it's always trying to send to localhost. I do also think it's important to have a simple unittest to catch any regressions

bodo/spawn/worker.py
ehsantn: Added a helper function.

bodo/spawn/worker.py
ehsantn: Added a unit test.

ehsantn:
Jupyter on multi-node Windows setups is out of scope right now. Added a unit test.

IsaacWarren:
> Jupyter on multi-node Windows setups is out of scope right now. Added a unit test.

We'll need to do this for the platform anyways but can we atleast throw an error?

IsaacWarren (approved):
Thanks Ehsan, can we throw an error in the multi node case, otherwise LGTM

ehsantn:
Added an error for multi-node case.

PR #325, opened 2025-03-06
Broadcast spawner hostname for jupyter windows multinode check.

scott-routledge2:
As titled

## Testing strategy
Locally in a jupyter notebook it was failing because only rank 0 defined `spawner_hostname`

## Checklist


IsaacWarren (commented):
I forget, are we guaranteed rank 0 is on the same node as the spawner?

scott-routledge2:
> I forget, are we guaranteed rank 0 is on the same node as the spawner?

Yes, there is an assert above I believe that checks this?

IsaacWarren (approved):
Thanks Scott

ehsantn (approved):
Thanks for the fix!

PR #327, opened 2025-03-06
Fix Jupyter output cell on Windows

ehsantn:
`ipykernel` redirects the output of threads to the first cell. This makes sure our worker output thread's prints appear in the correct cell.


IsaacWarren (approved):
Thanks Ehsan

scott-routledge2 (approved):
Thanks @ehsantn

PR #329, opened 2025-03-06
Support BodoDataFrame.map_partitions

ehsantn:

Adds a basic `BodoDataFrame.map_partitions` function to test our current AI use cases. Just a basic internal function (no user docs) for now until we develop the new dataframe library and data-parallel APIs.

## Testing strategy

Added unit test.

## User facing changes

Basic `map_partition` function is available for testing.

## Checklist

### DIFF
--- a/bodo/pandas/frame.py
+++ b/bodo/pandas/frame.py
@@ -375,6 +375,8 @@ class BodoDataFrame(pd.DataFrame, BodoLazyWrapper):
     def map_partitions(self, func, *args, **kwargs):
         """
         Apply a function to each partition of the dataframe.
+        NOTE: this pickles the function and sends it to the workers, so globals are
+        pickled. The use of lazy data structures as globals causes issues.
         """
         return bodo.spawn.spawner.submit_func_to_workers(
             func, [], self, *args, **kwargs
--- a/bodo/pandas/frame.py
+++ b/bodo/pandas/frame.py
@@ -371,3 +371,11 @@ class BodoDataFrame(pd.DataFrame, BodoLazyWrapper):
             storage_options=storage_options,
             mode=mode,
         )
+
+    def map_partitions(self, func, *args, **kwargs):
+        """
+        Apply a function to each partition of the dataframe.
+        """
+        return bodo.spawn.spawner.submit_func_to_workers(
+            func, [], self, *args, **kwargs
+        )
--- a/bodo/spawn/worker.py
+++ b/bodo/spawn/worker.py
@@ -37,6 +37,7 @@ from bodo.spawn.utils import (
 )
 from bodo.spawn.worker_state import set_is_worker
 from bodo.utils.typing import BodoWarning
+from bodo.utils.utils import is_distributable_typ
 
 DISTRIBUTED_RETURN_HEAD_SIZE: int = 5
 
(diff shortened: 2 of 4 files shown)


ehsantn:
@IsaacWarren could you try this branch to see if it works for your case?

IsaacWarren:
If I force `prompts` to collect onto the spawner before calling map_partitions it works but if it's on the workers something is forcing collection while submitting the function

ehsantn:
@IsaacWarren  Just fixed a small bug. Could you try again?

IsaacWarren:
> @IsaacWarren Just fixed a small bug. Could you try again?

Same issue if I don't force collection

bodo/spawn/worker.py
```diff
@@ -440,13 +439,19 @@ def exec_func_handler(
...
-    if func is not None and len(func.signatures) > 0:
+    if is_dispatcher and func is not None and len(func.signatures) > 0:
         # There should only be one signature compiled for the input function
         sig = func.signatures[0]
         assert sig in func.overloads
 
         # Extract return value distribution from metadata
         is_distributed = func.overloads[sig].metadata["is_return_distributed"]
+
+    # TODO: handle other types
```
srilman, line 450: Could you add a bit more to the comment. Clarifying that it must be a compiler-representable type?

bodo/pandas/frame.py
```diff
@@ -371,3 +371,11 @@ def to_json_wrapper(
             storage_options=storage_options,
             mode=mode,
         )
+
+    def map_partitions(self, func, *args, **kwargs):
+        """
+        Apply a function to each partition of the dataframe.
```
IsaacWarren, line 377: Can we add a comment/maybe a to ensure func isn't a jit function. Also something about the globals issue I faced?

IsaacWarren (approved):
Thanks Ehsan, one question otherwise LGTM

bodo/pandas/frame.py
srilman: Should we allow `bodo.wraps_python` functions here? Maybe not strictly necessary but I can imagine some people may get confused between the two and accidentally do this. Plus, it allows users to manually specify the return type for distribution sake.

srilman (approved):
LGTM just 2 things

bodo/spawn/worker.py
ehsantn: Added a comment. The output doesn't need to be JITable since we can just pickle (as we do with output scalars). The challenge is that scalars may not be replicated in the non-JIT cases like map_partitions, so we have to define
the semantics (e.g. gather all values across ranks in a list).

bodo/pandas/frame.py
ehsantn: Added a comment about globals. JIT and wrap_python functions work too since they are callable objects.

ehsantn: The MPI calls are fine, but we need to think about the distributed semantics. Let's discuss.

bodo/spawn/worker.py
srilman: What does `bodo.typeof` return if its not jittable? Or a type not known in the Numba type registry

ehsantn: I agree, this needs a try/except around it probably. I was focused on the larger design issue.

PR #333, opened 2025-03-07
Add Update Before Installing docker.io

hadia206:
Different jobs on Azure nightly has been failing with following error even with automatic retrying and manual retrying of job.
```
E: Unable to locate package docker.io
E: Couldn't find any package by glob 'docker.io'
E: Couldn't find any package by regex 'docker.io'
```
According to online search, adding `sudo apt-get update` would fix that.

## Testing strategy
N/A. It's inconsistent error on different jobs. Will monitor following nightly runs.

## User facing changes
No

## Checklist


IsaacWarren (approved):
Thanks Hadia, if this doesn't fix it we might just have to turn up the number of retries

PR #341, opened 2025-03-17
[BSE-4611] Polars NYC benchmark

scott-routledge2:
Adds Polars equivalent implementation for NYC taxi benchmark and local notebooks

## Testing strategy
Ran notebook/on cluster. Results
Locally on 1 parquet file: 4.81s
Full dataset (1xr6i.16xlarge): 305s 
Averaged over 3 runs. Note that for the full example a considerable portion of the time is spent computing the UDF in a single thread.

## Checklist


benchmarks/nyc_taxi/polars/nyc_taxi_precipitation.py
```diff
@@ -0,0 +1,79 @@
...
+    weather_dataset = weather_dataset.select(
+        pl.col("DATE").alias("date"),
+        (pl.col("PRCP") > 0.1).alias("date_with_precipitation"),
+    )
+
+    hvfhv_dataset = hvfhv_dataset.with_columns(
+        pl.col("pickup_datetime").dt.date().alias("date"),
+        pl.col("pickup_datetime").dt.month().alias("month"),
+        pl.col("pickup_datetime").dt.hour().alias("hour"),
+        pl.col("pickup_datetime").dt.weekday().is_in([1, 2, 3, 4, 5]).alias("weekday"),
```
scott-routledge2, line 22: Note: Monday is 1
https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.dt.weekday.html

benchmarks/nyc_taxi/README.md
```diff
@@ -142,6 +142,17 @@ In order to run the Spark benchmark:
     aws s3api delete-bucket --bucket $BUCKET_NAME --region us-east-2
     ```
 
+### Polars
+
+1. Install the [**AWS CLI**](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html) and run `aws configure` to set up your credentials.
+2. The script Ray to simplify setup and cleanup of an AWS EC2 instance, to install Ray, you can do `pip install -U "ray[default]"`. To run locally, you will also need to install polars: `pip install "polars[all]"`.
```
ehsantn, line 148:
```suggestion
2. The script uses Ray to simplify setup and cleanup of an AWS EC2 instance, to install Ray, you can do `pip install -U "ray[default]"`. To run locally, you will also need to install polars: `pip install "polars[all]"`.
```

benchmarks/nyc_taxi/polars/nyc_taxi_precipitation.py
```diff
@@ -0,0 +1,79 @@
...
+
+    monthly_trips_weather = monthly_trips_weather.group_by(groupby_columns).agg(
+        pl.col("hvfhs_license_num").count().alias("count"),
+        pl.col("trip_miles").mean().alias("avg_distance"),
+    )
+
+    monthly_trips_weather = monthly_trips_weather.sort(groupby_columns)
+
+    # evaluate the LazyDataframe and store the output
+    monthly_trips_weather = monthly_trips_weather.collect()
```
ehsantn, line 65: The `collect` call shouldn't be necessary. Do we need to return the output?

scott-routledge2: My understanding was that `collect` is needed in order to call `write_parquet` on the output. But they also have this`sink_parquet` API for LazyFrames which writes in streaming mode https://docs.pola.rs/api/python/stable/reference/api/polars.LazyFrame.sink_parquet.html ?

ehsantn: Yes, `sink_parquet` is the lazy version that should be used here.

IsaacWarren (approved):
Thanks Scott, do we need to update the results with this and daft?

scott-routledge2:
> Thanks Scott, do we need to update the results with this and daft?

Yes. I will do that either here or in a separate PR once we update the graph which will be soon.

PR #344, opened 2025-03-20
Vendor DuckDB optimizer

ehsantn:
## Changes included in this PR

Vendors DuckDB source code to make the optimizer available. Copied all of the `src` and `third_party` directories for now (excluding `extension`, `tools`, ... directories fortunately) but we should try to reduce the copied source over time.

DuckDB commit: 9a396a9bd7ef3cad31c5d4f3432a280edece89ab Mar 19, 2025

## Testing strategy

None.

## User facing changes

None right now.

## Checklist
- [NA] Pipelines passed before requesting review. To run CI you must include `[run CI]` in your commit message.
- [x] I am familiar with the [Contributing Guide](https://github.com/bodo-ai/Bodo/blob/main/CONTRIBUTING.md) 
- [x] I have installed + ran pre-commit hooks.

PR #346, opened 2025-03-20
Two read parquets, a merge and a filter and projection creating plans.

DrTodd13:

In initial checkin of some code.  Added top-level readParquet and merge and DataFrame merge and getitem and comparator.  Added some plan nodes that the above code uses to propagate the plan through the execution tree of the program.  This is a work-in-progress PR as there are problems such as a print on the dataframe at the end isn't detected as requiring the plan to be evaluated.  See the dataframe library frontend document for more discussion.

## Testing strategy

Untested other than a single test file I have an as above there are known problems.

## User facing changes

Documentation not updated yet but now there is readParquet and merge at the top-level as well as several other methods supported in the series and dataframe classes to some degree.

## Checklist

### DIFF
--- a/bodo/pandas/__init__.py
+++ b/bodo/pandas/__init__.py
@@ -5,6 +5,55 @@ from bodo.pandas.managers import LazyBlockManager, LazySingleBlockManager
 from bodo.pandas.array_manager import LazyArrayManager, LazySingleArrayManager
 from bodo.pandas.lazy_wrapper import BodoLazyWrapper
 from bodo.pandas.lazy_metadata import LazyMetadata
-from bodo.pandas.utils import get_lazy_manager_class, get_lazy_single_manager_class
+from bodo.pandas.base import *
+import os
 
+DataFrame = BodoDataFrame
+Series = BodoSeries
 
+# If not present or 0 then allow Python to give a symbol not found error
+# if they try to use a Pandas feature that we haven't explicitly implemented.
+# If present and non-zero then try to convert bodo dataframe/series to
+# Pandas equivalent and run standard Pandas for operations not explicitly
+# support by bodo.pandas yet.
+BODO_PANDAS_FALLBACK = int(os.environ.get("BODO_PANDAS_FALLBACK", 0))
+
+def add_fallback():
+    if BODO_PANDAS_FALLBACK != 0:
+        import pandas
+        import inspect
+        import sys
+
+        current_module = sys.modules[__name__]
+
+        # Get all the functions and everything else accessible at the top-level
+        # from the Pandas module.
+        pandas_attrs = dir(pandas)
+        # Do the same for things implemented in Bodo via the bodo.pandas.base import.
+        bodo_df_lib_attrs = dir(current_module)
+
+        # Get the top-level Pandas funcs.
+        pandas_funcs = [attr for attr in pandas_attrs if inspect.isfunction(getattr(pandas, attr))]
+        # Get everything else accessible at the top-level of Pandas.
+        pandas_nonfuncs = [attr for attr in pandas_attrs if not inspect.isfunction(getattr(pandas, attr))]
+
+        # Get the top-level functions support by bodo.pandas.
+        bodo_df_lib_funcs = [attr for attr in bodo_df_lib_attrs if inspect.isfunction(getattr(current_module, attr))]
+        # Get the top-level non-functions support by bodo.pandas.
+        bodo_df_lib_nonfuncs = [attr for attr in bodo_df_lib_attrs if not inspect.isfunction(getattr(current_module, attr))]
+
+        # Get the pandas functions that don't have an equivalent yet in bodo.
+        non_overloaded_funcs = set(pandas_funcs).difference(bodo_df_lib_funcs)
+        # Get the pandas non-functions that don't have an equivalent yet in bodo.
+        non_overloaded_nonfuncs = set(pandas_nonfuncs).difference(bodo_df_lib_nonfuncs)
+
+        for func in non_overloaded_funcs:
+            # Export the pandas functions into bodo.pandas.
+            setattr(current_module, func, getattr(pandas, func))
+
+        for nonfunc in non_overloaded_nonfuncs:
+            # Export the pandas non-functions into bodo.pandas.
+            setattr(current_module, nonfunc, getattr(pandas, nonfunc))
+
+# Must do this at the end so that all functions we want to provide already exist.
+add_fallback()
--- a/bodo/pandas/__init__.py
+++ b/bodo/pandas/__init__.py
@@ -32,28 +32,10 @@ def add_fallback():
         # Do the same for things implemented in Bodo via the bodo.pandas.base import.
         bodo_df_lib_attrs = dir(current_module)
 
-        # Get the top-level Pandas funcs.
-        pandas_funcs = [attr for attr in pandas_attrs if inspect.isfunction(getattr(pandas, attr))]
-        # Get everything else accessible at the top-level of Pandas.
-        pandas_nonfuncs = [attr for attr in pandas_attrs if not inspect.isfunction(getattr(pandas, attr))]
-
-        # Get the top-level functions support by bodo.pandas.
-        bodo_df_lib_funcs = [attr for attr in bodo_df_lib_attrs if inspect.isfunction(getattr(current_module, attr))]
-        # Get the top-level non-functions support by bodo.pandas.
-        bodo_df_lib_nonfuncs = [attr for attr in bodo_df_lib_attrs if not inspect.isfunction(getattr(current_module, attr))]
-
-        # Get the pandas functions that don't have an equivalent yet in bodo.
-        non_overloaded_funcs = set(pandas_funcs).difference(bodo_df_lib_funcs)
-        # Get the pandas non-functions that don't have an equivalent yet in bodo.
-        non_overloaded_nonfuncs = set(pandas_nonfuncs).difference(bodo_df_lib_nonfuncs)
-
-        for func in non_overloaded_funcs:
-            # Export the pandas functions into bodo.pandas.
+        for func in set(pandas_attrs).difference(bodo_df_lib_attrs):
+            # Export the pandas functions that aren't implemented by bodo
+            # into bodo.pandas.
             setattr(current_module, func, getattr(pandas, func))
-
-        for nonfunc in non_overloaded_nonfuncs:
-            # Export the pandas non-functions into bodo.pandas.
-            setattr(current_module, nonfunc, getattr(pandas, nonfunc))
 
 # Must do this at the end so that all functions we want to provide already exist.
 add_fallback()
--- a/bodo/pandas/frame.py
+++ b/bodo/pandas/frame.py
@@ -1,14 +1,16 @@
 import typing as pt
-from collections.abc import Callable
+from collections.abc import Callable, Iterable
 
 import pandas as pd
 
 import bodo
+from bodo.pandas import plans
 from bodo.pandas.array_manager import LazyArrayManager
 from bodo.pandas.lazy_metadata import LazyMetadata
 from bodo.pandas.lazy_wrapper import BodoLazyWrapper
 from bodo.pandas.managers import LazyBlockManager, LazyMetadataMixin
-from bodo.pandas.utils import get_lazy_manager_class
+from bodo.pandas.series import BodoSeries
+from bodo.pandas.utils import convert_to_pandas, get_lazy_manager_class
 from bodo.utils.typing import (
     BodoError,
     check_unsupported_args,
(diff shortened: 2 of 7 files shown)


bodo/pandas/frame.py
```diff
@@ -381,3 +395,95 @@ def map_partitions(self, func, *args, **kwargs):
         return bodo.spawn.spawner.submit_func_to_workers(
             func, [], self, *args, **kwargs
         )
+
+    def merge(
+        self,
+        right,  # BodoDataFrame | BodoSeries,
```
IsaacWarren, line 401: What happens if right is a pandas dataframe or series?

bodo/pandas/plan_operators.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+class ParquetRead(DataSourcePlanOperator):
+    def __init__(self, filename):
+        super().__init__()
+        self.filename = filename
+        self.schema = get_pandas_schema(filename)
+
+
+class Join(LogicalJoin):
```
IsaacWarren, line 129: Maybe rename to EqualityJoin to be specific?

bodo/pandas/plan_operators.py
```diff
@@ -0,0 +1,227 @@
+from bodo.pandas.parquet import get_pandas_schema
+
+
+class PlanOperator:
```
IsaacWarren, line 4: Maybe make these abstract base classes so it's clear they shouldn't be instantiated directly?

IsaacWarren (commented):
Thanks Todd, left a couple comments. The overall structure LGTM, didn't do a deep review since this is in progress

bodo/pandas/series.py
```diff
@@ -18,6 +21,37 @@ class BodoSeries(pd.Series, BodoLazyWrapper):
...
+        self,
+        data=None,
+        index=None,
+        dtype: Dtype | None = None,
+        name=None,
+        copy: bool | None = None,
+        fastpath: bool | lib.NoDefault = lib.no_default,
+        plan: plan_operators.PlanOperator | None = None,
+    ) -> None:
+        self.__dict__["plan"] = plan
```
ehsantn, line 34: Using `__dict__` doesn't seem necessary since "plan" is not a dynamic value?

bodo/pandas/series.py
```diff
@@ -18,6 +21,37 @@ class BodoSeries(pd.Series, BodoLazyWrapper):
...
+            else:
+                zero_size_other = other
+            if hasattr(other, "plan") and other.plan != None:
+                other = other.plan
+            new_metadata = zero_size_self._cmp_method(zero_size_other, op)
+            assert isinstance(new_metadata, pd.Series)
+            return BodoSeries(
+                new_metadata,
+                plan=plan_operators.BinaryOpPlanOperator(self.plan, other, op),
+            )
```
ehsantn, line 51: This code is fine as a placeholder but may not be very robust. Let's add some comments for now at least about the cases and data types that are possible.

bodo/pandas/plan_operators.py
```diff
@@ -0,0 +1,227 @@
+from bodo.pandas.parquet import get_pandas_schema
```
ehsantn, line 1: These nodes will change substantially. But is there a reason to have wrappers in regular Python vs just using the Cython nodes (which are wrappers on C++)?

https://github.com/bodo-ai/Bodo/pull/344/files

bodo/pandas/frame.py
```diff
@@ -24,6 +26,18 @@ class BodoDataFrame(pd.DataFrame, BodoLazyWrapper):
     # use it directly when available.
     _head_df: pd.DataFrame | None = None
 
+    def __init__(
+        self, data=None, index=None, columns=None, dtype=None, copy=None, plan=None
+    ):
+        self.__dict__["plan"] = plan
```
ehsantn, line 32: Using __dict__ doesn't seem necessary since "plan" is not a dynamic value?

bodo/pandas/frame.py
```diff
@@ -24,6 +26,18 @@ class BodoDataFrame(pd.DataFrame, BodoLazyWrapper):
...
+        super().__init__(data, index, columns, dtype)
+
+    @classmethod
+    def from_plan(cls, plan):
+        new_df = BodoDataFrame.__new__(BodoDataFrame)
+        new_df.__dict__["plan"] = plan
```
ehsantn, line 38: Same.

bodo/pandas/frame.py
```diff
@@ -381,3 +395,95 @@ def map_partitions(self, func, *args, **kwargs):
...
+        indicator=False,  # str | bool = False,
+        validate=None,  # MergeValidate | None = None,
+    ):  # -> BodoDataFrame:
+        if (
+            isinstance(right, BodoSeries)
+            or how != "inner"
+            or indicator != False
+            or validate != None
+        ):
+            assert False and "Unsupported option to DataFrame.merge"
```
ehsantn, line 420: Same as above about argument handling and fallback.

bodo/pandas/frame.py
```diff
@@ -381,3 +395,95 @@ def map_partitions(self, func, *args, **kwargs):
...
+                + [
+                    (self.columns.get_loc(a), right.columns.get_loc(b))
+                    for a, b in zip(left_on, right_on)
+                ],
+            )
+
+            return BodoDataFrame(new_metadata, plan=planComparisonJoin)
+
+    def __getitem__(self, key):
+        if hasattr(self, "plan") and self.plan != None:
```
ehsantn, line 460: Need more comments for readability I think.

ehsantn (commented):
Thanks @DrTodd13 . This is a good initial code to get started. Let's just clean it up a bit. See comments.

bodo/pandas/frame.py
DrTodd13: At this point, they have to manually convert to Bodo df before calling merge.

bodo/pandas/plan_operators.py
DrTodd13: I need to understand more about the duckdb classes because the standard join is comparison join with the equal op so I'm not sure what this regular "join" is supposed to be doing.

bodo/pandas/plan_operators.py
DrTodd13: Will do.

bodo/pandas/__init__.py
```diff
@@ -6,5 +6,83 @@
...
+import os
 
+DataFrame = BodoDataFrame
+Series = BodoSeries
 
+BODO_PANDAS_FALLBACK = int(os.environ.get("BODO_PANDAS_FALLBACK", 0))
+
+no_default=...
+
+def read_parquet(path,
```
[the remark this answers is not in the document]
DrTodd13: That was just temporary.  I already fixed it.

bodo/pandas/series.py
DrTodd13: Pandas overloads __getattr__ and that function doesn't recognize "plan" as being valid.  You'll get an error if you don't set it this way.  We could also do our own overload of __getattr__ in BodoDataFrame.

bodo/pandas/frame.py
DrTodd13: Same comment as above.

bodo/pandas/frame.py
DrTodd13: Same reason as above.

bodo/pandas/plan_operators.py
DrTodd13: It isn't clear to me if what you are suggesting is possible yet.  It definitely isn't possible if we are just talking LogicalOperators.  If you take the transitve closure of all types defined or used in LogicalOperators then perhaps but at a minimum it doesn't assemble cleanly as I think the duckdb objects generated for the same Python code are context dependent.

bodo/pandas/series.py
ehsantn: Let's add a comment to explain this since not trivial.

bodo/pandas/frame.py
ehsantn: Same.

bodo/pandas/frame.py
ehsantn: Same.

bodo/pandas/__init__.py
```diff
@@ -6,5 +6,83 @@
...
+        pandas_funcs = [attr for attr in pandas_attrs if inspect.isfunction(getattr(pandas, attr))]
+        pandas_nonfuncs = [attr for attr in pandas_attrs if not inspect.isfunction(getattr(pandas, attr))]
+        pandas_nonfuncs_types = [(attr, type(getattr(pandas, attr))) for attr in pandas_attrs if not inspect.isfunction(getattr(pandas, attr))]
+
+        """
+        print("funcs:", pandas_funcs)
+        print("------------------------")
+        print("non-funcs:", pandas_nonfuncs_types)
+        print("------------------------")
+        """
```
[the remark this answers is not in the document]
DrTodd13: Done.

bodo/pandas/__init__.py
```diff
@@ -6,5 +6,83 @@
...
+        print("non_overloaded_funcs:", non_overloaded_funcs)
+        print("------------------------")
+        print("non_overloaded_non-funcs:", non_overloaded_nonfuncs)
+        print("------------------------")
+        print("overloaded_funcs:", overloaded_funcs)
+        print("------------------------")
+        print("overloaded_non-funcs:", overloaded_nonfuncs)
+        print("------------------------")
+        """
```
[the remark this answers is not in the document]
DrTodd13: Done.

bodo/pandas/__init__.py
```diff
@@ -6,5 +6,83 @@
...
+        import sys
+
+        current_module = sys.modules[__name__]
+
+        pandas_attrs = dir(pandas)
+        bodo_df_lib_attrs = dir(current_module)
+
+        pandas_funcs = [attr for attr in pandas_attrs if inspect.isfunction(getattr(pandas, attr))]
+        pandas_nonfuncs = [attr for attr in pandas_attrs if not inspect.isfunction(getattr(pandas, attr))]
+        pandas_nonfuncs_types = [(attr, type(getattr(pandas, attr))) for attr in pandas_attrs if not inspect.isfunction(getattr(pandas, attr))]
```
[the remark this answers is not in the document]
DrTodd13: Done.

bodo/pandas/frame.py
DrTodd13: Fixed using _internal_names and _internal_names_set from Pandas.

bodo/pandas/series.py
DrTodd13: Fixed using the proper pandas mechanism.

bodo/pandas/__init__.py
```diff
@@ -6,5 +6,83 @@
...
+                 **kwargs):
+    if (engine != 'auto' or
+        columns != None or
+        storage_options != None or
+        use_nullable_dtypes != no_default or
+        dtype_backend != no_default or
+        filesystem != None or
+        filters != None or
+        len(kwargs) > 0):
+        assert False and "Unsupported option to read_parquet"
```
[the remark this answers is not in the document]
DrTodd13: Implemented fallback.

bodo/pandas/__init__.py
```diff
@@ -5,6 +5,55 @@
...
+        # Get the pandas non-functions that don't have an equivalent yet in bodo.
+        non_overloaded_nonfuncs = set(pandas_nonfuncs).difference(bodo_df_lib_nonfuncs)
+
+        for func in non_overloaded_funcs:
+            # Export the pandas functions into bodo.pandas.
+            setattr(current_module, func, getattr(pandas, func))
+
+        for nonfunc in non_overloaded_nonfuncs:
+            # Export the pandas non-functions into bodo.pandas.
+            setattr(current_module, nonfunc, getattr(pandas, nonfunc))
```
ehsantn, line 56: What is the reason for separating functions and non-functions? The code seems the same.

ehsantn (approved):
Thanks @DrTodd13. This is a good starting point to get going. We can discuss the details later (e.g. fallbacks, zero size compute, ...).

bodo/pandas/__init__.py
DrTodd13: Originally I was doing something different there but not anymore.  So, I can simplify it.

IsaacWarren (approved):
One comment otherwise LGTM, thanks Todd

bodo/pandas/plans.py
```diff
@@ -0,0 +1,301 @@
...
+            schema = pd.DataFrame(schema)
+        lazy_mgr = get_lazy_manager_class()(
+            None,
+            None,
+            result_id=str(plan),
+            nrows=1,
+            head=schema._mgr,
+            collect_func=collect_func,
+            del_func=del_func,
+            index_data=None,
```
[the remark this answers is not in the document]
DrTodd13: You may be right that a separate plan manager is better than to overload the existing code.  I think we need to develop a bit more to see which way is cleaner so we may switch to that approach later.

PR #347, opened 2025-03-24
[BSE-4643] Fix caching in BodoSQLContext.sql

scott-routledge2:

Using `bodo_exec` to cache the dynamically generated function. 

## Testing strategy
Running
```python
import bodo
import bodosql
import pandas as pd

bodo.set_verbose_level(2)
df = pd.DataFrame({'A': [1, 2, 3], 'B': [4, 5, 6]})
bc = bodosql.BodoSQLContext({"T1": df})
print(bc.sql("SELECT * FROM T1", cache=True))
```

## Checklist


BodoSQL/bodosql/context.py
```diff
@@ -1111,12 +1112,16 @@ def sql(self, sql, params_dict=None, dynamic_params_list=None, **jit_options):
 
             glbls.update(lowered_globals)
             loc_vars = {}
-            exec(
-                func_text,
-                glbls,
-                loc_vars,
-            )
-            impl = loc_vars["impl"]
+            if bodo.spawn_mode:
```
scott-routledge2, line 1115: I am pretty sure I ran into this problem outside of `bc.sql` case where some func text was executed on the spawner and not available on the workers. Maybe we could move this logic into `bodo_exec`? Not a blocker though.

IsaacWarren: I tried that originally but ran into a bunch of problems with circular dependencies

IsaacWarren (approved):

BodoSQL/bodosql/context.py
DrTodd13: Can't you just import in the function itself to get around that?

DrTodd13 (approved):

PR #348, opened 2025-03-24
Fix Output Print in the Platform

hadia206:
- Broadcast tcp:://hostname/port to workers
- Workers connect to and redirect their stdout to print

## Testing strategy

Code used in notebook for testing
```Python
import os
import bodo
bodo.set_verbose_level(2)
@bodo.jit
def f():
    print("------hello world from", bodo.get_rank(), "|", bodo.get_size(), "\n")
# comment on platform
os.environ["BODO_NUM_WORKERS"]="2"
f()
```

### Platform
  - Created cluster with 2 node
  -  Copy PR files to its corresponding place in `/opt/conda/lib/python3.12/site-packages/bodo/...`
  -  Connect notebook and test
 
Screenshot 2025-03-26 at 11 26 23 AM

### Locally on Mac
- Create a kernel and connect to it from a separate instance of jupyter lab
      -  `pip install ipykernel` and `conda install -c conda-forge jupyterlab`
      -   `python -m ipykernel install --user --name my_test_kernel --display-name "My Test Kernel"`
      -   Open JupyterLab `jupyter lab`.
      -   Create a new notebook and use code above.
      -   In the notebook interface, go to Kernel > Change Kernel > My Test Kernel (or whatever name you used for --display-name).

## Checklist

### DIFF
--- a/bodo/spawn/spawner.py
+++ b/bodo/spawn/spawner.py
@@ -653,7 +653,7 @@ class Spawner:
         # debug_msg(self.logger, "bcast_root: "+ str(self.bcast_root))
         # rank 0 broadcast its hostname and port for the zmq shell
         # if bodo.get_rank() == 0:
-            
+
         for attempt in range(max_attempts):
             try:
                 port = out_socket.bind_to_random_port("tcp://0.0.0.0")
--- a/bodo/spawn/spawner.py
+++ b/bodo/spawn/spawner.py
@@ -625,10 +625,12 @@ class Spawner:
         https://stackoverflow.com/questions/23947281/python-multiprocessing-redirect-stdout-of-a-child-process-to-a-tkinter-text
         """
 
-        # Skip if not in Jupyter or not on Windows
-        if not bodo.utils.utils.is_jupyter_on_windows():
-            self.worker_output_thread = None
-            return
+        # TODO: uncomment
+        # platform_cloud_provider = os.environ.get("BODO_PLATFORM_CLOUD_PROVIDER", None)
+        # Skip if not in Jupyter or not on Windows or not running on Bodo platform
+        # if platform_cloud_provider is None or not bodo.utils.utils.is_jupyter_on_windows():
+        #     self.worker_output_thread = None
+        #     return
 
         import errno
         import threading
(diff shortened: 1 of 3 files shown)


scott-routledge2:
I don't think there is necessarily an issue with this code. The example only prints on rank 0 if running outside of Jupyter (command line spawn mode or `mpiexec`). If you want to print on all ranks you could modify the example:

```
@bodo.jit
def f():
    with bodo.objmode:
        print(f"------hello world from {bodo.get_rank()} | {bodo.get_size()}")
```

This also works but is less clean looking as it sends separate messages:
```
@bodo.jit
def f():
    print("------hello world from", bodo.get_rank(), "|", bodo.get_size())
```

Although I am not exactly sure why.

bodo/spawn/spawner.py
```diff
@@ -646,17 +649,26 @@ def _init_win_jupyter(self):
         except AttributeError:
             win_in_use = None
 
+        ip_value = "127.0.0.1"
+        if bodo.utils.utils.is_jupyter_on_bodo_platform():
+            ip_value = "0.0.0.0"
```
IsaacWarren, line 654: Shouldn't we just always bind to 0.0.0.0, that way this will work on windows clusters?

bodo/spawn/spawner.py
```diff
@@ -646,17 +649,26 @@ def _init_win_jupyter(self):
...
+                port = out_socket.bind_to_random_port(f"tcp://{ip_value}")
             except zmq.ZMQError as ze:
                 # Raise if we have any error not related to socket binding
                 if ze.errno != errno.EADDRINUSE and ze.errno != win_in_use:
                     raise
                 if attempt == max_attempts - 1:
                     raise
 
-        self.worker_intercomm.bcast(port, self.bcast_root)
+        if bodo.utils.utils.is_jupyter_on_bodo_platform():
```
IsaacWarren, line 666: Same comment here, I think we can just always do this first path

bodo/spawn/worker.py
```diff
@@ -524,12 +524,18 @@ def worker_loop(
                 "Jupyter is not supported on multi-node Windows clusters yet"
             )
         port = spawner_intercomm.bcast(None, 0)
-
+        connection_info = f"tcp://{spawner_hostname}:{port}"
+    elif bodo.utils.utils.is_jupyter_on_bodo_platform():
```
IsaacWarren, line 528: Here as well

IsaacWarren (approved):
1 suggestion, otherwise LGTM, thanks Hadia

bodo/spawn/spawner.py
scott-routledge2: I agree I don't see a reason why this wouldn't work on Windows? I could test on my laptop tonight but testing on a cluster might be kind of complicated since we can't use the platform.

scott-routledge2 (approved):
LGTM, thanks @hadia206

bodo/spawn/spawner.py
IsaacWarren: As long as it works on one computer it's not a regression. I think we should just test it on one computer and merge if it works

hadia206: Ok. @scott-routledge2 could you test and let us know

hadia206: I'll merge this PR and have windows/platform cleanup in a followup to not block release

PR #351, opened 2025-03-25
[BSE-4651] Fix caching issues in `overload_gen_pandas_parquet_metadata`

scott-routledge2:
## Checklist

### DIFF
--- a/bodo/io/parquet_write.py
+++ b/bodo/io/parquet_write.py
@@ -7,10 +7,9 @@ import llvmlite.binding as ll
 import pandas as pd
 import pyarrow as pa
 from llvmlite import ir as lir
-from numba.core import cgutils, types
-from numba.extending import intrinsic, overload
 
 import bodo
+from bodo.decorators import wrap_python
 from bodo.hiframes.datetime_date_ext import datetime_date_array_type
 from bodo.hiframes.pd_index_ext import SingleIndexType, array_type_to_index
 from bodo.hiframes.pd_multi_index_ext import MultiIndexType
@@ -29,6 +28,8 @@ from bodo.utils.typing import (
     is_overload_none,
     is_str_arr_type,
 )
+from numba.core import cgutils, types
+from numba.extending import intrinsic, overload
 
 if pt.TYPE_CHECKING:
     from bodo.hiframes.pd_dataframe_ext import DataFrameType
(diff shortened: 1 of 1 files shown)


bodo/tests/test_parquet_write.py
```diff
@@ -13,6 +12,7 @@
 import pytz
 
 import bodo
+import numba
```
scott-routledge2, line 15: This looks like a bug in precommit, not sure why and it didn't change for me locally...

bodo/tests/test_parquet_write.py
```diff
@@ -1080,6 +1080,8 @@ def impl(df, path):
 
 def test_to_pq_multiIdx(check_write_func, memory_leak_check):
     """Test to_parquet with MultiIndexType"""
+    np.random.seed(0)
```
scott-routledge2, line 1083: This test was failing for me locally/on main on multiple ranks

ehsantn (approved):
Looks good to me. It'd be good to add a caching test since this is important functionality.

bodo/tests/caching_tests/test_io_caching.py
```diff
@@ -126,6 +129,27 @@ def impl(fpaths):
     check_caching(impl, (fpaths,), is_cached, fn_distribution, py_output=py_out)
 
 
+@pytest.mark.smoke
```
scott-routledge2, line 132: I have confirmed this fails on main and passes on this branch

bodo/io/parquet_write.py
```diff
@@ -488,41 +484,61 @@ def overload_gen_pandas_parquet_metadata(
...
+        def impl(
+            df,
+            col_names_arr,
+            partition_cols,
+            write_non_range_index_to_metadata,
+            write_rangeindex_to_metadata,
+        ):
+            index = df.index
+            index_names = index.names
+            range_info = (index.start, index.stop, index.step)
```
srilman, line 518: Can you add a comment indicating which line is different between the 2

srilman (approved):

bodo/io/parquet_write.py
scott-routledge2: Would it be possible to actually eliminate the need for two different impls if we just create the `range_info =...` part in python (the `wrap_python` helper?). Just a thought not sure if there are other considerations.

PR #364, opened 2025-03-30
Remove DuckDB Extension Features

srilman:

In order to remove some of the execution code related to reading (like compression-related code), I need to remove all hooks into DuckDB extensions from the codebase.

This is a bit more precarious than the previous PR because it involves modifying files instead of outright deleting them. The goal is to slowly end up deleting most of the files that I modify in this PR.

Note that I am assuming that we will build our Bodo Pandas stuff by using this as a static library with full visibility / access to internal APIs, rather than following their interface and method for setting up extensions.

## Testing strategy

Ran `make debug`. Also complied Bodo with the DF library changes and ran tests in `bodo/tests/test_df_lib`

## User facing changes

N/A

## Checklist


ehsantn:
We will be implementing our data source nodes (Parquet, Iceberg, ...) through extension functionality (`TableFunction` etc.). Will this cause any issues?

srilman:
I believe so, since the focus was to remove the restrictive DuckDB hooks and extension management code. I would have went further and even removed things like the SQL APIs for managing extensions. But if theres something I can do in particular to test, let me know.

ehsantn:
The `dataframe_library` branch has some very minimal unit tests. I'm working on getting read_parquet fully working which would make testing a lot easier.

Another issue is changing source files which makes upgrading DuckDB to pull upstream changes harder. Let's discuss.

ehsantn:
Building this branch fails for me locally:
```
  /Users/ehsan/dev/Bodo/bodo/pandas/_plan.cpp:3:10: fatal error: 'duckdb.hpp' file not found
      3 | #include "duckdb.hpp"
```

ehsantn:
Also, I'm assuming that adding functions to the plan for the optimizer to handle like `TableFunction` and `ScalarFunction` still works. Let's merge main into this and run CI to run our unit tests. I can try to test a `TableFunction` other than `LogicalGet` to confirm as well.

srilman:
@ehsantn I fixed the complication error and ran the tests in `bodo/tests/test_df_lib` to confirm this worked

bodo/pandas/vendor/duckdb/src/include/duckdb/storage/compression/chimp/algorithm/output_bit_stream.hpp
```diff
@@ -8,7 +8,6 @@
 
 #pragma once
 
-#include "duckdb.h"
```
ehsantn, line 11: Maybe it's better to avoid removing duckdb.h and just modify it to minimize changes since it's imported in many places looks like. But not a big deal since the imports are not too many looks like.

ehsantn (approved):
Thanks @srilman . Great to see so much code deleted. Took a look at the diff and seems safe to me. Hopefully we will remove some of the changed files to reduce our patch files.

IsaacWarren (commented):
Thanks Srinivas, before we merge this do we have a plan for generating the patches/the upgrade proces?

srilman:
@IsaacWarren my general thought was to do the following:
- In a separate folder, git clone the relevant sub-folders from the DuckDB repo on the commit that we originally vendored
- Perform the Git patch calculation
- Save the patch somewhere

In the future, we do the following with upgrades:
- git clone a new commit of the DuckDB repo. Again limiting to specific subfolders, ideally just the planner and optimization folders.
- Apply the Git patch
- Fix any conflicts, which can be merged back into the patch

srilman:
@ehsantn any idea on what the failing test is talking about? It looks like an error in test `bodo/tests/test_datetime_timedelta_array.py` but why would that go to DuckDB?

ehsantn:
`test_datetime_timedelta_array.py` works for me locally on this branch. `test_from_pandas.py::test_from_pandas` which is the next test fails. I think the I/O print is out of order probably.

bodo/pandas/vendor/duckdb/test.cpp
IsaacWarren, line 1: Remove this or turn it into a c++ unittest?

IsaacWarren (approved):
One question otherwise LGTM, thanks Srinivas

bodo/pandas/vendor/duckdb/test.cpp
srilman: Sure will remove. Mainly useful for local debugging

PR #367, opened 2025-03-30
Remove Vendored DuckDB Parser Code

srilman:

Remove the SQL parsing code. In particular the vendored Postgres parsing code and the code used to convert it into a DuckDB SQL statement. Also removed the utility code thats essentially a wrapper for building plans in C++.

I also ended up removing some of the logical nodes that caused some issues like for setting properties / settings.

## Testing strategy

PR CI

## User facing changes

N/A

## Checklist


DrTodd13 (approved):
I approve conditionally.  Have we already done these kinds of changes where we are removing individual unneeded lines from files?  My concern is that if we want to upgrade duckdb at some point then these one-off changes are going to make merging future duckdb changes potentially complicated.

srilman:
@DrTodd13 Yes, the problem is in order to incrementally delete code from vendored DuckDB, I need to move line-by-line changes in some files until I can reach the point to delete the file entirely. That doesn't apply to every file though (like the CMakeLists.txt for example), for those the thought is that I will remove as much code as I can and then later go back and smooth out the pieces, adding back smaller removals as necessary to make the upgrade easier.

ehsantn (approved):
Looks good to me. Thanks @srilman .

PR #374, opened 2025-04-01
Basic read_parquet infrastructure

ehsantn:

Creates basic end to end infrastructure for read_parquet with lots of placeholders to fill. Somethings runs now so much easier to test code and do sanity check.

## Testing strategy

Added an end-to-end test.

## User facing changes

None.

## Checklist
- [NA] Pipelines passed before requesting review. To run CI you must include `[run CI]` in your commit message.


bodo/pandas/base.py
```diff
@@ -44,8 +47,29 @@ def read_parquet(
...
+    arrow_schema = pq_dataset.schema
+    nrows = pq_dataset._bodo_total_rows
+
+    empty_df = pa.Table.from_pydict(
+        {k: [] for k in arrow_schema.names}, schema=arrow_schema
+    ).to_pandas()
+    empty_df.index = pd.RangeIndex(0)
+
+    metadata = LazyMetadata("dummy", empty_df, nrows, None)
+    return BodoDataFrame.from_lazy_metadata(
```
DrTodd13, line 70: Please use plan_optimizer.wrap_plan.  Let's try to centralize LazyMetadata or whatever it is we end up using rather than requiring this stuff in every location.  That was the goal of wrap_plan to have one single spot for the integration with existing lazy data structures.

DrTodd13 (changes requested):

bodo/pandas/base.py
ehsantn: I agree. But the wrap_plan code is not working and making these parts robust is part of a follow up task.

bodo/pandas/_executor.cpp
```diff
@@ -0,0 +1,70 @@
...
+    // TODO: support multiple pipelines
+    return pipelines[0].execute();
+}
+
+std::pair<int64_t, PyObject*> Pipeline::execute() {
+    // TODO: support multiple operators
+    return operators[0]->execute();
+}
+
+std::pair<int64_t, PyObject*> PhysicalReadParquet::execute() {
```
DrTodd13, line 36: Docstring.  What are the int64_t and PyObject and you're doing a pointer cast to int64_t?  Why?  I'm sure there's a good reason but please document.

bodo/pandas/_executor.h
```diff
@@ -0,0 +1,58 @@
...
+#include "duckdb/planner/operator/logical_get.hpp"
+
+/**
+ * @brief Physical operators to be used in the execution pipelines (NOTE: they
+ * are Bodo classes and not using DuckDB).
+ *
+ */
+class PhysicalOperator {
+   public:
+    virtual std::pair<int64_t, PyObject*> execute() = 0;
```
DrTodd13, line 15: Why use a pair here rather than make a class?  A class could be better self-documenting.

bodo/pandas/base.py
DrTodd13: Why isn't it working?  Let's just get it working rather than doing things we have to fix later.

bodo/pandas/managers.py
```diff
@@ -195,6 +197,29 @@ def _collect(self):
         """
         Collect data from workers if needed.
         """
+        # Execute the plan if it's lazy
+        if self._plan is not None:
+            from bodo.ext import plan_optimizer
+
+            optimized_plan = plan_optimizer.py_optimize_plan(self._plan)
```
DrTodd13, line 204: Not immediately obvious to me why constructing a lazy manager with a plan requires the immediate execution of that plan.  Some documentation on when this is called and why it is required behavior at that point.

DrTodd13 (changes requested):

bodo/pandas/_executor.cpp
ehsantn: Added comments to the header files. Our convention is to only add docstrings to header files (Google guidelines or something?).

bodo/pandas/_executor.h
ehsantn: It's passed to Cython so a bit easier. This is just a placeholder and will probably change soon anyways.

bodo/pandas/base.py
ehsantn: The whole code will be replaced in the next immediate issue (too large to add to scope of this PR).

bodo/pandas/managers.py
ehsantn: Collect requires the execution of the plan to return the results. It's not constructing a lazy manager. See spawn project documentation for more details.

DrTodd13 (approved):
@ehsantn Please review my changes to this PR as well.  It passed the test case fine.

ehsantn:
Thanks @DrTodd13. Looks good.

PR #378, opened 2025-04-02
BSE-4698: Cleanup plan handling

IsaacWarren:

- Centralizes plan in manager, removes series/df holding its own reference to the plan
- Completes plan handling in all 4 managers

## Testing strategy
Ran df lib tests locally.

## Checklist

### DIFF
--- a/bodo/pandas/frame.py
+++ b/bodo/pandas/frame.py
@@ -29,9 +29,14 @@ class BodoDataFrame(pd.DataFrame, BodoLazyWrapper):
 
     @property
     def _plan(self):
-        if hasattr(self._mgr, "plan") and self._mgr.plan is not None:
-            return self._mgr.plan
-        return plan_optimizer.LogicalGetDataframeRead(self._mgr._md_result_id)
+        if hasattr(self._mgr, "plan"):
+            if self._mgr.plan is not None:
+                return self._mgr.plan
+            else:
+                return plan_optimizer.LogicalGetDataframeRead(self._mgr._md_result_id)
+        raise NotImplementedError(
+            "Plan not available for this manager, recreate this dataframe with from_pandas"
+        )
 
     @staticmethod
     def from_lazy_mgr(
--- a/bodo/pandas/series.py
+++ b/bodo/pandas/series.py
@@ -21,9 +21,15 @@ class BodoSeries(pd.Series, BodoLazyWrapper):
 
     @property
     def _plan(self):
-        if hasattr(self._mgr, "plan") and self._mgr.plan is not None:
-            return self._mgr.plan
-        return plan_optimizer.LogicalGetSeriesRead(self._mgr._md_result_id)
+        if hasattr(self._mgr, "plan"):
+            if self._mgr.plan is not None:
+                return self._mgr.plan
+            else:
+                return plan_optimizer.LogicalGetSeriesRead(self._mgr._md_result_id)
+
+        raise NotImplementedError(
+            "Plan not available for this manager, recreate this series with from_pandas"
+        )
 
     def _cmp_method(self, other, op):
         """Called when a BodoSeries is compared with a different entity (other)


srilman (approved):

bodo/pandas/series.py
```diff
@@ -19,13 +19,16 @@ class BodoSeries(pd.Series, BodoLazyWrapper):
...
+    def _plan(self):
+        if hasattr(self._mgr, "plan"):
+            return self._mgr.plan
+        return None
 
     def _cmp_method(self, other, op):
         from bodo.pandas.base import empty_like
 
-        if hasattr(self, "plan") and self.plan != None:
+        if hasattr(self, "plan") and self._plan != None:
```
ehsantn, line 31: Remove `hasattr(self, "plan")`?

DrTodd13: Yes, let's get rid of these checks.  The consequence of that is there's going to be dead code.  If you look at what comes after these checks there's often additional code that now should never be reached.  The "if plan is None" should not fall to these cases but create a data source for the existing dataframe/series.  I made that change in a few places and that is what is causing the conflicts in the rebase that needs to be done.

DrTodd13 (changes requested):

DrTodd13 (approved):

bodo/pandas/frame.py
```diff
@@ -31,7 +31,7 @@ class BodoDataFrame(pd.DataFrame, BodoLazyWrapper):
     def _plan(self):
         if hasattr(self._mgr, "plan"):
             return self._mgr.plan
-        return None
+        return plan_optimizer.LogicalGetDataframeRead(self._mgr._md_result_id)
```
DrTodd13, line 34: I'm not sure here.  I think we need to look at self._mgr.plan and if that is None then create the LogicalGet node.  If there is no plan field at all then how did that happen?  Perhaps creating a LogicalGet is correct in that case but would like to understand that scenario better to say if this is the right behavior.

bodo/pandas/frame.py
```diff
@@ -451,11 +451,6 @@ def merge(
             suffixes=suffixes,
         )
 
-        self_plan = (
-            self._plan
-            if self._plan is not None
-            else plan_optimizer.LogicalGetDataframeRead(self._mgr._md_result_id)
-        )
         right_plan = (
```
DrTodd13, line 454: Do the same fix for right_plan.  That should be right..plan now right?

bodo/pandas/series.py
```diff
@@ -21,41 +21,34 @@ class BodoSeries(pd.Series, BodoLazyWrapper):
 
     @property
     def _plan(self):
-        if hasattr(self._mgr, "plan"):
+        if hasattr(self._mgr, "plan") and self._mgr.plan is not None:
```
DrTodd13, line 24: Same comment as for dataframe version.

bodo/pandas/series.py
```diff
@@ -21,41 +21,34 @@ class BodoSeries(pd.Series, BodoLazyWrapper):
...
         zero_size_self = _empty_like(self)
         zero_size_other = _empty_like(other) if isinstance(other, BodoSeries) else other
         # This is effectively a check for a dataframe or series.
-        if hasattr(other, "_plan"):
-            # The other.plan==None part of this line is untested.
-            other = (
-                other._plan
-                if other._plan is not None
-                else plan_optimizer.LogicalGetSeriesRead(other._mgr._md_result_id)
+        if not isinstance(other, (BodoSeries, BodoDataFrame)):
```
DrTodd13, line 39: This change should be reverted.  "other" can be many things, e.g., constant integer.

bodo/pandas/series.py
```diff
@@ -21,41 +21,34 @@ class BodoSeries(pd.Series, BodoLazyWrapper):
...
             )
+
         # Compute schema of new series.
         new_metadata = zero_size_self._cmp_method(zero_size_other, op)
         assert isinstance(new_metadata, pd.Series)
         return plan_optimizer.wrap_plan(
             new_metadata,
             plan=plan_optimizer.LazyPlan(
-                plan_optimizer.LogicalBinaryOp, self_plan, other, op
+                plan_optimizer.LogicalBinaryOp, self._plan, other._plan, op
```
DrTodd13, line 51: Revert the other._plan part.  If other has a _plan then use it else use the original value.

DrTodd13 (changes requested):

bodo/pandas/frame.py
IsaacWarren: Self._mgr might not be a lazy manager in which case it won't have the plan attribute

DrTodd13: If self._mgr isn't a lazy mgr, will it have _md_result_id?

DrTodd13 (approved):

PR #379, opened 2025-04-02
Add get_gpu_ranks

hadia206:

As Title
## Checklist

### DIFF
--- a/bodo/libs/distributed_api.py
+++ b/bodo/libs/distributed_api.py
@@ -4672,10 +4672,33 @@ def get_num_nodes():  # pragma: no cover
     return len(get_host_ranks())
 
 
-def get_gpu_ranks(num_gpus_in_node):  # pragma: no cover
+def get_num_gpus(framework="torch"):  # pragma: no cover
+    """Get number of GPU devices on this host"""
+    if framework == "torch":
+        try:
+            import torch
+
+            return torch.cuda.device_count()
+        except ImportError:
+            raise RuntimeError(
+                "PyTorch is not installed. Please install PyTorch to use GPU features."
+            )
+    elif framework == "tensorflow":
+        try:
+            import tensorflow as tf
+
+            return len(tf.config.list_physical_devices("GPU"))
+        except ImportError:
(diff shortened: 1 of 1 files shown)


bodo/libs/distributed_api.py
```diff
@@ -4672,6 +4672,40 @@ def get_num_nodes():  # pragma: no cover
     return len(get_host_ranks())
 
 
+def get_gpu_ranks(num_gpus_in_node):  # pragma: no cover
```
IsaacWarren, line 4675: Could we use https://pytorch.org/docs/stable/generated/torch.cuda.device_count.html instead of passing in num_gpus_in_node and only allow this function to be called if torch is installed?

hadia206: We could. 
However, I still need to use it in these lines so didn't feel the need to.
```
model = TaxiDemandNN(input_size=X_train.shape[1]).to(f"cuda:{local_rank % num_gpus}" if rank in gpu_ranks else "cpu")
    model = torch.nn.parallel.DistributedDataParallel(model, device_ids=[local_rank % num_gpus] if rank in gpu_ranks else None)
```

IsaacWarren: Theoretically that should use the torch api as well, IMO specifying the total number of gpus in a cluster should never have to be done manually

hadia206: It's using torch api, not manual, see [code](https://github.com/bodo-ai/Bodo/pull/379/files) 
`num_gpus = torch.cuda.device_count()`
I was just saying if I'm doing it there, didn't see need to hide it in `get_gpu_ranks`

hadia206: ok. That makes sense. Let me update code then

hadia206: Updated code using `get_num_gpus` from
https://github.com/bodo-ai/Bodo-Archive/pull/8581/files

IsaacWarren (approved):
Thanks Hadia

PR #384, opened 2025-04-04
Run dataframe library plans in parallel

ehsantn:

Adds support for running the plans in parallel and updates from_pandas and read_parquet to work in parallel. Added a flag to run the plan sequentially for debugging purposes.

## Testing strategy

Added unit test.

## User facing changes

None.

## Checklist

### DIFF
--- a/bodo/pandas/_executor.cpp
+++ b/bodo/pandas/_executor.cpp
@@ -56,9 +56,17 @@ std::pair<int64_t, PyObject*> PhysicalReadParquet::execute() {
     arrow::py::import_pyarrow_wrappers();
     PyObject* pyarrow_schema = arrow::py::wrap_schema(table->schema());
 
+    // Get this rank's portion of rows
+    int rank = dist_get_rank();
+    int num_ranks = dist_get_size();
+    int64_t start = dist_get_start(table->num_rows(), num_ranks, rank);
+    int64_t end = dist_get_end(table->num_rows(), num_ranks, rank);
+    int64_t length = end - start;
+    std::shared_ptr<arrow::Table> sliced_table = table->Slice(start, length);
+
     auto* bodo_pool = bodo::BufferPool::DefaultPtr();
     std::shared_ptr<table_info> out_table =
-        arrow_table_to_bodo(table, bodo_pool);
+        arrow_table_to_bodo(sliced_table, bodo_pool);
 
     return {reinterpret_cast<int64_t>(new table_info(*out_table)),
             pyarrow_schema};
--- a/bodo/pandas/base.py
+++ b/bodo/pandas/base.py
@@ -4,7 +4,7 @@ from pandas._libs import lib
 from bodo.ext import plan_optimizer
 from bodo.pandas.frame import BodoDataFrame
 from bodo.pandas.series import BodoSeries
-from bodo.pandas.utils import LazyPlan, scatter_data
+from bodo.pandas.utils import LazyPlan
 
 
 def from_pandas(df):
@@ -29,7 +29,7 @@ def from_pandas(df):
     arrow_schema = pa.Schema.from_pandas(df)
 
     if bodo.dataframe_library_run_parallel:
-        res_id = scatter_data(df)
+        res_id = bodo.spawn.utils.scatter_data(df)
         plan = LazyPlan("LogicalGetPandasReadParallel", res_id, arrow_schema)
     else:
         plan = LazyPlan("LogicalGetPandasReadSeq", df, arrow_schema)
--- a/bodo/pandas/base.py
+++ b/bodo/pandas/base.py
@@ -4,12 +4,14 @@ from pandas._libs import lib
 from bodo.ext import plan_optimizer
 from bodo.pandas.frame import BodoDataFrame
 from bodo.pandas.series import BodoSeries
-from bodo.pandas.utils import LazyPlan
+from bodo.pandas.utils import LazyPlan, scatter_data
 
 
 def from_pandas(df):
     """Convert a Pandas DataFrame to a BodoDataFrame."""
     import pyarrow as pa
+
+    import bodo
 
     if not isinstance(df, pd.DataFrame):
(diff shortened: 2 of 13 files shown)


bodo/pandas/base.py
```diff
@@ -4,24 +4,35 @@
...
     if not isinstance(df, pd.DataFrame):
         raise TypeError("Input must be a pandas DataFrame")
 
+    # TODO: Add support for Index
+    if (
+        not isinstance(df.index, pd.RangeIndex)
+        or df.index.start != 0
+        or df.index.step != 1
+    ):
+        raise ValueError("Only RangeIndex with start=0 and step=1 is supported")
```
ehsantn, line 25: Created an issue since we need to add Index handling in general to our tables similar to Arrow probably. Parquet read development is probably a good time to do it.

bodo/pandas/base.py
```diff
@@ -4,24 +4,35 @@
...
     arrow_schema = pa.Schema.from_pandas(df)
 
-    # TODO: distribute to workers and get result_id
-    plan = plan_optimizer.LazyPlan(
-        plan_optimizer.LogicalGetPandasRead, df, arrow_schema
-    )
-    # TODO: Add support for Index
+    if bodo.dataframe_library_run_parallel:
+        res_id = scatter_data(df)
+        plan = LazyPlan("LogicalGetPandasReadParallel", res_id, arrow_schema)
```
ehsantn, line 33: Avoiding Cython in the plan since it will be pickled to workers.

bodo/pandas/utils.py
```diff
@@ -91,3 +98,83 @@ def cpp_table_to_df(cpp_table, arrow_schema):
...
+            if isinstance(x, LazyPlan):
+                return x.generate_duckdb(cache=cache)
+            else:
+                return x
+
+        # Convert any LazyPlan in the args or kwargs.
+        args = [recursive_check(x) for x in self.args]
+        kwargs = {k: recursive_check(v) for k, v in self.kwargs.items()}
+        # Create real duckdb class.
+        ret = getattr(plan_optimizer, self.plan_class)(*args, **kwargs)
```
ehsantn, line 137: This line is the only change in this class.

DrTodd13 (approved):
LGTM.

bodo/pandas/_executor.cpp
```diff
@@ -56,9 +56,17 @@ std::pair<int64_t, PyObject*> PhysicalReadParquet::execute() {
     arrow::py::import_pyarrow_wrappers();
     PyObject* pyarrow_schema = arrow::py::wrap_schema(table->schema());
 
+    // Get this rank's portion of rows
+    int rank = dist_get_rank();
+    int num_ranks = dist_get_size();
+    int64_t start = dist_get_start(table->num_rows(), num_ranks, rank);
+    int64_t end = dist_get_end(table->num_rows(), num_ranks, rank);
+    int64_t length = end - start;
+    std::shared_ptr<arrow::Table> sliced_table = table->Slice(start, length);
```
IsaacWarren, line 65: Add a comment saying this will need to change when we integrate with our actual parquet reader?

bodo/tests/test_df_lib/test_from_pandas.py
```diff
@@ -14,12 +14,24 @@ def test_from_pandas(datapath):
             "c": ["a", "b", "c"],
         }
     )
+    # Sequential test
+    with temp_config_override("dataframe_library_run_parallel", False):
+        bdf = bd.from_pandas(df)
+        assert bdf._lazy
+        assert bdf.plan is not None
+        assert bdf.plan.plan_class == "LogicalGetPandasReadSeq"
```
IsaacWarren, line 22: Add these asserts after the execution too?

bodo/pandas/utils.py
IsaacWarren: I know I'm probably going to get overridden but I don't like this. Even just having a python wrapper class for the cython classes or overriding pickling for the cython class (if that's possible) if we don't want to do full serialization.

bodo/pandas/utils.py
```diff
@@ -91,3 +98,83 @@ def cpp_table_to_df(cpp_table, arrow_schema):
...
+
+    if bodo.dataframe_library_run_parallel:
+        import bodo.spawn.spawner
+
+        return bodo.spawn.spawner.submit_func_to_workers(_exec_plan, [], plan)
+
+    return _exec_plan(plan)
+
+
+def scatter_data(data: pd.DataFrame | pd.Series | pd.Index | pd.array) -> str:
```
IsaacWarren, line 168: Is this not already somewhere in the spawner? if not we should replace where we do something similar in the spawner and move this function to that file

IsaacWarren (commented):
Thanks Ehsan, a couple comments

bodo/pandas/utils.py
ehsantn: I agree. These indirections are confusing and not very robust. Looks like we need separate Python classes to avoid these Cython issues in general. Let's discuss with the team and address in a future PR (hopefully soon).

bodo/pandas/_executor.cpp
ehsantn: Top of the function already has a comment and there is an issue. An extra comment here can be confusing I think.

bodo/tests/test_df_lib/test_from_pandas.py
ehsantn: Done.

bodo/pandas/utils.py
DrTodd13: Personally, I find the indirection elegant and the getattr for the class name and the *args/**kwargs makes it, I think, very robust.

bodo/pandas/utils.py
ehsantn: It's not in the spawner but move this to the spawn/utility.py which is probably more clear.

IsaacWarren (approved):

PR #385, opened 2025-04-04
[WIP] Logical to Physical Translator and Physical Executor

srilman:
## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -520,10 +520,7 @@ target_link_directories(native_typer PRIVATE ${CONDA_LIB_DIR})
 target_link_libraries(native_typer PRIVATE fmt::fmt)
 install(TARGETS native_typer DESTINATION "bodo/transforms/type_inference/")
 
-
-
 # ---------------------- Cython Target - plan_optimizer -----------------------
-
 add_custom_command(
   OUTPUT bodo/pandas/plan_optimizer.cpp
   DEPENDS bodo/pandas/plan_optimizer.pyx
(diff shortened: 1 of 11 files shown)


DrTodd13 (commented):
One thing is making me nervous here.  This PR seems to make the assumption that there is always one and only one result that needs to be transferred between operators.  Perhaps we will try to do something with duckdb that you can't do in SQL.  Something like df1[df2.col > X] would seem to be valid Pandas, yes?  The filtering operation here seems to require access to two dataframes.  Maybe this just means our frontend has to generate extra code in this case to transform it into something that is valid SQL and which would then have this property of single source.  I presume this design is coming from duckdb's approach?  It is also never the case that the output of one operator is needed by multiple downstream operators?

ehsantn (commented):
@srilman looks like this PR is on the right track. But it has a lot of gaps and doesn't even compile. Let's merge main into it and make sure the tests pass. I'd be happy to help.

ehsantn:
> One thing is making me nervous here. This PR seems to make the assumption that there is always one and only one result that needs to be transferred between operators. Perhaps we will try to do something with duckdb that you can't do in SQL. Something like df1[df2.col > X] would seem to be valid Pandas, yes? The filtering operation here seems to require access to two dataframes. Maybe this just means our frontend has to generate extra code in this case to transform it into something that is valid SQL and which would then have this property of single source. I presume this design is coming from duckdb's approach? It is also never the case that the output of one operator is needed by multiple downstream operators?

@DrTodd13 I think the frontend has to the extra work since changing the backend away from well-known SQL-style vectorized execution can get complicated and risky. We may decide to not support some less common cases in optimized plan and trigger execution earlier. Your example may be in that category.

bodo/libs/_bodo_common.cpp
```diff
@@ -142,6 +145,23 @@ Bodo_CTypes::CTypeEnum arrow_to_bodo_type(arrow::Type::type type) {
     }
 }
 
+bodo_array_type::arr_type_enum type_to_array_type(Bodo_CTypes::CTypeEnum typ) {
```
ehsantn, line 148: @DrTodd13 @srilman what can we do to make these type conversions more complete and robust without a lot of effort?

ehsantn:
@srilman I can't tag you for review since you are the original author. Please review.

DrTodd13 (approved):
LGTM.

PR #387, opened 2025-04-07
[BSE-4679] Get Parquet dataset without row count.

scott-routledge2:

Removes getting nrows eagerly. Note that this will give incorrect `shape` and `len` when called before the plan is executed. (The columns will be correct but length is 1), this impacts our testing because we will have to make sure we call `_collect` before using any of the `pd.testing` functionality since they check the shape first.

## Testing strategy
`pytest -svW ignore bodo/tests/test_df_lib`

## Checklist

### DIFF
--- a/bodo/pandas/base.py
+++ b/bodo/pandas/base.py
@@ -80,16 +80,15 @@ def read_parquet(
 
     # Read Parquet schema and row count
     # TODO: Make this more robust (e.g. handle Index, etc.)
-    use_hive = True
+    use_hive = kwargs.get("_bodo_use_hive", True)
     pq_dataset = get_parquet_dataset(
         path,
-        get_row_counts=True,
+        get_row_counts=False,
         storage_options=storage_options,
         read_categories=True,
         partitioning="hive" if use_hive else None,
     )
     arrow_schema = pq_dataset.schema
-    nrows = pq_dataset._bodo_total_rows
 
     empty_df = pa.Table.from_pydict(
         {k: [] for k in arrow_schema.names}, schema=arrow_schema
@@ -97,7 +96,7 @@ def read_parquet(
     empty_df.index = pd.RangeIndex(0)
 
     plan = LazyPlan("LogicalGetParquetRead", path.encode(), arrow_schema)
-    return plan_optimizer.wrap_plan(empty_df, plan=plan, nrows=nrows)
+    return plan_optimizer.wrap_plan(empty_df, plan=plan, nrows=None)
 
 
 def merge(lhs, rhs, *args, **kwargs):
--- a/bodo/pandas/base.py
+++ b/bodo/pandas/base.py
@@ -80,7 +80,7 @@ def read_parquet(
 
     # Read Parquet schema and row count
     # TODO: Make this more robust (e.g. handle Index, etc.)
-    use_hive = kwargs.get("_bodo_use_hive", True)
+    use_hive = True
     pq_dataset = get_parquet_dataset(
         path,
         get_row_counts=False,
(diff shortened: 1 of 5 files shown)


bodo/tests/test_df_lib/test_read_parquet.py
```diff
@@ -11,6 +11,8 @@ def test_read_parquet(datapath):
     bodo_out = bd.read_parquet(path)
     py_out = pd.read_parquet(path)
 
+    print(bodo_out)
```
scott-routledge2, line 14: Print is here to ensure the results are on the spawner, otherwise the index would be wrong which would lead to `bodo_out.shape` to not match.

bodo/tests/test_df_lib/test_read_parquet.py
```diff
@@ -11,6 +11,9 @@ def test_read_parquet(datapath):
     bodo_out = bd.read_parquet(path)
     py_out = pd.read_parquet(path)
 
+    # Evaluate LazyDataFrame to get correct shape
+    bodo_out._mgr._collect()
```
scott-routledge2, line 15: Our `_test_equal` first tests the shape so we need to make sure we execute the plan first since now the row count won't be known until execution time. Maybe as a followup we could add an option to "collect data first" in our `_test_equal` function.

DrTodd13: Once we trigger on requests for shape then we won't need the explicit collect.

DrTodd13 (changes requested):
Just documenting our video conversation.  Needs a trigger on request for shape and needs similar changes made in array_manager.py.

bodo/pandas/managers.py
```diff
@@ -431,6 +432,7 @@ def _collect(self):
             data = execute_plan(self._plan)
             self._plan = None
             self.blocks = data._mgr.blocks
+            self.axes = data._mgr.axes
```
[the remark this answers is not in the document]
ehsantn: Let's add a comment to the code itself.

ehsantn (approved):
I'm ok with merging this but fixing len/shape in an immediate follow up PR since this isn't released yet.

scott-routledge2:
Updated to add collect before `len`/`shape`

bodo/tests/test_df_lib/test_read_parquet.py
```diff
@@ -15,3 +15,19 @@ def test_read_parquet(datapath):
...
+
+
+def test_read_parquet_len_shape(datapath):
+    """Test length/shape after read parquet is correct"""
+    path = datapath("example.parquet")
+
+    bodo_out = bd.read_parquet(path)
+    py_out = pd.read_parquet(path)
+
+    len(bodo_out) == py_out
```
ehsantn, line 27:
```suggestion
    len(bodo_out) == len(py_out)
```

ehsantn (approved):

bodo/pandas/base.py
```diff
@@ -58,21 +58,20 @@ def read_parquet(
...
-    nrows = pq_dataset._bodo_total_rows
 
     empty_df = pa.Table.from_pydict(
         {k: [] for k in arrow_schema.names}, schema=arrow_schema
     ).to_pandas()
     empty_df.index = pd.RangeIndex(0)
 
     plan = LazyPlan("LogicalGetParquetRead", path.encode(), arrow_schema)
-    return plan_optimizer.wrap_plan(empty_df, plan=plan, nrows=nrows)
+    return plan_optimizer.wrap_plan(empty_df, plan=plan, nrows=None)
```
DrTodd13, line 74:
```suggestion
    return plan_optimizer.wrap_plan(empty_df, plan=plan)
```

DrTodd13 (approved):
LGTM with that last suggestion.  Thanks.

PR #390, opened 2025-04-08
[WIP] Avoid collecting lazy plan immediately after executing

scott-routledge2:

When executing a plan, try to keep results distributed across nodes. Also adds workaround for getting length from lazy dataframes generated with plan by first executing and returning metadata. 

It ended up being a little messier than I thought, the main issue is that our `_collect`behaves differently when called with a plan vs a regular Lazy DataFrame. When executing a plan without collecting, we would probably want to just return a new `BodoDataFrame`. 

## Checklist

### DIFF
--- a/bodo/pandas/frame.py
+++ b/bodo/pandas/frame.py
@@ -43,6 +43,12 @@ class BodoDataFrame(pd.DataFrame, BodoLazyWrapper):
             "Plan not available for this manager, recreate this dataframe with from_pandas"
         )
 
+    @property
+    def index(self):
+        if self.is_lazy_plan():
+            self._mgr._collect()
+        return super().index
+
     @staticmethod
     def from_lazy_mgr(
         lazy_mgr: LazyArrayManager | LazyBlockManager,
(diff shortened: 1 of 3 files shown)


bodo/pandas/frame.py
```diff
@@ -43,6 +43,12 @@ def _plan(self):
             "Plan not available for this manager, recreate this dataframe with from_pandas"
         )
 
+    @property
+    def index(self):
```
scott-routledge2, line 47: This avoids using dummy metadata for `shape`, `len`, `index`, which would just be a single row without executing a plan.

bodo/tests/test_df_lib/test_read_parquet.py
```diff
@@ -11,10 +11,24 @@ def test_read_parquet(datapath):
...
+    """
+    path = datapath("example.parquet")
+
+    bodo_out = bd.read_parquet(path)
+    py_out = pd.read_parquet(path)
+
+    assert len(bodo_out) == len(py_out)
+    assert bodo_out.shape == py_out.shape
+
+    bodo_out.head()
```
scott-routledge2, line 32: This statement causes the Dataframe to be collected on the spawner.

bodo/pandas/managers.py
```diff
@@ -209,12 +209,15 @@ def _collect(self):
             )
             data = execute_plan(self._plan)
             self._plan = None
-            self.blocks = data._mgr.blocks
             self.axes = data._mgr.axes
-            self._md_result_id = None
-            self._md_nrows = None
-            self._md_head = None
-            BlockManager._rebuild_blknos_and_blklocs(self)
+            self._md_result_id = data._mgr._md_result_id
```
scott-routledge2, line 213: evidently copying these attributes here may not be enough, we would probably need to initialize a new `BodoDataFrame`

PR #398, opened 2025-04-10
[DF Lib] Fix index update for Lazy Series

scott-routledge2:

Fixes `bodo/tests/test_df_lib/test_end_to_end.py::test_projection`. Also minor script issue in PR CI.

## Testing strategy
https://github.com/bodo-ai/Bodo/actions/runs/14393125025/job/40364369689?pr=398

## Checklist

### DIFF
--- a/bodo/pandas/managers.py
+++ b/bodo/pandas/managers.py
@@ -425,6 +425,8 @@ class LazySingleBlockManager(SingleBlockManager, LazyMetadataMixin[SingleBlockMa
             self.axes = [mgr.axes[1]]
         elif isinstance(mgr, SingleBlockManager):
             self.axes = mgr.axes
+        else:
+            raise ValueError("Unknown manager type:", type(mgr))
 
     def _collect(self):
         """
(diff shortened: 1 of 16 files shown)


bodo/pandas/managers.py
```diff
@@ -416,6 +416,16 @@ def __repr__(self) -> str:
         else:
             return super().__repr__()
 
+    def _update_index_from_mgr(self, mgr):
```
scott-routledge2, line 419: In theory we should not need this function since `execute_plan` should always return a Series that has a `SingleBlockManager`, so we can set indexes as before. Just a workaround for now to verify the fix works for now.

DrTodd13 (commented):
Do you need equivalent changes in array_manager.py compared to manager.py?

scott-routledge2:
> Do you need equivalent changes in array_manager.py compared to manager.py?

I am assuming so, however once we merge https://github.com/bodo-ai/Bodo/pull/398, I was going to remove because the axes will be the same length.

DrTodd13 (approved):
LGTM.

ehsantn (approved):

PR #400, opened 2025-04-10
Fix plan execution result type when it should be a series.

DrTodd13:
1) Move wrap_plan from plan_optimizer to utils.  No reason for it to be in Cython so easier debugging in Python.

2) Add dataframe or series flag to LazyPlan which is filled in my wrap_plan and passed to py_execute_plan which uses it to call cpp_table_to_series if the output is of series type. 3) add cpp_table_to_series which uses the existing cpp_table_to_df and then selects the singular column to get the series.

## Testing strategy

run_ci

## User facing changes

None

## Checklist

### DIFF
--- a/bodo/pandas/plan_optimizer.pyx
+++ b/bodo/pandas/plan_optimizer.pyx
@@ -449,7 +449,7 @@ cpdef py_optimize_plan(object plan):
     optimized_plan.c_logical_operator = optimize_plan(move(wrapped_operator.c_logical_operator))
     return optimized_plan
 
-cpdef py_execute_plan(object plan, bint dataframe_output):
+cpdef py_execute_plan(object plan, output_func):
     """Execute a logical plan in the C++ backend
     """
     cdef LogicalOperator wrapped_operator
@@ -464,10 +464,5 @@ cpdef py_execute_plan(object plan, bint dataframe_output):
     exec_output = execute_plan(move(wrapped_operator.c_logical_operator))
     cpp_table = exec_output.first
     arrow_schema = <object>exec_output.second
-    assert dataframe_output is not None
-    if dataframe_output:
-        from bodo.pandas.utils import cpp_table_to_df
-        return cpp_table_to_df(cpp_table, arrow_schema)
-    else:
-        from bodo.pandas.utils import cpp_table_to_series
-        return cpp_table_to_series(cpp_table, arrow_schema)
+    assert output_func is not None
+    return output_func(cpp_table, arrow_schema)
--- a/bodo/pandas/plan_optimizer.pyx
+++ b/bodo/pandas/plan_optimizer.pyx
@@ -449,54 +449,9 @@ cpdef py_optimize_plan(object plan):
     optimized_plan.c_logical_operator = optimize_plan(move(wrapped_operator.c_logical_operator))
     return optimized_plan
 
-def _del_func(x):
-    # Intentionally do nothing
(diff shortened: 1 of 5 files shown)


bodo/pandas/plan_optimizer.pyx
```diff
@@ -509,5 +464,10 @@ cpdef py_execute_plan(object plan):
...
     arrow_schema = <object>exec_output.second
-
-    return cpp_table_to_df(cpp_table, arrow_schema)
+    assert dataframe_output is not None
+    if dataframe_output:
+        from bodo.pandas.utils import cpp_table_to_df
+        return cpp_table_to_df(cpp_table, arrow_schema)
+    else:
+        from bodo.pandas.utils import cpp_table_to_series
+        return cpp_table_to_series(cpp_table, arrow_schema)
```
scott-routledge2, line 473: What happens if a plan produces something other than a Series of DataFrame e.g. `Series.sum()`? Would it just produce a series with a single element?

scott-routledge2 (approved):
Couple questions but this LGTM, thanks @DrTodd13

bodo/pandas/utils.py
```diff
@@ -365,11 +374,75 @@ def _exec_plan(plan):
...
+            schema = pd.DataFrame(schema)
+        metadata = LazyMetadata(
+            "LazyPlan_" + str(plan.plan_class) if res_id is None else res_id,
+            schema,
+            nrows=nrows,
+            index_data=index_data,
+        )
+        mgr = get_lazy_manager_class()
+        new_df = BodoDataFrame.from_lazy_metadata(
+            metadata, collect_func=mgr._collect, del_func=_del_func, plan=plan
```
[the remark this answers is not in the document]
DrTodd13: Why do you say the value isn't being used?

bodo/pandas/plan_optimizer.pyx
DrTodd13: I think you bring up an important topic here we should think about more.  If it is just going to return a number then there is no object to which we can attach a lazy plan.  So, perhaps sum creates a non-lazy node for the sum and the source of that is the lazy plan for whatever comes before.  That way the entity executing the plan knows it isn't going to return a dataframe or series.  The question then becomes do we have baked into the backend that the result is always table_info type.

bodo/pandas/utils.py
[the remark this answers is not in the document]
scott-routledge2: When we collect a lazy dataframe with a plan, we first execute the plan, which generates a new lazy BodoDataFrame and wraps it with a different collect func (see spawner.py:379). We then immediately collect that DataFrame and bring it onto the spawner, setting the collect func of the original BodoDataFrame to None and updating it's blocks.

bodo/pandas/plan_optimizer.pyx
ehsantn: I think we may need to extend the flag to signal scalars too.

DrTodd13: Will do.

bodo/pandas/utils.py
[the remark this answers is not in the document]
DrTodd13: What about the case where it is evaluated but not brought to spawn node?

ehsantn (approved):
Thanks @DrTodd13 . Looks good to me.

PR #402, opened 2025-04-11
BSE-4705: spawn plan interop

IsaacWarren:
## Checklist

### DIFF
--- a/bodo/spawn/spawner.py
+++ b/bodo/spawn/spawner.py
@@ -292,6 +292,13 @@ class Spawner:
 
             signal.signal(signal.SIGUSR1, handler)
 
+        # If we get a df/series with a plan we need to execute it and get the result id
+        # so we can build the arg metadata
+        for arg in args:
+            if isinstance(arg, BodoLazyWrapper) and arg._plan is not None:
+                # execute the plan and replace the arg
+
+
         debug_msg(self.logger, "submit_func_to_workers")
         self.worker_intercomm.bcast(CommandType.EXEC_FUNCTION.value, self.bcast_root)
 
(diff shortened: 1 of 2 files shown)


bodo/spawn/spawner.py
```diff
@@ -443,9 +460,9 @@ def _get_arg_metadata(self, arg, arg_name, is_replicated, dist_flags):
         """
         dist_comm_meta = ArgMetadata.BROADCAST if is_replicated else ArgMetadata.SCATTER
         if isinstance(arg, BodoLazyWrapper):
+            dist_flags.append(arg_name)
             if arg._lazy:
                 return ArgMetadata.LAZY
-            dist_flags.append(arg_name)
```
IsaacWarren, line 448: I think this was a general spawn mode bug

bodo/tests/test_df_lib/test_jit_interop.py
```diff
@@ -14,6 +16,24 @@ def test_bodo_df_to_jit():
         }
     )
     bdf = bd.from_pandas(df)
+    return df, bdf
+
+
+def df_from_parquet():
+    path = datapath_util("example.parquet")
```
scott-routledge2, line 23:
```suggestion
    path = datapath_util("example_no_index.parquet")
```
This test will fail now since there is an index column, was silently passing before since we just ignored the index and it happened to match.

scott-routledge2 (approved):
LGTM, please see comment about the test it will fail now that we properly check for index columns in `read_parquet`.

ehsantn (approved):
Looks good to me. Thanks @IsaacWarren !

PR #410, opened 2025-04-16
[BSE-4737] Implement interface for using bodo as engine in Pandas UDFs.

scott-routledge2:

Subclass BaseExecutionEngine and implement BodoExecutionEngine, add as an attribute to `bodo.jit` / `return_wrapped_fn`.
See https://github.com/pandas-dev/pandas/pull/61032 for context.

As a followup we can also implement kwargs (but not key word only arguments, basically do something similar to numba where they fold keyword arguments into `args` using the function signature.)

Also need to do `Series.map` support, however that cannot be tested right now because it hasn't been implemented in Pandas.

## Testing strategy
In order to test, I needed to build Pandas from source. Here are the steps I took to test:
1. Remove pandas from environment i.e. `pixi remove pandas`. Also need to update pyproject.toml. Note pyproject.toml and pixi.lock/toml will be restored before merging.
2. Uninstall bodo, etc: `pixi run clean`
2. Clone pandas and install from source: `cd pandas; pip install .`
3. Build bodo: `pixi run build-bodo`

Ran PR CI for changes to jit decorator.

Run tests locally in spawn mode.

## User facing changes

Once Pandas 3 is released, users can use Bodo to accelerate their UDFs in Pandas.
e.g.
``` py
df.apply(func, axis=1, engine=bodo.jit, args=args)
```

## Checklist

### DIFF
--- a/bodo/pandas_compat.py
+++ b/bodo/pandas_compat.py
@@ -3,9 +3,8 @@ import inspect
 import warnings
 
 import numpy as np
-import pyarrow as pa
-
 import pandas as pd
+import pyarrow as pa
 
 pandas_version = tuple(map(int, pd.__version__.split(".")[:2]))
 
@@ -63,7 +62,6 @@ if pandas_version < (1, 4):
 # Pandas code: https://github.com/pandas-dev/pandas/blob/ca60aab7340d9989d9428e11a51467658190bb6b/pandas/core/arrays/string_arrow.py#L141
 def ArrowStringArray__init__(self, values):
     import pyarrow as pa
-
     from pandas.core.arrays.string_ import StringDtype
     from pandas.core.arrays.string_arrow import ArrowStringArray
 
--- a/bodo/pandas_compat.py
+++ b/bodo/pandas_compat.py
@@ -3,8 +3,9 @@ import inspect
 import warnings
 
 import numpy as np
-import pandas as pd
 import pyarrow as pa
+
+import pandas as pd
 
 pandas_version = tuple(map(int, pd.__version__.split(".")[:2]))
 
@@ -62,6 +63,7 @@ if pandas_version < (1, 4):
 # Pandas code: https://github.com/pandas-dev/pandas/blob/ca60aab7340d9989d9428e11a51467658190bb6b/pandas/core/arrays/string_arrow.py#L141
 def ArrowStringArray__init__(self, values):
     import pyarrow as pa
+
     from pandas.core.arrays.string_ import StringDtype
     from pandas.core.arrays.string_arrow import ArrowStringArray
 
@@ -247,7 +249,7 @@ def get_conversion_factor_to_ns(in_reso: str) -> int:
 
 
 # Class responsible for executing UDFs using Bodo as the engine in
-# newer version of pandas.
+# newer version of Pandas. See:
 # https://github.com/pandas-dev/pandas/pull/61032
 bodo_pandas_udf_execution_engine = None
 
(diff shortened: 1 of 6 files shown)


bodo/pandas_compat.py
```diff
@@ -244,3 +244,53 @@ def get_conversion_factor_to_ns(in_reso: str) -> int:
...
+if pandas_version >= (3, 0):
+    from collections.abc import Callable
+    from typing import Any
+
+    from pandas._typing import AggFuncType, Axis
+    from pandas.core.apply import BaseExecutionEngine
+
+    class BodoExecutionEngine(BaseExecutionEngine):
+        @staticmethod
+        def map(
```
scott-routledge2, line 263: This is not implemented yet in Pandas so there is no way to test. Leaving as a followup

pixi.toml
```diff
@@ -70,7 +70,6 @@ pip = "*"
 # Core Python Deps
 numba = "==0.61.0"
 numpy = ">=1.24,<2.2"
-pandas = ">=2.2,<2.3"
```
scott-routledge2, line 73: For testing locally, will restore before merging.

DrTodd13 (approved):
LGTM. Thanks.

pyproject.toml
```diff
@@ -37,7 +37,7 @@ classifiers = [
 dependencies = [
     "numba==0.61.0",
     "pyarrow==19.0.0",
-    "pandas>=2.2,<2.3",
+    "pandas>=2.2,<3.1",
```
ehsantn, line 40: Revert?

bodo/pandas_compat.py
```diff
@@ -244,3 +244,113 @@ def get_conversion_factor_to_ns(in_reso: str) -> int:
...
+
+            # Embed args as a string e.g. (args[0], args[1], ...) in func text
+            # to avoid typing issues with Bodo.
+            args_str = ""
+            if len(args):
+                args_str = ", ".join(f"args[{i}]" for i in range(len(args)))
+                args_str += ","
+            else:
+                # Add dummy value for args for spawn mode compatibility.
+                # TODO: fix in spawn mode.
```
scott-routledge2, line 338: I have a PR to fix this here: https://github.com/bodo-ai/Bodo/pull/414

bodo/pandas_compat.py
```diff
@@ -244,3 +244,113 @@ def get_conversion_factor_to_ns(in_reso: str) -> int:
...
+            apply_func_text += f"  return data.apply(udf, axis=axis, args=({args_str}))"
+
+            glbls = {"udf": func}
+            if spawn_mode:
+                # In the spawn mode case we need to bodo_exec on the workers as well
+                # so the code object is available to the caching infra.
+                def f(func_text, glbls, loc_vars, __name__):
+                    bodo_exec(func_text, glbls, loc_vars, __name__)
+
+            spawner.submit_func_to_workers(f, [], apply_func_text, glbls, {}, __name__)
```
ehsantn, line 351: Wrong indentation looks like. Should be inside `if spawn_mode`.

bodo/tests/test_pandas_udf_compat.py
```diff
@@ -0,0 +1,119 @@
...
+    bodo_out = df.apply(udf, axis=1, args=(4,), d=16, b="hi", engine=engine)
+
+    pandas_out = df.apply(udf, axis=1, args=(4,), d=16, b="hi")
+
+    _test_equal(bodo_out, pandas_out, check_pandas_types=False)
+
+
+def test_udf_cache():
+    """Tests that we can call the same UDF multiple times with cache flag on
+    without any errors. TODO: check cache."""
```
ehsantn, line 90: Have you verified caching manually at least?

ehsantn (approved):
Thanks @scott-routledge2 ! Looks good to me.

bodo/pandas_compat.py
scott-routledge2: Good catch, I just realized this codepath is taken even when `engine=bodo.jit(spawn=False, distributed=False)`. Would there be a way to also case on the decorator?

bodo/tests/test_pandas_udf_compat.py
scott-routledge2: Yes

bodo/pandas_compat.py
ehsantn: What do you mean?

scott-routledge2: It will exec the func text on the workers even when `spawn=False`. Discussed offline it would be better to just avoid func_text altogether.

bodo/pandas_compat.py
datapythonista: I'll open a PR adding `engine` to `Series.map` shortly. I have the implementation already, but map and apply tests are structured very differently, and I need to see how to implement the common fixtures without making the mess bigger or refactoring all the tests.

PR #412, opened 2025-04-17
Add general infrastructure for calling back to Python during plan execution

ehsantn:

Adds support for calling any Python function during plan execution to be able to reuse Pandas, Bodo JIT etc. Adds support for Series.str.lower as an example and also updates df.apply support.

## Testing strategy

Existing unit tests and added a new one.

## User facing changes

None.

## Checklist


bodo/pandas/utils.py
```diff
@@ -419,12 +419,49 @@ def impl_df_to_cpp_table(df):
...
+        input = input.iloc[:, 0]
+
+    if is_method:
+        func = input
+        for atr in func_path_str.split("."):
+            func = getattr(func, atr)
+        out = func(*args, **kwargs)
+    else:
+        # TODO: test this path
+        func = _get_function_from_path(func_path_str)
```
IsaacWarren, line 461: Why do we need this path, can't we always pass a method?

IsaacWarren (approved):
One question otherwise LGTM, thanks Ehsan

bodo/pandas/utils.py
ehsantn: I think we will create custom functions soon. I usually avoid "over-generalizing" but I think this will be used very soon.

DrTodd13 (approved):
LGTM.

PR #415, opened 2025-04-17
Improve iceberg write schema validation error message

IsaacWarren:
As title, this bugged me while working on a partitioned write example.

## Testing strategy
Locally and unit tests that check this error message.

## User facing changes
Better error messages

## Checklist

### DIFF
--- a/bodo/io/iceberg/write.py
+++ b/bodo/io/iceberg/write.py
@@ -263,7 +263,7 @@ def _update_field(
 
 def are_schemas_compatible(
     table_schema: pa.Schema, df_schema: pa.Schema, allow_downcasting: bool = False
-) -> bool:
+) -> tuple[bool, str | None]:
     """
     Check if the input DataFrame schema is compatible with the Iceberg table's
     schema for append-like operations (including MERGE INTO). Compatibility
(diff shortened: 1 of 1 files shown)


bodo/io/iceberg/write.py
```diff
@@ -316,8 +323,18 @@ def are_schemas_compatible(
...
         new_field = _update_field(df_field, pa_field, allow_downcasting)
         df_schema = df_schema.set(idx, new_field)
-
-    return df_schema.equals(table_schema)
+    equals = df_schema.equals(table_schema)
+    if not equals:
+        mismatched_fields = [
+            field.name
+            for idx, field in enumerate(df_schema)
+            if field.equals(table_schema.field(idx))
```
scott-routledge2, line 331: `if field.equals(table_schema.field(idx))` is True if the fields are not equal?

scott-routledge2 (approved):
LGTM, thanks @IsaacWarren

bodo/io/iceberg/write.py
IsaacWarren: Oops, thanks for catching, got confused when I was reading the test output because one string column was large_string and the other was string

srilman (approved):

PR #420, opened 2025-04-18
Support non-constant UDF args in df.apply()

ehsantn:
## Changes included in this PR

Adds support for `args` argument of `apply` to be a non-constant tuple.

## Testing strategy

Added unit tests.

## User facing changes

Non-constant args now supported.

## Checklist
- [ ] Pipelines passed before requesting review. To run CI you must include `[run CI]` in your commit message.
- [x] I am familiar with the [Contributing Guide](https://github.com/bodo-ai/Bodo/blob/main/CONTRIBUTING.md) 
- [x] I have installed + ran pre-commit hooks.

PR #422, opened 2025-04-21
Remove func text in Pandas UDF interface

scott-routledge2:
Simplify UDF interface implementation by removing func text.
Followup on: https://github.com/bodo-ai/Bodo/pull/420

## Testing strategy
Ran tests against Pandas 3 locally.

## Checklist


scott-routledge2:
> LGTM but why was the code originally written with func_text. Did something underlying change so that these are now 100% equivalent or are we giving something up with this change now?

Yes, before there was an issue with passing non-const tuples to df.apply, which was fixed in a recent PR. All the tests still pass and they should be equivalent now.

ehsantn (approved):
Thanks @scott-routledge2 .

PR #427, opened 2025-04-21
[WIP] Support Reading Dictionary<string> columns in Parquet Read

scott-routledge2:
* Handle dictionary case in `PhysicalResultCollector`
* Changes handling of Categorical Arrays in Parquet Reader to use DICT (only works with int32 indices at the moment)

TODO: Support Categorical types elsewhere, for example:
``` py
df = pd.DataFrame(
        {"A": pd.Categorical(["A", "B", "AB", "A", None, "B"] * 4 + [None, "C"])}
    )

print(bd.from_pandas(df))
```
Might be out of scope of this PR, requires DICT arrays have index types other than int32 (which I'm assuming would be a lot of effort) Or we would need to do some conversion in `arrow_dictionary_array_to_bodo`

## Testing strategy
New test added, which essentially does:
``` py
import pyarrow as pa
import pyarrow.parquet as pq

dictionary_values = pa.array(['apple', 'banana', 'cherry'])

indices = pa.array([0, 1, 2, 0, 1, 2, 1, 0], type=pa.int32())

dict_array = pa.DictionaryArray.from_arrays(indices, dictionary_values)

table = pa.Table.from_arrays([dict_array], names=['dict_col'])

pq.write_table(table, 'dictionary_column.parquet')
```
Read the file generated:
``` py
import bodo.pandas as bd

print(bd.read_parquet("dictionary_column.parquet"))
```

## Checklist


bodo/io/arrow_reader.cpp
```diff
@@ -873,6 +873,7 @@ TableBuilder::TableBuilder(std::shared_ptr<arrow::Schema> schema,
         if (arrow::is_primitive(type) || arrow::is_decimal(type) ||
             is_categorical) {
             if (is_categorical) {
+                std::cout << "Creating column from index type!!" << std::endl;
```
scott-routledge2, line 876: This is the root cause of the issue since we don't support categorical data anywhere in our New C++ code. This is the case that gets triggered by arrow type dictionary (which is how partition columns are represented).

bodo/io/arrow_reader.cpp
```diff
@@ -869,19 +870,9 @@ TableBuilder::TableBuilder(std::shared_ptr<arrow::Schema> schema,
...
-                columns.push_back(std::make_unique<PrimitiveBuilder>(
-                    dict_type->index_type(), num_rows, false, is_categorical));
-            } else {
-                columns.push_back(std::make_unique<PrimitiveBuilder>(
-                    field->type(), num_rows, nullable_field, is_categorical));
-            }
+        if (arrow::is_primitive(type) || arrow::is_decimal(type)) {
+            columns.push_back(std::make_unique<PrimitiveBuilder>(
+                field->type(), num_rows, nullable_field, false));
         } else if (arrow::is_string(type) &&
```
scott-routledge2, line 876: Note: It looks like Arrow coerces types of index to always be int32 for dictionary types when reading from parquet. https://github.com/apache/arrow/issues/27616

bodo/libs/_array.cpp
```diff
@@ -1,6 +1,8 @@
 #include <Python.h>
 #include <datetime.h>
 #include <iostream>
+#include <memory>
+#include <sstream>
```
scott-routledge2, line 5: TODO: remove unused includes

ehsantn (approved):

ehsantn:
@scott-routledge2 please review. I can't assign for review since you are the author.

scott-routledge2:
Looks good to me, thanks @ehsantn !

bodo/libs/_array.cpp
```diff
@@ -377,11 +378,37 @@ void info_to_numpy_array(array_info* info, uint64_t* n_items, char** data,
...
+        uint8_t* bitmap =
+            reinterpret_cast<uint8_t*>(codes_info->null_bitmask());
+        int32_t* codes_data = reinterpret_cast<int32_t*>(
+            codes_info->data1<bodo_array_type::NULLABLE_INT_BOOL>());
+        for (uint64_t i = 0; i < codes_info->length; ++i) {
+            if (!GetBit(bitmap, i)) {
+                codes_data[i] = -1;
+            }
+        }
+        info = new array_info(*codes_info);
```
IsaacWarren, line 398: Use a unique ptr so it deletes when it goes out of scope?

IsaacWarren (approved):

bodo/libs/_array.cpp
ehsantn: Explicit delete is a lot more clear in this particular case I think.

IsaacWarren: IMO it's better with a comment and a unique ptr than a new variable and conditional but not a big deal

PR #441, opened 2025-04-24
Add more Series methods

scott-routledge2:
Small bug fix and two new methods: Series.map and Series.str.strip
Also changes how we get the Dtype of sample output in UDF to avoid typing issues with `object`. The new Pandas version that used pyarrow by default should fix this.

Demo:
``` py
import bodo.pandas as bd
import pandas as pd

def prompt(prompt):
    return "hi there"

def query_model(prompt_):
    """
    Sends a prompt to the AI Suite and returns the response.
    """
    response = prompt(prompt_)
    return response

bdf = bd.read_parquet("prompts.parquet")

cleaned_prompts = bdf.prompts.str.lower().str.strip()

responses = cleaned_prompts.map(query_model)

print(responses)
```

## Testing strategy
Added unit tests for new methods.

## Checklist

### DIFF
--- a/bodo/pandas/frame.py
+++ b/bodo/pandas/frame.py
@@ -577,11 +577,13 @@ class BodoDataFrame(pd.DataFrame, BodoLazyWrapper):
         self._mgr._plan = mgr_plan
         out_sample = df_sample.apply(func, axis)
 
-        if isinstance(out_sample, pd.Series):
-            out_sample = out_sample.to_frame()
+        assert isinstance(out_sample, pd.Series), (
+            "BodoDataFrame.apply(), expected output to be Series."
+        )
+        out_sample = out_sample.to_frame()
         empty_df = arrow_to_empty_df(pa.Schema.from_pandas(out_sample))
-
-        # empty_df = out_sample.iloc[:0]
+        # convert back to Series
+        empty_df = empty_df.squeeze()
 
         plan = LazyPlan(
             "LogicalProjectionPythonScalarFunc",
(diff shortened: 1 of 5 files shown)


bodo/pandas/frame.py
```diff
@@ -573,7 +577,11 @@ def apply(
         self._mgr._plan = mgr_plan
         out_sample = df_sample.apply(func, axis)
 
-        empty_df = out_sample.iloc[:0]
+        if isinstance(out_sample, pd.Series):
+            out_sample = out_sample.to_frame()
+        empty_df = arrow_to_empty_df(pa.Schema.from_pandas(out_sample))
```
scott-routledge2, line 582: Workaround for now since an empty df will be typed as "object" which is equivalent to arrow type NULL

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -255,3 +255,43 @@ def test_str_lower(datapath, index_val):
     assert out_bodo._lazy
     assert out_bodo.plan is not None
     _test_equal(out_bodo, out_pd, check_pandas_types=False)
+
+
+def test_str_strip(datapath, index_val):
```
scott-routledge2, line 260: Once we add more string methods we could probably add a method fixture to test them more concisely.

DrTodd13 (approved):
LGTM.  Thanks.

ehsantn (approved):
Looks good to me. Thanks @scott-routledge2.

PR #446, opened 2025-04-28
Multiple filter with string support.

DrTodd13:

Multiple expression filter support.
String support.

## Testing strategy

run_ci
## User facing changes

None

## Checklist


bodo/pandas/physical/filter.h
```diff
@@ -45,6 +45,23 @@ class PhysicalFilter : public PhysicalSourceSink {
                 "Filter expression tree did not result in a boolean array");
         }
 
+        if (!input_batch->column_names[input_batch->column_names.size() - 1]
```
ehsantn, line 48: This has to be a proper projection in the plan I believe. Also, needs to use the metadata in `input_batch->metadata` to know if it's a RangeIndex.

bodo/pandas/utils.py
```diff
@@ -329,18 +329,66 @@ def generate_duckdb(self, cache=None):
...
             if isinstance(x, LazyPlan):
-                return x.generate_duckdb(cache=cache)
+                return x.generate_duckdb(cache=cache, in_filter=in_filter)
             else:
                 return x
 
-        # Convert any LazyPlan in the args or kwargs.
-        args = [recursive_check(x) for x in self.args]
-        kwargs = {k: recursive_check(v) for k, v in self.kwargs.items()}
+        # Description of LogicalProjectionOrColRef node type.
```
ehsantn, line 339: It'd be best if we could avoid this workaround. What is the specific example that is problematic? Let's discuss.

ehsantn (commented):
Took a quick look and added some comments. Would be happy to discuss.

bodo/pandas/utils.py
DrTodd13: Is my extensive comment there not clear?  Projection nodes take ownership of unique_ptr sources.  So, if you have two projection in one filter they both will try to claim ownership and the second one can't possibly get it because the source node (cython obj) that it knows about no longer has a reference to the owned unique_ptr.  It is just an impedance mismatch between the two APIs and yet another layer of lazy.

bodo/pandas/physical/filter.h
DrTodd13: Removed.

bodo/pandas/series.py
```diff
@@ -59,6 +59,57 @@ def _cmp_method(self, other, op):
             plan=LazyPlan("LogicalBinaryOp", self._plan, other, op),
         )
 
+    def _conjunction_binop(self, other, op):
+        """Called when a BodoSeries is element-wise boolean combined with a different entity (other)"""
+        from bodo.pandas.base import _empty_like
+
+        # Get empty Pandas objects for self and other with same schema.
+        zero_size_self = _empty_like(self)
+        zero_size_other = _empty_like(other) if isinstance(other, BodoSeries) else other
```
ehsantn, line 68: I think this should check for `other` to be the supported values more narrowly to avoid failures. It could be array, unsupported scalars etc.

bodo/pandas/utils.py
ehsantn: Let's merge this for now. I'm wondering if we can extract the expression and avoid the logical operators with unique pointers.

ehsantn (approved):
Thanks @DrTodd13. Looks good to me overall. I think we can revisit some infrastructure issues like trying to use Arrow compute later.

IsaacWarren (approved):
Thanks Todd

bodo/pandas/_plan.h
```diff
@@ -238,6 +238,22 @@ make_projection_python_scalar_func(
...
+ */
+duckdb::unique_ptr<duckdb::Expression> make_const_float_expr(float val);
+
+/**
+ * @brief Create an expression from a constant string.
+ *
+ * @param val - the constant string for the expression
+ * @return duckdb::unique_ptr<duckdb::Expression> - the const string expr
+ */
+duckdb::unique_ptr<duckdb::Expression> make_const_string_expr(std::string val);
```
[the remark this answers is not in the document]
IsaacWarren: I meant the make_const_x_expr functions

PR #456, opened 2025-05-02
Enable read_parquet tests for DataFrame library

scott-routledge2:
## Checklist

### DIFF
--- a/.github/workflows/_test_python_source.yml
+++ b/.github/workflows/_test_python_source.yml
@@ -115,6 +115,7 @@ jobs:
             PYTEST_MARKER: ${{ inputs.pytest-marker }}
             BODO_TEST_SPAWN_MODE: ${{ inputs.pytest-marker == 'spawn_mode' && '1' || '0' }}
             BODO_ENABLE_DATAFRAME_LIBRARY: ${{ inputs.pytest-marker == 'df_lib' && '1' || '0' }}
+            BODO_ENABLE_TEST_DATAFRAME_LIBRARY: ${{ inputs.pytest-marker == 'df_lib' && '1' || '0' }}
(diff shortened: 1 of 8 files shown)


bodo/pandas/vendor/duckdb/src/common/types.cpp
```diff
@@ -344,7 +344,7 @@ idx_t GetTypeIdSize(PhysicalType type) {
 	case PhysicalType::LIST:
 		return sizeof(list_entry_t); // offset + len
 	default:
-		throw InternalException("Invalid PhysicalType for GetTypeIdSize");
+		throw InternalException("Invalid PhysicalType for GetTypeIdSiz_______" + to_string((int)type));
```
scott-routledge2, line 347: Debugging, todo: remove

.github/workflows/_test_python_source.yml
```diff
@@ -115,6 +115,7 @@ jobs:
             PYTEST_MARKER: ${{ inputs.pytest-marker }}
             BODO_TEST_SPAWN_MODE: ${{ inputs.pytest-marker == 'spawn_mode' && '1' || '0' }}
             BODO_ENABLE_DATAFRAME_LIBRARY: ${{ inputs.pytest-marker == 'df_lib' && '1' || '0' }}
```
IsaacWarren, line 117: Can we get rid of this now?

IsaacWarren (approved):
Thanks Scott, can you create and link an issue to check filter conditions in plans too, seems like something that needs confirmed in tests

.github/workflows/_test_python_source.yml
scott-routledge2: Still need to set this in CI scripts until it is turned on by default.

PR #463, opened 2025-05-06
Support setting DataFrame column

ehsantn:

Adds support for setting a dataframe column with a Series plan that is created from the same dataframe. This works by creating a new projection plan that adds the data columns of the original dataframe as well as the new column from the Series. This PR also fixes a bug when chaining multiple Series function calls.

I think the plan manipulation code is hard to read and potentially fragile because of not having proper Python classes for the plan. We need to discuss the path forward if we are going to have a lot of these plan handling cases.

## Testing strategy

Added tests.

## User facing changes

Setting dataframe column now works.

## Checklist

### DIFF
--- a/bodo/pandas/frame.py
+++ b/bodo/pandas/frame.py
@@ -575,7 +575,6 @@ class BodoDataFrame(pd.DataFrame, BodoLazyWrapper):
         This is done by creating a new plan that add the new
         column in the existing dataframe plan using a projection.
         """
-        from bodo.pandas.base import _empty_like
 
         # Match cases like df["B"] = df["A"].str.lower()
         if (
@@ -584,12 +583,9 @@ class BodoDataFrame(pd.DataFrame, BodoLazyWrapper):
             and isinstance(value, BodoSeries)
             and value.is_lazy_plan()
         ):
-            if (new_plan := _get_set_column_plan(self._plan, value._plan)) is not None:
-                # Set plan metadata
-                new_metadata = _empty_like(self)
-                new_metadata[key] = _empty_like(value)
-                new_plan.out_schema = new_metadata
-                new_plan.output_func = cpp_table_to_df
+            if (
+                new_plan := _get_set_column_plan(self._plan, value._plan, key)
+            ) is not None:
                 # Update internal state
                 self.plan = new_plan
                 self._mgr._plan = new_plan
@@ -689,6 +685,7 @@ class BodoDataFrame(pd.DataFrame, BodoLazyWrapper):
 def _get_set_column_plan(
     df_plan: LazyPlan,
     value_plan: LazyPlan,
+    key: str,
 ) -> LazyPlan | None:
     """
     Get the plan for setting a column in a dataframe or return None if not supported.
@@ -713,4 +710,9 @@ def _get_set_column_plan(
         df_plan,
         tuple(exprs),
     )
+    # Set plan metadata
+    new_metadata = df_plan.out_schema.copy()
+    new_metadata[key] = value_plan.out_schema.copy()
+    new_plan.out_schema = new_metadata
+    new_plan.output_func = cpp_table_to_df
     return new_plan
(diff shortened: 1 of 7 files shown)


bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -487,6 +505,50 @@ def func(x):
     _test_equal(out_bodo, out_pd, check_pandas_types=False)
 
 
+def test_set_column(datapath, index_val, set_stream_batch_size_three):
+    """Very simple test for Series.str.strip() for sanity checking."""
```
IsaacWarren, line 509: Docstring

bodo/pandas/frame.py
```diff
@@ -566,6 +569,35 @@ def __getitem__(self, key):
...
 
+    @check_args_fallback("none")
+    def __setitem__(self, key, value) -> None:
+        """Supports setting columns (df[key] = value) when value is a Series created
+        from the same dataframe.
+        This is done by creating a new plan that add the new
+        column in the existing dataframe plan using a projection.
+        """
+
+        # Match cases like df["B"] = df["A"].str.lower()
```
IsaacWarren, line 580: I'm assuming the multi-column case is a followup, do we have an issue?

DrTodd13 (approved):
LGTM.

bodo/pandas/frame.py
```diff
@@ -566,6 +569,35 @@ def __getitem__(self, key):
...
+                # Update internal state
+                self.plan = new_plan
+                self._mgr._plan = new_plan
+                head_val = value._head_s
+                self._head_df[key] = head_val
+                with self.disable_collect():
+                    super().__setitem__(key, head_val)
+                return
+
+        return super().__setitem__(key, value)
```
IsaacWarren, line 599: What happens here if value isn't a bodoseries? it collects?

bodo/pandas/frame.py
```diff
@@ -566,6 +569,35 @@ def __getitem__(self, key):
...
+            if (
+                new_plan := _get_set_column_plan(self._plan, value._plan, key)
+            ) is not None:
+                # Update internal state
+                self.plan = new_plan
+                self._mgr._plan = new_plan
+                head_val = value._head_s
+                self._head_df[key] = head_val
+                with self.disable_collect():
+                    super().__setitem__(key, head_val)
```
IsaacWarren, line 596: Why is this necessary?

bodo/pandas/frame.py
```diff
@@ -637,3 +669,147 @@ def apply(
...
+    input data column index.
+    """
+    in_col_ind = func_expr.args[2][0]
+    n_source_cols = len(new_source_plan.out_schema.columns)
+    index_cols = tuple(
+        range(
+            n_source_cols,
+            n_source_cols + get_n_index_arrays(new_source_plan.out_schema.index),
+        )
+    )
```
IsaacWarren, line 699: Some comments here would be good

IsaacWarren (approved):
Thanks Ehsan, my main question/concern is around the need for the super setitem and disabling collection, otherwise LGTM

bodo/pandas/frame.py
ehsantn: Created a follow up issue.

bodo/pandas/frame.py
ehsantn: Yes. Basically, falls back for everything else.

bodo/tests/test_df_lib/test_end_to_end.py
ehsantn: Done.

bodo/pandas/frame.py
ehsantn: Updates internal data manager (e.g. add a new block). Added a comment.

bodo/pandas/frame.py
ehsantn: Done.

PR #470, opened 2025-05-08
Non-pushdown limit.

DrTodd13:

Added mpi_allgather implementation of non-pushdown limit.  Changed pipeline processing so that returning finish terminates the pipeline and have_more_input actually does the right thing of calling the operator again.

## Testing strategy

run_ci
## User facing changes

None

## Checklist


bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -440,18 +471,13 @@ def test_head(datapath):
     )
 
     bodo_df2 = bodo_df1.head(3)
-    py_df2 = py_df1.head(3)
+    py_df1.head(3)
```
ehsantn, line 474: Remove line.

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -425,7 +429,34 @@ def test_head_pushdown(datapath):
...
+@pytest.mark.skip(reason="Not working.")
+def test_projection_head_pushdown(datapath):
+    """Test for head pushed down to read parquet."""
+    bodo_df1 = bd.read_parquet(datapath("dataframe_library/df1.parquet"))
+    bodo_df2 = bodo_df1["D"]
+    bodo_df3 = bodo_df2.head(3)
+
+    # Make sure bodo_df2 is unevaluated at this point.
+    assert bodo_df3.is_lazy_plan()
+    assert bodo_df3.plan is not None
```
ehsantn, line 441: Remove redundant check.
```suggestion
```

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -425,7 +429,34 @@ def test_head_pushdown(datapath):
...
+    py_df2 = py_df1["D"]
+    py_df3 = py_df2.head(3)
+
+    _test_equal(
+        bodo_df3,
+        py_df3,
+        check_pandas_types=False,
+        sort_output=True,
+        reset_index=True,
+    )
```
ehsantn, line 457: The Pandas `head` result won't be equal.
```suggestion
```

bodo/pandas/_pipeline.cpp
```diff
@@ -2,36 +2,69 @@
 
 #include "physical/result_collector.h"
 
+/*
+ * This has to be a recursive routine.  Each operator in the pipeline could
+ * say that it HAVE_MORE_OUTPUT in which case we need to call it again for the
+ * same input batch (or an empty input batch) once we've call the subsequent
```
ehsantn, line 8: I think using same input batch can be dangerous depending on operator behavior and empty input batch is safer.

bodo/pandas/_pipeline.cpp
```diff
@@ -2,36 +2,69 @@
 
 #include "physical/result_collector.h"
 
+/*
+ * This has to be a recursive routine.  Each operator in the pipeline could
```
ehsantn, line 6: I assume this is at least inspired by DuckDB. Could you add links to the corresponding DuckDB codes for reference?

bodo/pandas/_pipeline.cpp
```diff
@@ -2,36 +2,69 @@
...
+            // Process this batch with this operator.
+            std::pair<std::shared_ptr<table_info>, OperatorResult> result =
+                op->ProcessBatch(batch);
+            OperatorResult op_result = result.second;
+
+            // Execute subsequent operators and If any of them said that
+            // no more output is needed or the current operator knows no
+            // more output is needed then return true to terminate the pipeline.
+            if (midPipelineExecute(idx + 1, result.first) ||
+                op_result == OperatorResult::FINISHED) {
```
ehsantn, line 32: The `op_result == OperatorResult::FINISHED` check has to be first to short circuit? Otherwise early termination doesn't happen.

bodo/pandas/physical/limit.h
```diff
@@ -20,22 +21,43 @@ class PhysicalLimit : public PhysicalSourceSink {
...
+        int myrank = 0;
+        MPI_Comm_size(MPI_COMM_WORLD, &n_pes);
+        MPI_Comm_rank(MPI_COMM_WORLD, &myrank);
 
-        std::vector<int64_t> rowInds(select_this_time);
-        for (uint64_t i = 0; i < select_this_time; ++i) {
+        // Gather how many rows are on each rank
+        std::vector<uint64_t> row_counts(n_pes);
+        uint64_t cur_rows = input_batch->nrows();
+        CHECK_MPI(MPI_Allgather(&cur_rows, 1, MPI_UNSIGNED_LONG_LONG,
```
ehsantn, line 38: Using blocking collectives in the middle of pipeline loop is not safe since ranks may have different number of iterations (e.g. some rank has more data). I think using async collectives is the proper approach but to limit the scope right now, each rank can generate `n` rows at most and move this code to the finalize function of the operator.

ehsantn (commented):
Thanks for the effort @DrTodd13 . The main problem is that using blocking collectives inside the streaming loop is not safe.

bodo/tests/test_df_lib/test_end_to_end.py
DrTodd13: done

bodo/tests/test_df_lib/test_end_to_end.py
DrTodd13: done

bodo/tests/test_df_lib/test_end_to_end.py
DrTodd13: done

bodo/pandas/_pipeline.cpp
DrTodd13: Inspired by listening to some videos of people talking about the requirements but not from looking at the code.

bodo/pandas/_pipeline.cpp
DrTodd13: No, I think this order is correct.  You still need to propagate the current batch through the rest of the pipeline to get to the sink.  So, "finished" means that we don't need to generate more output AFTER THIS ONE but the current one needs to go through the pipeline to get to the output.

bodo/pandas/physical/limit.h
DrTodd13: We should talk.

bodo/pandas/_pipeline.cpp
ehsantn: Is finding the relevant code too difficult?

bodo/pandas/_executor.h
```diff
@@ -23,19 +23,25 @@ class Executor {
         PhysicalPlanBuilder builder;
         builder.Visit(*plan);
         pipelines = std::move(builder.finished_pipelines);
-
-        if (builder.active_pipeline != nullptr) {
-            pipelines.push_back(builder.active_pipeline->BuildEnd(out_schema));
-        }
+        assert(builder.active_pipelines.size() == 1);
```
ehsantn, line 26: What does multiple active pipelines mean? Why is it necessary? There is only one pipeline that is being built at a time. Operators like LIMIT will just finish a pipeline and start a new active pipeline. We are following DuckDB's code and I think we should keep doing it.

bodo/pandas/_executor.h
```diff
@@ -23,19 +23,25 @@ class Executor {
...
-        // TODO: support multiple pipelines
-        pipelines[0]->Execute();
-
-        return pipelines[0]->GetResult();
+        // We must execute all pipelines once.
+        // Though there are dependencies between them
+        // the method of construction should guarantee that
+        // earlier items in the vector are dependencies to
+        // later items so we can still process them in order.
+        for (size_t i = 0; i < pipelines.size(); ++i) {
```
ehsantn, line 40: Why do we need to keep track of dependencies in pipelines? Does DuckDB do it? Pipelines just need to run one at a time in order in regular SQL I believe.

ehsantn (changes requested):
I think we should follow regular vectorized SQL conventions and follow DuckDB's code for pipelines as closely as possible to avoid surprises down the line (both correctness and performance).

bodo/pandas/_executor.h
DrTodd13: Yes, limit doesn't need multiple active pipelines but joins do.  It is a post-order traversal of the graph and you start a pipeline at one leaf that carries through to the root of the tree but when you get to a node with more than one child then one of those children will be in the main pipeline but the other has to form its own pipeline.  The original pipeline say with the left sub-tree is still active in the sense that you will add more to it eventually but while you are traversing the right sub-tree you will have a different active pipeline (top of the stack).

bodo/pandas/_executor.h
DrTodd13: Here is the way I see it.  Some pipelines do depend on the results of other pipelines (if they didn't then they'd be separate queries).  So, those pipelines need a way to get the results from their dependent pipelines.  There's probably multiple ways of arranging to do that but we already had the ability to query a pipeline to get its results (i.e., GetResults) so if you just have a reference to the pipelines you are dependent on then you can get their results as needed.

bodo/pandas/_executor.h
DrTodd13: To be clear, this is talking about pipeline formation, not pipeline execution.  The loop that executes pipelines does go one at a time.

bodo/pandas/_pipeline.cpp
DrTodd13: From what I can tell, duckdb passes the same batch. src/parallel/pipeline_executor.cpp

`			// Operator(s) in the pipeline have returned `HAVE_MORE_OUTPUT` in the last Execute call
			// the operators have to be called with the same input chunk to produce the rest of the output
			D_ASSERT(source_chunk.size() > 0);
			result = ExecutePushInternal(source_chunk, chunk_budget);`

bodo/pandas/physical/operator.h
```diff
@@ -17,10 +17,10 @@ enum class OperatorType : uint8_t {
...
 
 /// Specifies the status of the physical operator in the execution:
-/// 1. NEED_MORE_INPUT means the operator needs more input to produce output
-/// 2. HAVE_MORE_OUTPUT means the operator has more output to produce in later
-/// iterations
-/// 3. FINISHED means the operator has finished processing
+/// 1. NEED_MORE_INPUT means the operator is ready for additional input
+/// 2. HAVE_MORE_OUTPUT means the operator can produce more output without
+/// additional input.
+/// 3. FINISHED means the pipeline should terminate
```
IsaacWarren, line 23: Do we need a different value for the operator is done? I guess we can add it if we need it

bodo/pandas/physical/limit.h
```diff
@@ -3,41 +3,134 @@
...
-        for (uint64_t i = 0; i < select_this_time; ++i) {
-            rowInds[i] = i;
-        }
-        std::shared_ptr<table_info> out_table_info =
-            RetrieveTable(input_batch, rowInds);
-        return {out_table_info, OperatorResult::FINISHED};
+    /**
+     * @brief GetResult - just for API compatability but should never be called
+     */
+    std::shared_ptr<table_info> GetResult() {
```
IsaacWarren, line 111:
```suggestion
    std::shared_ptr<table_info> GetResult() override {
```

bodo/pandas/physical/limit.h
```diff
@@ -3,41 +3,134 @@
...
+        // all in one go.
+        throw std::runtime_error("GetResult called on a limit node.");
+    }
+
+    /**
+     * @brief ProduceBatch - act as a data source
+     *
+     * returns std::pair<std::shared_ptr<table_info>, ProducerResult>
+     */
+    std::pair<std::shared_ptr<table_info>, ProducerResult> ProduceBatch() {
```
IsaacWarren, line 123:
```suggestion
    std::pair<std::shared_ptr<table_info>, ProducerResult> ProduceBatch() override {
```

bodo/pandas/physical/limit.h
```diff
@@ -3,41 +3,134 @@
...
+                return acc + item->nrows();
+            });
+        CHECK_MPI(MPI_Allgather(&cur_rows, 1, MPI_UNSIGNED_LONG_LONG,
+                                row_counts.data(), 1, MPI_UNSIGNED_LONG_LONG,
+                                MPI_COMM_WORLD),
+                  "PhysicalLimit: MPI error on MPI_Allgather:");
+
+        uint64_t select_local = 0;
+
+        for (int64_t i = 0; i < n_pes && n > 0; ++i) {
```
IsaacWarren, line 49: Very minor comment, feel free to ignore but i could be a uint16

bodo/pandas/physical/limit.h
```diff
@@ -3,41 +3,134 @@
...
+        return {collected_rows[produced - 1],
+                produced >= collected_rows.size()
+                    ? ProducerResult::FINISHED
+                    : ProducerResult::HAVE_MORE_OUTPUT};
     }
 
    private:
-    uint64_t n;
+    uint64_t n, local_remaining, produced;
+    std::vector<std::shared_ptr<table_info>> collected_rows;
```
IsaacWarren, line 135: This should probably be a chunked table builder

IsaacWarren (approved):
Thanks Todd, only blocking comment is the chunked table builder for storing rows in PhysicalLimit

bodo/pandas/physical/limit.h
IsaacWarren: This and the other missing override were preventing me from building on my mac

bodo/pandas/physical/sample.h
```diff
@@ -0,0 +1,55 @@
+
+#pragma once
+
+#include <memory>
+#include <utility>
+#include "../libs/_array_utils.h"
+#include "operator.h"
+
+/**
+ * @brief Physical node for projection.
```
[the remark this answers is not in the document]
DrTodd13: Done.

bodo/pandas/physical/sample.h
```diff
@@ -0,0 +1,55 @@
...
+        for (uint64_t i = 0; i < select_this_time; ++i) {
+            rowInds[i] = i;
+        }
+        std::shared_ptr<table_info> out_table_info =
+            RetrieveTable(input_batch, rowInds);
+        return {out_table_info, OperatorResult::NEED_MORE_INPUT};
+    }
+
+   private:
+    float percentage;
```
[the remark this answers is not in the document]
DrTodd13: Done.

bodo/pandas/physical/operator.h
DrTodd13: What do you mean by "done?"  In the normal case where the operator has done its thing on a given input batch and it is ready for another batch then it returns NEED_MORE_INPUT.  Perhaps that would be better called READY_FOR_INPUT.

bodo/pandas/physical/limit.h
DrTodd13: done.

bodo/pandas/physical/operator.h
IsaacWarren: Good point

bodo/pandas/physical/limit.h
DrTodd13: Reduced to 32 bit but signed/unsigned comparison in the look keeping iterator signed.

bodo/pandas/physical/limit.h
DrTodd13: Thanks for the suggestion.  I was looking for something like this for a while but here is what I'm thinking now.  If we get in the situation where to get N rows on one rank we need multiple batches (which is already pretty unlikely) then when the limit node is acting as a source for the next pipeline then we still want to follow the batch size specification which is what we already have in the separate saved table_infos in collected_rows.  Yes, we could form one table_info using chunked table builder but then when we turn around and have the limit node act as a source then we'd have to break it up again into batches anyway.  So, to me it just seems like overhead.  Am I missing a reason to do this anyway?

IsaacWarren: Under the hood the chunked table builder also stores a list of table infos essentially so there shouldn't be overhead associated with turning it back into batches. Also it transparently supports spilling the buffer once we setup that infra

ehsantn (approved):
Looks good to me. The main comment is about using chunked table builder to simplify the code.

PR #472, opened 2025-05-08
Add test-type to PR CI

scott-routledge2:
## Checklist

### DIFF
--- a/.github/workflows/_test_python_source.yml
+++ b/.github/workflows/_test_python_source.yml
@@ -23,6 +23,11 @@ on:
         required: false
         default: 'ubuntu-latest'
         type: string
+      # Options: DF_LIB, SPAWN, NORMAL
+      test-type:
+        type: string
+        description: The kind of tests to run e.g. spawn tests
+        required: true
 
 jobs:
   run:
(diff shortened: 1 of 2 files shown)


.github/workflows/_test_python_source.yml
```diff
@@ -23,6 +23,11 @@ on:
         required: false
         default: 'ubuntu-latest'
         type: string
+      # Options: DF_LIB, SPAWN, NORMAL
+      test-type:
+        type: string
```
IsaacWarren, line 28: I think there's an ENUM type so it will create a dropdown with just those options

IsaacWarren (approved):
Thanks Scott

hadia206 (approved):

.github/workflows/_test_python_source.yml
scott-routledge2: That's what I had initially and I was getting `invalid value "choice" for input type of workflow_call event. it must be one of "boolean", "number", or "string"` when trying to commit. I think the dropdown is just for actions.

IsaacWarren: Thanks for checking

PR #479, opened 2025-05-12
[BSE-4785] Project after filter

scott-routledge2:
Reuse column bindings from source node instead of getting a table index. See [here](https://github.com/bodo-ai/Bodo/blob/66193172a1c125b3300bf242c7908e8f38e413c7/bodo/pandas/vendor/duckdb/src/planner/binder/query_node/plan_setop.cpp#L76) as another example. 

One thing to watch out for is ensuring the col_idx is always the correct index into the Column Bindings (e.g. a ProjectionMap)

## Testing strategy
New test added, confirmed it fails on Main.
DF lib tests pass locally (except certain read_parquet test)

## User facing changes
Users can now do projections after filters.

## Checklist

### DIFF
--- a/.github/workflows/_test_python_source.yml
+++ b/.github/workflows/_test_python_source.yml
@@ -119,7 +119,9 @@ jobs:
             BODO_TESTING_ONLY_RUN_1D_VAR: true
             PYTEST_MARKER: ${{ inputs.pytest-marker }}
             BODO_TEST_SPAWN_MODE: ${{ inputs.test-type == 'SPAWN' && '1' || '0' }}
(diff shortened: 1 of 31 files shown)


DrTodd13 (approved):
My only comment is that getting the column bindings and indexing them is now duplicated twice.  I had a function I was using in those places and that could just get the column bindings and index.  It is pretty short so I don't know if it is worth it to make it a function.  What do you think?

scott-routledge2:
> My only comment is that getting the column bindings and indexing them is now duplicated twice. I had a function I was using in those places and that could just get the column bindings and index. It is pretty short so I don't know if it is worth it to make it a function. What do you think?

I think it's ok for now since it's only 2 lines, we can always refactor later

PR #483, opened 2025-05-12
[BSE-4798] Enable Series.map tests for DataFrame Library

scott-routledge2:

Marks relevant compiler tests for map as Df_lib. Link issues to failing tests, PR CI and bug fixes.

The main issues I am noticing with this implementation is that:
1. It is not very robust when functions occasionally return None, if map(head(1)) returns None then it just fails
2. The conversion via `convert_dtypes` is buggy. Sometimes it is a little too aggressive with the conversion (i.e. float gets downcasted to int) and in other cases different workers can have different types (for example one worker might have all None output data while the other worker doesn't.
3. Unsupported dtypes: Categorical, duration[ns] and Decimal.
## Testing strategy
Ran DF library tests locally. PR CI for compiler tests.

## User facing changes
None

## Checklist

### DIFF
--- a/.github/workflows/_test_python_source.yml
+++ b/.github/workflows/_test_python_source.yml
@@ -69,9 +69,6 @@ jobs:
         # Run Tests
         - name: Run Tests
           run: |
-            export SPARK_LOCAL_IP=127.0.0.1
-            export PYSPARK_SUBMIT_ARGS="--conf spark.driver.bindAddress=127.0.0.1 --conf spark.driver.host=127.0.0.1 pyspark-shell"
-
             if [[ "$RUNNER_OS" == "Windows" ]]; then
               # On Windows we set HADOOP_HOME to a dummy directory.
               # Spark needs to verify HADOOP_HOME exists at initialization even if it is not used.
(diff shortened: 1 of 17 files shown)


bodo/tests/test_dataframe_part2.py
```diff
@@ -584,8 +584,7 @@ def test_impl(df):
         {"A": ["AA", "B", "CC", "C", "AA"], "B": [3, 1, 2, 5, 9]}, index=[3, 1, 4, 6, 0]
     )
 
-    check_pandas_types = not bodo.test_dataframe_library_enabled
-    check_func(test_impl, (df,), check_pandas_types=check_pandas_types)
+    check_func(test_impl, (df,))
```
scott-routledge2, line 587: Just refactoring here since we disable check_pandas_types by default in df lib version of check func

.github/workflows/_test_python_source.yml
```diff
@@ -127,6 +127,12 @@ jobs:
             BODO_SPAWN_MODE: "0"
             BODO_BUFFER_POOL_REMOTE_MODE: "1"
             BODO_BUFFER_POOL_DEBUG_MODE: "1"
+            # Pyspark Config
```
scott-routledge2, line 130: https://github.com/bodo-ai/Bodo/actions/runs/15003331406/job/42155711176?pr=483

.github/workflows/pr_ci.yml
```diff
@@ -83,11 +83,12 @@ jobs:
...
-      - name: Build from Source
-        uses: ./.github/actions/build-source
-        with:
-          build-all: false
+      # TODO: Reenable and fix Sccache
+      # - uses: actions/checkout@v4
+      # - name: Build from Source
+      #   uses: ./.github/actions/build-source
+      #   with:
+      #     build-all: false
```
scott-routledge2, line 91: Skipping this step can save 30 minutes and is not necessary until we get Cache working IMO. (We'd just have to check it compiles with Bodo arrow fork?)

IsaacWarren (approved):
Thanks Scott!

ehsantn (approved):

PR #484, opened 2025-05-12
BSE-4790: Fix parallel get data scope

IsaacWarren:
Fixes an issue where parallel gets could try to read an entry from the result registry that's already been deleted. Resolved by attaching the source lazymanager to the lazy plan. This introduced some complications in sending the plan to the workers because the managers need to be removed first so collection isn't triggered.

## Testing strategy
Unittests locally/PR CI

## User facing changes
Fix the title bug

## Checklist

### DIFF
--- a/bodo/pandas/base.py
+++ b/bodo/pandas/base.py
@@ -31,9 +31,9 @@ def from_pandas(df):
     res_id = None
     if bodo.dataframe_library_run_parallel:
         res_id = bodo.spawn.utils.scatter_data(df)
-        plan = LazyPlan("LogicalGetPandasReadParallel", empty_df, [], res_id)
+        plan = LazyPlan("LogicalGetPandasReadParallel", empty_df, res_id)
     else:
-        plan = LazyPlan("LogicalGetPandasReadSeq", empty_df, [], df)
+        plan = LazyPlan("LogicalGetPandasReadSeq", empty_df, df)
 
     return wrap_plan(plan=plan, nrows=n_rows, res_id=res_id)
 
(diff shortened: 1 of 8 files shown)


DrTodd13 (changes requested):
Can instead of storing res_id in the _plan case, can you store (self._mgr, and res_id)?  Then, you don't need this parameter that is [] almost everywhere?  Then, extract res_id from the tuple in the right places?

scott-routledge2 (approved):
Thanks @IsaacWarren . Can we create a test for this?

IsaacWarren:
> Can instead of storing res_id in the _plan case, can you store (self._mgr, res_id)? Then, you don't need this parameter that is [] almost everywhere? Then, extract res_id from the tuple in the right places?

I spent a day trying to get this to work. I can try to clean it up in my followup to fix scattered data never being deleted

ehsantn:
@IsaacWarren I'm wondering if the code can be simpler and avoid changing all other code. Why can't the manager be just attached to the parallel read node and nothing else? Let's discuss sometime.

bodo/pandas/base.py
```diff
@@ -32,7 +33,11 @@ def from_pandas(df):
     res_id = None
     if bodo.dataframe_library_run_parallel:
         res_id = bodo.spawn.utils.scatter_data(df)
-        plan = LazyPlan("LogicalGetPandasReadParallel", empty_df, res_id)
+        plan = LazyPlan(
+            "LogicalGetPandasReadParallel",
+            empty_df,
+            LazyPlanDistributedArg(None, res_id),
```
DrTodd13, line 39: Just curious why this is "None" here.  When does distributed data created with scatter_data get collected?  Is res_id supposed to be put inside a lazy manager that is then responsible for it?  In short, how does res_id not leaking here?

DrTodd13 (approved):
LGTM.  Thanks!

bodo/pandas/base.py
IsaacWarren: It is, that's what the ticket I've been talking about where objects are getting leaked is coming from, I'm fixing this in a followup

IsaacWarren:
Failing tests are fixed in another PR, merging

PR #487, opened 2025-05-14
Only create plan for head if manager is lazy

scott-routledge2:
Checks the manager to determine whether it needs to create a plan for retrieving `head`. Creating a plan for `head` is not possible when the manager is not lazy. Similar checks are performed in other places.

## Testing strategy
`test_bodo_series_pandas_manager`, equivalent test for the dataframe case added.

## Checklist

### DIFF
--- a/bodo/tests/test_lazy/test_bodo_frame_pandas_manager.py
+++ b/bodo/tests/test_lazy/test_bodo_frame_pandas_manager.py
@@ -1,0 +1,22 @@
+import pandas as pd
+
+from bodo.pandas.frame import BodoDataFrame
+from bodo.tests.test_lazy.utils import pandas_managers  # noqa
+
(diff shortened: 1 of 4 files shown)


bodo/tests/test_lazy/test_bodo_frame_pandas_manager.py
```diff
@@ -0,0 +1,22 @@
+import pandas as pd
```
scott-routledge2, line 1: There wasn't an equivalent test for the DataFrame case but this test was also broken too. Also created a new file since `test_bodo_frame.py` are marked for spawn mode.

ehsantn (approved):

IsaacWarren (approved):
Thanks Scott

PR #494, opened 2025-05-16
Fix buffer pool error in filter example

ehsantn:
As title.

### DIFF
--- a/bodo/pandas/physical/read_pandas.h
+++ b/bodo/pandas/physical/read_pandas.h
@@ -110,9 +110,11 @@ class PhysicalReadPandas : public PhysicalSource {
             arrow::py::unwrap_table(pa_table).ValueOrDie();
 
         // Convert Arrow arrays to Bodo arrays
-        auto* bodo_pool = bodo::BufferPool::DefaultPtr();
+        // (passing nullptr for pool since not allocated through Bodo so can't
+        // support spilling etc,
+        // TODO: pass bodo's buffer pool to Arrow)
         std::shared_ptr<table_info> out_table =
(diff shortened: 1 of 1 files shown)


DrTodd13 (approved):
Thanks.

bodo/pandas/physical/read_pandas.h
```diff
@@ -110,9 +110,11 @@ class PhysicalReadPandas : public PhysicalSource {
             arrow::py::unwrap_table(pa_table).ValueOrDie();
 
         // Convert Arrow arrays to Bodo arrays
-        auto* bodo_pool = bodo::BufferPool::DefaultPtr();
+        // (passing nullptr for pool since not allocated through Bodo so can't
+        // support spilling etc,
+        // TODO: pass bodo's buffer pool to Arrow)
         std::shared_ptr<table_info> out_table =
-            arrow_table_to_bodo(table, bodo_pool);
+            arrow_table_to_bodo(table, nullptr);
```
scott-routledge2, line 117: I thought we already were setting the BodoBuffer pool as the default MemoryPool? I guess we can't guarantee that the allocation happens before we import bodo now?

scott-routledge2 (approved):

bodo/pandas/physical/read_pandas.h
ehsantn: Arrow may wrap Numpy arrays and not allocate new data?

scott-routledge2: Discussed offline, requires some further investigation.

PR #496, opened 2025-05-19
Handle DuckDB table and column reorderings in output

ehsantn:

Handles DuckDB's table/column reordering such as probe/build side changes in joins. Makes sure the output table matches output schema by reordering columns. Found several issues in the process including output column name issues.

## Testing strategy

Added unit test.

## User facing changes

Smaller probe table in join now works.

## Checklist


bodo/pandas/_plan.h
```diff
@@ -49,7 +60,10 @@ class BodoScanFunctionData : public duckdb::TableFunctionData {
  */
 class BodoParquetScanFunction : public BodoScanFunction {
    public:
-    BodoParquetScanFunction() : BodoScanFunction("bodo_read_parquet") {
+    BodoParquetScanFunction(const std::shared_ptr<arrow::Schema> arrow_schema)
+        : BodoScanFunction(
+              fmt::format("bodo_read_parquet({})",
+                          schemaColumnNamesToString(arrow_schema))) {
```
ehsantn, line 66: Just improves plan prints for source nodes to include column names to allow noticing DuckDB's plan reorderings better.

bodo/libs/_bodo_common.cpp
```diff
@@ -663,7 +663,7 @@ std::unique_ptr<Schema> Schema::Project(size_t first_n) const {
 }
 
 std::unique_ptr<Schema> Schema::Project(
-    const std::span<const int64_t> column_indices) const {
```
IsaacWarren, line 666: Why change this?

bodo/pandas/_plan.h
```diff
@@ -14,6 +15,16 @@
...
 
+/**
+ * @brief Return a string representation of the column names in the Arrow schema
+ * for printing purposes (e.g. plan prints).
+ *
+ * @param arrow_schema input Arrow schema
+ * @return std::string string representation of the column names
+ */
+std::string schemaColumnNamesToString(
+    const std::shared_ptr<arrow::Schema> arrow_schema);
```
IsaacWarren, line 26: I think there's a better place for this than this file

bodo/pandas/physical/join.h
```diff
@@ -76,6 +76,12 @@ class PhysicalJoin : public PhysicalSourceSink, public PhysicalSink {
         // Create the probe output schema, same as here for consistency:
         // https://github.com/bodo-ai/Bodo/blob/a2e8bb7ba455dcba7372e6e92bd8488ed2b2d5cc/bodo/libs/streaming/_join.cpp#L1138
         this->output_schema = std::make_shared<bodo::Schema>();
+        std::vector<std::string> col_names;
+        if (probe_table_schema->column_names.empty() ||
+            build_table_schema->column_names.empty()) {
+            throw std::runtime_error(
+                "Join input tables must have column names.");
+        }
```
IsaacWarren, line 84: Why is this true?

bodo/pandas/physical/read_pandas.h
```diff
@@ -77,6 +76,12 @@ class PhysicalReadPandas : public PhysicalSource {
         }
 
         num_rows = PyObject_Length(df);
+
+        this->output_schema = bodo::Schema::FromArrowSchema(arrow_schema);
+        if (!selected_columns.empty()) {
+            this->output_schema =
+                this->output_schema->Project(selected_columns);
+        }
```
IsaacWarren, line 84: This could be made into a class function so output_schema could be const

bodo/pandas/physical/result_collector.h
```diff
@@ -35,6 +48,32 @@ class PhysicalResultCollector : public PhysicalSink {
...
+        }
+        std::shared_ptr<table_info> out_table = std::make_shared<table_info>(
+            out_columns, in_table->nrows(), out_schema->column_names,
+            out_schema->metadata);
+        return out_table;
     }
+
+   private:
+    std::shared_ptr<bodo::Schema> in_schema;
+    std::shared_ptr<bodo::Schema> out_schema;
```
IsaacWarren, line 78:
```suggestion
    const std::shared_ptr<bodo::Schema> in_schema;
    const std::shared_ptr<bodo::Schema> out_schema;
```

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -786,3 +786,33 @@ def test_merge():
         sort_output=True,
         reset_index=True,
     )
+
```
IsaacWarren, line 789: Separate into its own test?

IsaacWarren (commented):
Thanks Ehsan, a couple comments

DrTodd13 (approved):
Agree with most of Isaac's review.  Nothing else to add.

bodo/libs/_bodo_common.cpp
ehsantn: To match the required use with vector<int>. The function was unused otherwise.

bodo/pandas/_plan.h
ehsantn: Moved the the util files and changed their name. We can discuss how to organize the rest of the code for a separate PR.

bodo/pandas/physical/read_pandas.h
ehsantn: Done.

bodo/tests/test_df_lib/test_end_to_end.py
ehsantn: Done.

bodo/pandas/physical/join.h
IsaacWarren: I'm not sure doing it on column names is safe
```
a = pd.DataFrame([range(2)] * 10, columns=['a','a'])
```
is valid pandas code

bodo/libs/_bodo_common.cpp
IsaacWarren: Do the cast at the call site? We use int64 for column indices everywhere iirc and span is preferable to vector in an API in my opinion

bodo/pandas/physical/join.h
ehsantn: Yes, but requiring column names to be unique is a reasonable restriction at this point I think.

bodo/libs/_bodo_common.cpp
ehsantn: There is just a single use right now and no need to generalize yet. We actually use vector of int for column indices in the new code looks like.

bodo/pandas/physical/join.h
IsaacWarren: Ok can you add it to the docs? Now is also probably a good time to add join to the docs generally

ehsantn: We need to add checks in from_pandas and update its docs, not join (will do separately). Join is not ready for docs yet (working on a couple of PRs).

IsaacWarren (approved):

PR #497, opened 2025-05-19
Execute plan before copying Series

scott-routledge2:
As titled. Copying a Series does not preserve the index because the manager copies the axes first, which doesn't trigger an execute/collect. See `pandas/core/managers.py:557` for more details.

Note that this doesn't change whether we collect in copy or not, it just changes when the collect happens to make sure the attributes needed by copy are correct up front.

## Testing strategy
* Added a new test based off of the Credit Card Fraud example.
* Ran the credit card fraud example locally to confirm it works end to end.
* PR CI

## User facing changes
Users will be able to copy Bodo DataFrames/Series into new dataframes e.g. `pd.DataFrame(bodo_df)` or `df.copy()`, however the result is always going to be collected. As a followup we can think about how to make the copying lazy.

## Checklist


IsaacWarren (approved):
Thanks Scott!

ehsantn (approved):
Looks good to me.

PR #508, opened 2025-05-21
[BSE-4421] Auto-generate Series.str methods with return type Series

chrisoh7:

Automates a subset of Series.str methods that do not take in arguments. Keeps track of return types in "rettypes" list and "series_str_methods" list that maps method names to the corresponding return type in "rettypes" via matching indices. 
Examines signatures using inspect.signature(), formats args and kwargs to set positional & keyword arguments for each method. 
When failed to validate arguments, raises TypeError, traceback formatted as below: 

```
Traceback (most recent call last):
  File "/Path/to/file.py", line 8, in <module>
    df.A.str.find("a", foo=1)
  File "/Path/to/series.py", line 462, in str_method
    sig_bind(name, *args, **kwargs)  # Argument validation
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Path/to/series.py", line 453, in sig_bind
    raise TypeError(f"StringMethods.{name}() {msg}")
TypeError: StringMethods.find() got an unexpected keyword argument 'foo'
```
Following methods are not supported. They are stored in "series_str_methods_unsupported" list and should be kept up-to-date with any additional implementation. 

```
Series.str.cat()
Series.str.decode()
Series.str.encode()
Series.str.extract()
Series.str.extractall()
Series.str.join()
Series.str.normalize()
Series.str.partition()
Series.str.rpartition()
Series.str.split()
Series.str.rsplit()
Series.str.get_dummies()
```

## Testing strategy

Pytest automation using Pytest parameters. Provides sanity check-level testing for all supported methods. 
* More thorough/exhaustive tests needed. 

## User facing changes

Large subset of Series.str methods supported. 

## Checklist


bodo/pandas/series.py
```diff
@@ -369,6 +369,18 @@ def upper(self):
             self._series._plan, new_metadata, "str.upper", (), {}
         )
 
+    @check_args_fallback(unsupported="none")
```
scott-routledge2, line 372: Why is `isalpha` done separately?

bodo/pandas/series.py
```diff
@@ -435,3 +447,49 @@ def _get_series_python_func_plan(series_proj, empty_data, func_name, args, kwarg
...
+    "isalnum": 1,
+    "isdigit": 1,
+    "isdecimal": 1,
+    "isspace": 1,
+    "islower": 1,
+    "isupper": 1,
+    "istitle": 1,
+}
+
+rettypes = [pd.ArrowDtype(pa.large_string()), bool]
```
scott-routledge2, line 490: Another way to organize this could be as a list of `tuple(str, return_type)`, or if you don't want to duplicate the return type a bunch of times, `tuple(list of methods, return_type)`

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -862,3 +879,30 @@ def test_dataframe_copy(index_val):
     pdf_from_bodo = pd.DataFrame(bdf)
 
     _test_equal(df1, pdf_from_bodo, sort_output=True)
+
+
+def gen_str_scalar_test(name):
```
scott-routledge2, line 884: Instead of generating the test function this way, could you look into using a pytest parameter? 
There should be examples in this file.

chrisoh7: Changed test file to automate pytest parameters. Thanks!

bodo/pandas/series.py
chrisoh7: Combined the two lists into `tuple(list of methods, return_type)`, thanks for the suggestion!

bodo/pandas/series.py
chrisoh7: Deleted. Thanks!

scott-routledge2 (approved):
Some minor comments and also let's make sure we create a followup issue to update the documentation. Otherwise LGTM, nice work!

bodo/pandas/series.py
```diff
@@ -445,3 +410,171 @@ def _get_series_python_func_plan(series_proj, empty_data, func_name, args, kwarg
...
+            "contains",
+            "match",
+            "fullmatch",
+        ],
+        pd.ArrowDtype(pa.bool_()),
+    ),
+    # idx = 2: Series(Float)
+    (
+        [
+            "len",
```
ehsantn, line 552: `len` should return integers not floats. Pandas by default may return float since Numpy integer arrays can't have nulls. But Pandas 3 fixes this issue by using Arrow instead of Numpy by default. We don't need to match the wrong behavior of current Pandas.

bodo/tests/test_df_lib/test_series_str.py
```diff
@@ -0,0 +1,215 @@
...
+    )
+    def test_func(args, kwargs):
+        df = pd.DataFrame(
+            {
+                "A": pd.array([1, 2, 3, 7], dtype="Int64"),
+                "B": ["Apple", "Banana", " Exc ited ", "Dog"],
+                "C": pd.array([4, 5, 6, -1], dtype="Int64"),
+            }
+        )
+        df.index = pd.RangeIndex(len(df))
```
ehsantn, line 31: This line is redundant.

ehsantn (approved):
Looks good to me. Just one comment about "len" output type needs addressed. Thanks @chrisoh7 !

bodo/pandas/series.py
chrisoh7: Moved it to Series(int) list, and deleted Series(float) entry since `len` was the only member. Thanks!

bodo/tests/test_df_lib/test_series_str.py
chrisoh7: Deleted. Thanks :)!

PR #511, opened 2025-05-22
[BSE-4831] Unify parquet schemas

scott-routledge2:

Unify parquet schemas when `read_parquet` is called to make sure the schemas are correct. This is also a bug in our compiler so ideally we can fix in both places.

Experimenting with a couple different options for a solution:
1. Call get_pq_dataset with count_rows=True. The advantage of this approach is that we can call in parallel, and possibly store the result so we don't have to also call at runtime? Disadvantage is we could end up filtering out a lot of these files at runtime so would probably have to recompute anyways.

2. For the most common case where the schema is null in the first file for some of the columns, we could just iterate over fragments in the dataset until the null columns are resolved. In the worst case, we would have to iterate over all the files which could be really slow especially if we can't process the files in parallel.

This PR goes with approach 2.

On the NYC taxi example (averaged 3 runs):
Getting the dataset the buggy way (main): ~0.39s
Getting the dataset until all null columns are resolved (this branch): ~1.14s
Getting the dataset entire dataset using `get_row_counts=True`:  ~7.700s (NOTE run on the spawner, could potentially be faster if we ran on workers?)

``` py
import time
import bodo.pandas as bd

times = []
for _ in range(3):
    start = time.time()
    df = bd.read_parquet("s3://bodo-example-data/nyc-taxi/fhvhv_tripdata/")
    end = time.time()
    times.append(end-start)
    print(end-start)

print(f"avg creating parquet read lazy plan took {sum(times)/len(times):2f}s")
```

## Testing strategy
PR CI, perf testing on the NYC taxi example, which has a null column (airport_fee) which is Null in the first 6 files and then double + null in the remaining 56 files.

## User facing changes
Slight decrease in initial `read_parquet` time if a null column is present, but more robust to the case where there is a null in the first file but non-null values after.

## Checklist

### DIFF
--- a/bodo/io/parquet_pio.py
+++ b/bodo/io/parquet_pio.py
@@ -661,7 +661,8 @@ def unify_fragment_schema(dataset: ParquetDataset, piece: ParquetPiece, frag):
     """Unifies schema of *dataset* with incoming piece/fragment.
 
     Args:
-        dataset (ParquetDataset): The Parquet dataset with schema to unify.
+        dataset (ParquetDataset): The Parquet dataset to update with the unified
+            schema.
         piece (ParquetPiece): Piece corresponding to the fragment being unified.
         frag (pa.Dataset.Fragment): Fragment of the dataset to unify.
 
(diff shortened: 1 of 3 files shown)


bodo/io/parquet_pio.py
```diff
@@ -657,6 +657,39 @@ def unify_schemas_across_ranks(dataset: ParquetDataset, total_rows_chunk: int):
     ev.finalize()
 
 
+def unify_fragment_schema(dataset: ParquetDataset, piece: ParquetPiece, frag):
+    """Unifies schema of *dataset* with incoming piece/fragment.
+
+    Args:
+        dataset (ParquetDataset): The Parquet dataset with schema to unify.
```
DrTodd13, line 664: Maybe some indication here that it is this argument that is updated with the unified schema.

DrTodd13 (approved):
LGTM.  Thanks.

IsaacWarren (approved):
Thanks Scott!

PR #512, opened 2025-05-23
Support basic groupby

ehsantn:

Adds initial support for groupby for the single column sum case. Also updates count(*) to use some similar code paths. Will open a bunch of issues for follow ups.

## Testing strategy

Unit test.

## User facing changes

groupby for the single column sum case is supported now.

## Checklist

### DIFF
--- a/bodo/pandas/utils.py
+++ b/bodo/pandas/utils.py
@@ -895,7 +895,7 @@ def count_plan(self):
 
     # Can't be known statically so create count plan on top of
     # existing plan.
-    count_star_schema = pd.Series(dtype="uint64", name="count_star()")
+    count_star_schema = pd.Series(dtype="uint64", name="count_star")
     aggregate_plan = LazyPlan(
         "LogicalAggregate",
         count_star_schema,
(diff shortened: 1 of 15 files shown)


bodo/pandas/utils.py
```diff
@@ -900,9 +900,18 @@ def count_plan(self):
...
-        0,
-        0,
-        [LazyPlan("FunctionExpression", count_star_schema, "count_star()")],
+        [],
+        [
+            LazyPlan(
+                "AggregateExpression",
+                count_star_schema,
+                self._plan,
+                "count_star",
```
DrTodd13, line 909: Consider changing the column name to remove () since I don't think you are using that with your new aggregate impl.

DrTodd13: That is above on line 898.

DrTodd13 (approved):
LGTM.  Thanks.

bodo/pandas/utils.py
ehsantn: Done.

PR #530, opened 2025-06-02
[BSE-4854] Expand Series method support

chrisoh7:

Expands Series coverage to include support for methods listed below. Further generalizes auto-generation of Series methods into single function with various parameters. Generated method inspects corresponding signatures to validate arguments, and exception cases with *args & **kwargs are kept in a dictionary and checked separately. 

Adds logic to infer output datatypes of certain methods whose output dtypes are dependent on input dtypes.

Couple of restrictions regarding timedelta (duration) and timezone-related methods should be noted in the docs and addressed in the future. 

```
Series.isin
Series.notnull
Series.isnull
# Next four methods examine input type and infer output type to match Pandas behavior
Series.clip 
Series.replace
Series.abs
Series.round

Series.dt.normalize
Series.dt.floor
Series.dt.ceil
Series.dt.month_name
Series.dt.day_name
Series.dt.quarter

Series.str.partition
Series.str.rpartition
```
Updates docs to include changes. 

`Series.ffill`, `Series.bfill`, `Series.dt.end_time`, `Series.dt.total_seconds` could not be implemented and are left as TODO. 
* `strftime` formats second values (%S) to include nanoseconds by default

## Testing strategy

PR CI, adds automation for datetime methods with arguments, sanity check level. More rigorous check needed. 

## User facing changes

Support for more Series methods, especially direct methods. 

## Checklist


bodo/pandas/series.py
```diff
@@ -446,6 +446,11 @@ def _get_result_id(self) -> str | None:
             return self._mgr._md_result_id
         return None
 
+    @property
+    def empty(self):
```
scott-routledge2, line 451: Instead of executing the whole plan, it might be more efficient to check len(self) == 0 ? That should in theory just add a count on top of the plan and execute it.

bodo/pandas/series.py
```diff
@@ -585,85 +574,106 @@ def _get_series_python_func_plan(series_proj, empty_data, func_name, args, kwarg
     )
 
 
-def sig_bind(name, *args, **kwargs):
+# TODO: create sample series dictionary
```
scott-routledge2: Can you clarify what is meant by this TODO?

bodo/pandas/series.py
```diff
@@ -585,85 +574,106 @@ def _get_series_python_func_plan(series_proj, empty_data, func_name, args, kwarg
...
-                inspect.Parameter(
-                    "drop_whitespace", inspect.Parameter.KEYWORD_ONLY, default=True
-                ),
-                inspect.Parameter(
-                    "break_long_words", inspect.Parameter.KEYWORD_ONLY, default=True
-                ),
-                inspect.Parameter(
-                    "break_on_hyphens", inspect.Parameter.KEYWORD_ONLY, default=True
-                ),
+        # print("Binding args and kwargs for Series method:", name)
```
```suggestion
```

bodo/pandas/series.py
```diff
@@ -585,85 +574,106 @@ def _get_series_python_func_plan(series_proj, empty_data, func_name, args, kwarg
...
-    def str_method(self, *args, **kwargs):
-        """Generalized template for Series.str methods and argument validation using signature"""
-        sig_bind(name, *args, **kwargs)  # Argument validation
-
-        index = self._series.head(0).index
+    raise TypeError(f"{accessor_names.get(accessor_type, '')}{name}() {msg}")
+
+
+# Maps Series methods to signatures
+sig_map = {
```
scott-routledge2: Might be good to add a type anotation i.e. dict[str, list[tuple, ... There are some good examples of useful type annotations in `workers.py`

Also why does the default value need to be a Tuple, is it just to indicate when something doesn't have default?

bodo/pandas/series.py
```diff
@@ -794,16 +808,96 @@ def dt_accessor(self):
...
+            "day_name",
+            # TODO: fix precision of seconds (%S by default prints up to nanoseconds)
+            "strftime",
+        ],
+        pd.ArrowDtype(pa.large_string()),
+    ),
 ]
 
 # Generates Series.str methods
 for str_pair in series_str_methods:
```
scott-routledge2: Let's refactor these for loops into functions e.g. `def _install_series_str` and call the functions at the bottom. We do this in other places in the compiler if you search `_install_` and filter for python. Just makes the code look a little cleaner IMO.

bodo/pandas/series.py
```diff
@@ -794,16 +808,96 @@ def dt_accessor(self):
...
+            "isin",
+            "notnull",
+            "isnull",
+        ],
+        pd.ArrowDtype(pa.bool_()),
+    ),
+    (  # idx = 1: Series(Float)
+        [
+            "ffill",
+            "clip",
```
scott-routledge2: I am pretty sure abs has the same issue as clip (return type depends on input for example int vs float). Maybe have a category with return type "None" that infers type based on inputs? Same with replace it should match the type of the input.

Secondly, bfill/ffill are not going to be possible since they potentially require communication across batches/workers. For example if your batch was [None, None, None] And the next batch was [1, 1, 1], The correct thing to do would be to fill the prev batch with 1's (but that wouldn't work with the way it's currently set up). Let's remove for now.

bodo/pandas/series.py
```diff
@@ -794,16 +808,96 @@ def dt_accessor(self):
...
+        ],
+        pd.ArrowDtype(pa.float64()),
+    ),
+    # idx = 2: Series(String)
+    (
+        [
+            # TODO: Add locale support
+            "month_name",
+            "day_name",
+            # TODO: fix precision of seconds (%S by default prints up to nanoseconds)
```
scott-routledge2: Let's open an issue

bodo/pandas/series.py
```diff
@@ -794,16 +808,96 @@ def dt_accessor(self):
...
+        [
+            # TODO: implement total_seconds (support timedelta)
+            # "total_seconds",
+        ],
+        pd.ArrowDtype(pa.float64()),
+    ),
+    # idx = 2: Series(String)
+    (
+        [
+            # TODO: Add locale support
```
scott-routledge2: Can we fallback to Pandas in this case?

bodo/pandas/series.py
```diff
@@ -794,16 +808,96 @@ def dt_accessor(self):
...
+    # idx = 0: Series(Boolean)
+    (
+        [
+            "normalize",
+            "floor",
+            "ceil",
+        ],
+        pd.ArrowDtype(pa.timestamp("ns")),
+    ),
+    # idx = 1: Series(String)
```
scott-routledge2: Update comments

bodo/pandas/series.py
```diff
@@ -794,16 +808,96 @@ def dt_accessor(self):
...
         pd.ArrowDtype(pa.bool_()),
     ),
+    # idx = 4: Series(Timestamp)
+    (
+        [
+            # "end_time",
+        ],
+        pd.ArrowDtype(pa.timestamp("ns")),
+    ),
+    # idx = 5: Series(String)
```
scott-routledge2: Remove or add TODO so that it's clear we need to implement end_time?

bodo/pandas/series.py
```diff
@@ -794,16 +808,96 @@ def dt_accessor(self):
...
+        setattr(BodoDatetimeProperties, name, property(accessor))
+
+# Generates Series.dt methods
+for dt_method_pair in dt_methods:
+    for name in dt_method_pair[0]:
+        method = gen_method(name, dt_method_pair[1], accessor_type="dt.")
+        setattr(BodoDatetimeProperties, name, method)
+
+# Generates direct Series.<method>
+for dir_method_pair in dir_methods:
```
scott-routledge2: Add tests for these?

bodo/pandas/series.py
chrisoh7: Yup, in other words when something is a required arg. At first i wanted to tell them apart by looking at whether it was positional or not, but there were a few optional positional args, so now if its a tuple i know it's required.

bodo/pandas/series.py
chrisoh7: Sure, done. Thanks.

bodo/pandas/series.py
chrisoh7: Did some testing, found out that locale actually works, but the locale string arg should be super strictly formatted. 🎉

bodo/pandas/series.py
chrisoh7: Done.

bodo/pandas/series.py
chrisoh7: Done.

bodo/pandas/series.py
chrisoh7: Done.

bodo/tests/test_df_lib/test_dir.py
```diff
@@ -0,0 +1,98 @@
+import pandas as pd
```
scott-routledge2: I would rename this file to `test_direct_series_methods.py` to be more clear.

bodo/tests/test_df_lib/test_dir.py
```diff
@@ -0,0 +1,98 @@
...
+        arg_sets,
+        ids=[
+            f"{name}-args{args}-kwargs{sorted(kwargs.items())}"
+            for args, kwargs in arg_sets
+        ],
+    )
+    def test_func(args, kwargs):
+        df = pd.DataFrame(
+            {
+                "A": pd.array([1.445, 2.12, None, -4.133], dtype="Float64"),
```
scott-routledge2: Can you add some `None` values to this dataframe?

bodo/pandas/series.py
```diff
@@ -467,21 +472,16 @@ def map(self, arg, na_action=None):
...
-        index = self.head(0).index
-        new_metadata = pd.Series(
-            dtype=pd.ArrowDtype(pa.bool_()),
-            name=self.name,
-            index=index,
-        )
-        return _get_series_python_func_plan(
-            self._plan, new_metadata, "isin", (values,), {}
-        )
+def _find_(s, col):
```
scott-routledge2: Let's refactor to something more descriptive like `_str_partition_helper`

bodo/pandas/series.py
```diff
@@ -585,85 +585,177 @@ def _get_series_python_func_plan(series_proj, empty_data, func_name, args, kwarg
...
+        dtype = pd.ArrowDtype(pa.list_(pa.large_string()))
+
+        index = series.head(0).index
+        new_metadata = pd.Series(
+            dtype=dtype,
+            name=series.name,
+            index=index,
+        )
+
+        tuple_plan = _get_series_python_func_plan(
```
scott-routledge2: Name of this variable might be a bit confusing since it is a BodoSeries, not a plan

bodo/pandas/series.py
```diff
@@ -796,14 +890,110 @@ def dt_accessor(self):
...
+        [
+            "isin",
+            "notnull",
+            "isnull",
+        ],
+        pd.ArrowDtype(pa.bool_()),
+    ),
+    (  # idx = 1: Series(Float)
+        [
+            "replace",
```
scott-routledge2: I think these methods also depend on the input type, for example:
```
>>> x = pd.Series([1,2,4,5], dtype=pd.ArrowDtype(pa.int8()))
>>> x.replace(1, 2)
0    2
1    2
2    4
3    5
dtype: int8[pyarrow]
>>> x.round()
0    1
1    2
2    4
3    5
dtype: int8[pyarrow]
```

bodo/tests/test_df_lib/test_datetime.py
```diff
@@ -31,7 +32,126 @@ def test_func():
...
+    @pytest.mark.parametrize(
+        "args, kwargs",
+        arg_sets,
+        ids=[
+            f"{name}-args{args}-kwargs{sorted(kwargs.items())}"
+            for args, kwargs in arg_sets
+        ],
+    )
+    def test_func(args, kwargs):
+        date_m = pd.Series(pd.date_range("20130101 09:10:12", periods=10, freq="MS"))
```
scott-routledge2: Add None-values to input DataFrames

bodo/tests/test_df_lib/test_series_str.py
```diff
@@ -24,7 +24,7 @@ def test_func(args, kwargs):
         df = pd.DataFrame(
             {
                 "A": pd.array([1, 2, 3, 7], dtype="Int64"),
-                "B": ["Apple", "Banana", " Exc ited ", "Dog"],
+                "B": ["App-le", "B-anan-a", " E-xc i-ted ", "Do-g"],
```
scott-routledge2: Let's make sure we are also testing with some of the values being None

scott-routledge2 (approved):
See my comments, overall LGTM thanks @chrisoh7 !

bodo/tests/test_df_lib/test_series_str_methods.py
```diff
@@ -51,7 +58,20 @@ def test_func(args, kwargs):
...
+            _test_equal(out_bodo, out_pd, check_pandas_types=False)
+        except AssertionError as e:
+            """
+            Exception case handler: currently partition and rpartition returns same outputs, but 
+            with different index types causing the equality test to fail. 
+            In this case, we print out the outputs Bodo vs. Pandas for manual inspection. 
+            """
+            if name in ["partition", "rpartition"]:  # Exception list
+                print(
+                    f"Outputs may or may not differ, manually compare: \nPandas:\n{out_pd}\nBodo:\n{out_bodo}"
```
ehsantn, line 71: What about adding `reset_index=True` to `_test_equal` for these cases?

bodo/tests/test_df_lib/test_series_str_methods.py
```diff
@@ -23,15 +23,22 @@ def gen_str_param_test(name, arg_sets):
...
+                    "Do-g",
+                ],
             }
         )
         bdf = bd.from_pandas(df)
 
-        pd_func = getattr(df.B.str, name)
-        bodo_func = getattr(bdf.B.str, name)
+        pd_func = getattr(df.A.str, name)
+        bodo_func = getattr(bdf.A.str, name)
```
ehsantn, line 41: Just curious, why was this change necessary?

bodo/tests/test_df_lib/test_direct_series_methods.py
```diff
@@ -0,0 +1,99 @@
...
+            {
+                "A": pd.array([1.445, 2.12, None, -4.133], dtype="Float64"),
+                "B": pd.array([1, 2, 3, 4], dtype="Int64"),
+            }
+        )
+        bdf = bd.from_pandas(df)
+
+        keys = ["A", "B"]
+
+        for key in keys:
```
ehsantn, line 32: Looks like there is some duplicated code in test generation that could be refactored in a follow up PR.

bodo/pandas/series.py
```diff
@@ -585,85 +584,177 @@ def _get_series_python_func_plan(series_proj, empty_data, func_name, args, kwarg
...
-            func = getattr(str_accessor, name)
+            sample_series = (
+                pd.Series(pd.to_datetime(["2023-01-01"]))
+                if accessor_type == "dt."
+                else pd.Series(["a"])
+            )
+            if accessor_type:
+                sample_series = (
+                    sample_series.str if accessor_type == "str." else sample_series.dt
+                )
```
ehsantn: This sample series generation is a bit dangerous since it assumes dt/str or Series with string data. More guardrails would be useful.

bodo/pandas/utils.py
```diff
@@ -627,7 +628,7 @@ def run_func_on_table(cpp_table, arrow_schema, result_type, in_args):
     else:
         # TODO: test this path
         func = _get_function_from_path(func_path_str)
-        out = func(input, *args, **kwargs)
+        out = func(input, *args)
```
ehsantn: Why is kwargs removed here?

ehsantn (approved):
Looks good to me. Thanks @chrisoh7 . See comments.

bodo/tests/test_df_lib/test_series_str_methods.py
chrisoh7: Before the change, the test initialized a dataframe with columns A, B, C, among which A and C were not utilized. I deleted columns A and C, and in that process decided to rename B->A. Pretty trivial but thought it'd be neater.

bodo/tests/test_df_lib/test_direct_series_methods.py
chrisoh7: Sure, had that thought too. I'll create an issue and assign it to myself.

bodo/pandas/utils.py
chrisoh7: Removed during debugging, just added it back. Thanks!

bodo/tests/test_df_lib/test_series_str_methods.py
ehsantn: Ok, thanks.

bodo/tests/test_df_lib/test_series_str_methods.py
chrisoh7: I've tried it, but `_test_equal` still treats them as different output values, so I decided to live with that for now. Let me know if you think I should further look into it.

ehsantn: Thanks, ok for now.

PR #534, opened 2025-06-03
Update DataFrame set column docs

ehsantn:
Updates docs setting dataframe columns to include setting constant values. Setting another dataframe column is also implied in the new language (not worth emphasizing).


chrisoh7 (approved):
Looks good to me, thanks @ehsantn !

PR #536, opened 2025-06-03
Improve the performance of Python calls in the pipeline

ehsantn:

Improves performance substantially (~10x in my testing) by using Arrow in Python/C++ data conversions instead of Numba JIT functions.

## Testing strategy

Existing tests.

## User facing changes

Improved performance.

## Checklist


ehsantn:
Looks like CI revealed existing issues in bodo<>arrow conversion codes. Will ask again for review when fixed since it may take a while.

bodo/pandas/utils.py
```diff
@@ -59,46 +59,13 @@ def get_lazy_single_manager_class() -> type[
...
-
-@numba.njit
-def cpp_table_to_py(in_table, out_cols_arr, out_table_type):
-    """Convert a C++ table pointer to a Python table.
-    Args:
-        in_table (int64): C++ table pointer
-        out_cols_arr (array(int64)): Array of column indices to be extracted
-        out_table_type (types.Type): Type of the output table
-    """
-    cpp_table = cast_int64_to_table_ptr(in_table)
```
scott-routledge2, line 80: I think we can delete the definition of `cast_int64_to_table_ptr` since it's no longer used

bodo/pandas/utils.py
```diff
@@ -824,6 +776,79 @@ def arrow_to_empty_df(arrow_schema):
...
+    if isinstance(arr, pa.ChunkedArray):
+        arr = arr.combine_chunks()
+
+    new_arrs = [
+        _fix_struct_arr_names(arr.field(i), pa_type.field(i).type)
+        for i in range(arr.type.num_fields)
+    ]
+    names = [pa_type.field(i).name for i in range(pa_type.num_fields)]
+    new_arr = pa.StructArray.from_arrays(new_arrs, names)
+    # Arrow's from_arrays ignores nulls (bug as of Arrow 19) so we add them back
```
scott-routledge2, line 786: Do we know if this is fixed in Arrow 20?

bodo/pandas_compat.py
```diff
@@ -140,6 +140,40 @@ def _concat_same_type(cls, to_concat):
 pd.core.arrays.arrow.array.ArrowExtensionArray._concat_same_type = _concat_same_type
 
 
+pd_str_find = pd.core.arrays.arrow.array.ArrowExtensionArray._str_find
+
+
+def _str_find(self, sub: str, start: int = 0, end: int | None = None):
```
scott-routledge2, line 146: Maybe it would be better to add this fallback logic in the definition of our `Series.str.find`? Can be a followup.

scott-routledge2 (approved):
LGTM, thanks @ehsantn !

bodo/pandas/utils.py
ehsantn: It's already deleted above.

scott-routledge2: I meant `cast_table_ptr_to_int64`

bodo/pandas/utils.py
ehsantn: Actually, we may be able to use the `mask` parameter and get it working now. Will leave it as follow up though.

bodo/pandas_compat.py
ehsantn: I think it would make our code more complicated. Also, this is a Pandas issue that we are fixing (helps outside of Bodo context too) so makes sense to be here.

bodo/pandas/utils.py
ehsantn: Thanks. Done.

bodo/pandas/utils.py
```diff
@@ -824,6 +762,79 @@ def arrow_to_empty_df(arrow_schema):
...
+
+    Returns:
+        pd.DataFrame: The converted pandas DataFrame.
+    """
+    if arrow_schema is None:
+        arrow_schema = arrow_table.schema
+
+    df = pd.DataFrame(
+        {
+            i: _arrow_to_pd_array(arrow_table.columns[i], field.type)
```
DrTodd13, line 829: The comment on 833 isn't clear to me.  How is that line handling duplicate names and how is that not the equivalent of doing "field.name:" here instead of "i:"?

bodo/pandas/utils.py
```diff
@@ -951,3 +962,29 @@ def count_plan(self):
 
     data = execute_plan(projection_plan)
     return data[0]
+
+
+def ensure_datetime64ns(df):
+    """Convert datetime columns in a DataFrame to 'datetime64[ns]' dtype.
+    Avoids datetime64[us] that is commonly used in Pandas but not supported in Bodo.
```
DrTodd13, line 970: The docs and the code don't match.  The code avoids everything except ns right?  I recall some other code above where we support second, milli, micro, nano so I think we need to be consistent and define which precisions we support and if it is only us we avoid then the code should reflect that.

DrTodd13 (approved):
thanks

bodo/pandas/utils.py
ehsantn: `"field.name:"` in a dictionary would replace duplicated values. Updated the comment.

bodo/pandas/utils.py
ehsantn: We convert non-ns precisions to ns in some places but I agree we should clarify and be more consistent. Generally, anything that goes through Arrow is fine but Pandas support with Numpy dtypes may have issues.

PR #538, opened 2025-06-03
Reduce codecov target

IsaacWarren:
As title, until we get coverage working for spawn mode/df lib tests, no reason to have this always be red


scott-routledge2 (approved):
Thanks @IsaacWarren

PR #544, opened 2025-06-06
[BSE-4882] Refactor Series method tests

chrisoh7:

Refactors test codes for Series.str, Series.dt, and Series methods and addresses duplicated code issue. 
Updates docs to note the different behaviors of Bodo and Pandas in the output values of datetime predicate methods on NaT series. 

## Testing strategy

PR CI, adds more test cases such as null series. 

## User facing changes

Documentation properly notes behavior of Bodo that differs from Pandas. 

## Checklist


bodo/tests/test_df_lib/test_series_generator.py
```diff
@@ -0,0 +1,125 @@
...
+                _test_equal(out_bodo, out_pd, check_pandas_types=False)
+            except AssertionError as e:
+                """
+                Exception case handler: currently month_name and day_name with locale arg
+                returns outputs that trivially differ from Pandas. Also, partition and rpartition returns same outputs, but 
+                with different index types causing the equality test to fail. 
+                In this case, we print out the outputs Bodo vs. Pandas for manual inspection. 
+                """
+                if name in [
+                    "partition",
```
scott-routledge2: For partition/rpartition I think the issue is that we are not propagating the index arrays. We should look into fixing that. (you can look at what we do in `_get_series_python_func_plan` for example of selecting index arrays).

scott-routledge2: Also instead of this manual checking, I think it would be better if we just constructed the "expected" result for each case where the result does not match pandas. For example, you could have an "expected_result" field in your input:
``` py
"partition": [
(args, kwargs, expected_results), ...
]
```

scott-routledge2 (approved):
LGTM, one comment looks like I meant to make on a previous PR.
Thanks @chrisoh7

bodo/tests/test_df_lib/test_series_generator.py
chrisoh7: Done!

ehsantn (approved):
Looks good to me. Thanks @chrisoh7 .

PR #551, opened 2025-06-09
Map partitions in lazy plan

scott-routledge2:
## Checklist

### DIFF
--- a/bodo/pandas/frame.py
+++ b/bodo/pandas/frame.py
@@ -554,9 +554,13 @@ class BodoDataFrame(pd.DataFrame, BodoLazyWrapper):
                 empty_series = get_scalar_udf_result_type(
                     self, "map_partitions", func, *args, **kwargs
                 )
-            except BodoLibNotImplementedException:
+            except BodoLibNotImplementedException as e:
                 required_fallback = True
-                msg = ("",)
+                msg = (
+                    f"map_paritions(): encountered exception: {e}, while trying to "
+                    "build lazy plan. Executing plan and running map_paritions on "
+                    "workers (may be slow or run out of memory)."
+                )
                 warnings.warn(BodoLibFallbackWarning(msg))
 
                 df_arg = self.execute_plan()
--- a/bodo/pandas/frame.py
+++ b/bodo/pandas/frame.py
@@ -530,6 +530,19 @@ class BodoDataFrame(pd.DataFrame, BodoLazyWrapper):
         NOTE: this pickles the function and sends it to the workers, so globals are
         pickled. The use of lazy data structures as globals causes issues.
         """
+        if self._exec_state == ExecState.PLAN:
+            try:
+                empty_series = get_scalar_udf_result_type(
+                    self, "map_partitions", func, *args, **kwargs
+                )
+            except BodoLibNotImplementedException:
+                collected_self = self.execute_plan()
+                return collected_self.map_partitions(func, *args, **kwargs)
+
+            return _get_df_python_func_plan(
+                self._plan, empty_series, func, args, kwargs
+            )
+
         return bodo.spawn.spawner.submit_func_to_workers(
             func, [], self, *args, **kwargs
         )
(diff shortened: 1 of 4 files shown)


bodo/pandas/utils.py
```diff
@@ -548,13 +548,14 @@ def run_func_on_table(cpp_table, result_type, in_args):
     return the result as a C++ table and column names.
     """
     input = cpp_table_to_df(cpp_table)
-    func_path_str, is_series, is_attr, args, kwargs = in_args
```
scott-routledge2, line 551: Since map_partitions isn't a pandas method, it feels easier to just call *func* directly.

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -1002,3 +1002,27 @@ def test_series_compound_expression(datapath):
...
+            "E": [1.1, 2.2, 13.3] * 2,
+            "A": pd.array([2, 2, 3] * 2, "Int64"),
+        },
+        index=[0, 41, 2] * 2,
+    )
+
+    bodo_df = bd.from_pandas(df)
+
+    def f(df, a, b=1):
+        return df.A + df.E + a + b
```
IsaacWarren, line 1021: Can we test that f is getting multiple rows?

bodo/tests/test_df_lib/test_end_to_end.py
IsaacWarren, line 1: Can we add a test for the fallback logic too?

docs/docs/api_docs/dataframe_lib/dataframe.md
```diff
@@ -130,6 +130,62 @@ Output:
...
+## bodo.pandas.BodoDataFrame.map_partitions
+``` py
+BodoDataFrame.map_partitions(func, *args, **kwargs) -> BodoSeries | BodoDataFrame
+```
+
+Apply a function to groups of rows in a DataFrame and return a DataFrame or Series of the same size.
+
+<p class="api-header">Parameters</p>
+
+: __func : *Callable*:__ A function that takes in a DataFrame and returns a DataFrame or Series (with the same number of rows). Currently, functions that return a DataFrame will trigger execution, but the function will be evaluated in parallel across all workers.
```
IsaacWarren: If returning a dataframe triggers execution what's the point of the lazy plan?

docs/docs/api_docs/dataframe_lib/dataframe.md
```diff
@@ -130,6 +130,62 @@ Output:
 
 ---
 
+## bodo.pandas.BodoDataFrame.map_partitions
+``` py
+BodoDataFrame.map_partitions(func, *args, **kwargs) -> BodoSeries | BodoDataFrame
+```
+
+Apply a function to groups of rows in a DataFrame and return a DataFrame or Series of the same size.
```
IsaacWarren, line 138: Maybe say something about how the whole local chunk that each worker has will be passed to func?

IsaacWarren (commented):
Thanks Scott

docs/docs/api_docs/dataframe_lib/dataframe.md
scott-routledge2: Sure, I will clarify thanks. Also if it's in the Lazy path, then each worker will process `batch_size` at once.

docs/docs/api_docs/dataframe_lib/dataframe.md
scott-routledge2: This was the scope we decided on for this ticket originally, couldn't you say the same thing about apply? Some benefits of just having it in the Series case for now:
* If you are setting a column to map_partitions and then you later decided you don't need to use that column, the map_partitions can be optimized out
* Streaming execution
* limit pushdown?

IsaacWarren: Good point, didn't realize this was the case for apply too

docs/docs/api_docs/dataframe_lib/dataframe.md
IsaacWarren: Good point

bodo/pandas/frame.py
```diff
@@ -527,11 +529,49 @@ def to_json_wrapper(
...
+        If self is a lazy plan, then the result will also be a lazy plan
+        (assuming result is Series and the dtype can be infered). Otherwise, the lazy
+        plan will be evaluated.
         NOTE: this pickles the function and sends it to the workers, so globals are
         pickled. The use of lazy data structures as globals causes issues.
+
+        Args:
+            func (Callable): A callable which takes in a DataFrame as it's first
+                argument and returns a DataFrame or Series that has the same length
+                it's input.
```
ehsantn:
```suggestion
            func (Callable): A callable which takes in a DataFrame as its first
                argument and returns a DataFrame or Series that has the same length
                its input.
```

ehsantn (approved):
Looks good to me. Thanks @scott-routledge2!

bodo/tests/test_df_lib/test_end_to_end.py
scott-routledge2: Tested locally on 2 workers, it took 1 row (to determine output type) and then took 3 rows on each worker

PR #552, opened 2025-06-09
Series filter.

DrTodd13:

Make filter work on series.

## Testing strategy

run_ci

## User facing changes

None

## Checklist


bodo/pandas/series.py
```diff
@@ -328,6 +329,40 @@ def __floordiv__(self, other):
...
+    @check_args_fallback("all")
+    def __getitem__(self, key):
+        """Called when df[key] is used."""
+
+        from bodo.pandas.base import _empty_like
+
+        # Only selecting columns or filtering with BodoSeries is supported
+        if not isinstance(key, BodoSeries):
+            raise BodoLibNotImplementedException(
+                "only string and BodoSeries keys are supported"
```
ehsantn:
```suggestion
                "only BodoSeries keys are supported"
```

bodo/pandas/series.py
```diff
@@ -328,6 +329,40 @@ def __floordiv__(self, other):
...
+            )
+
+        zero_size_self = _empty_like(self)
+
+        key_plan = (
+            # TODO: error checking for key to be a projection on the same dataframe
+            # with a binary operator
+            get_proj_expr_single(key._plan)
+            if key._plan is not None
+            else plan_optimizer.LogicalGetSeriesRead(key._mgr._md_result_id)
```
ehsantn, line 353: Do we need to handle the collected case for BodoSeries here?

bodo/pandas/series.py
```diff
@@ -328,6 +329,40 @@ def __floordiv__(self, other):
...
+        )
+        zero_size_key = _empty_like(key)
+        zero_size_index = zero_size_key.index
+        empty_data = zero_size_self.__getitem__(zero_size_key)
+        empty_data_index = empty_data.index
+        if isinstance(zero_size_index, pd.RangeIndex) and not isinstance(
+            empty_data_index, pd.RangeIndex
+        ):
+            # Drop the explicit index.
```
ehsantn:
```suggestion
            # Drop the explicit integer Index generated from filtering RangeIndex (TODO: support RangeIndex properly).
```

bodo/pandas/utils.py
```diff
@@ -385,8 +385,9 @@ def recursive_check(x):
                 return x
 
         # Convert any LazyPlan in the args or kwargs.
-        args = [recursive_check(x) for x in self.args]
         kwargs = {k: recursive_check(v) for k, v in self.kwargs.items()}
+        args = [recursive_check(x) for x in reversed(self.args)]
+        args.reverse()
```
ehsantn, line 400: This needs comments with specific examples. Let's discuss the issues today.

ehsantn (commented):
Thanks Todd. Let's discuss the plan issue today.

bodo/pandas/series.py
DrTodd13: Isn't accessing key._plan going to distribute if it is collected?

bodo/pandas/utils.py
DrTodd13: Done.

bodo/pandas/series.py
scott-routledge2: I think we should add a test for this case to verify this

bodo/pandas/utils.py
```diff
@@ -372,7 +372,8 @@ def generate_duckdb(self, cache=None):
         if cache is None:
             cache = {}
         # If previously converted then use the last result.
-        if id(self) in cache:
+        # Don't cache expression nodes.
+        if "Expression" not in self.plan_class and id(self) in cache:
```
scott-routledge2, line 378: Maybe add a TODO to disable cache in followup?

DrTodd13: done.

bodo/pandas/series.py
DrTodd13: We can't isolate the key and treat it differently than the source because at this point we require both of them to be the same series.  So, I added a test to put the source and key into plan, distributed, or collected state and indeed there were issues because in two of those cases the plan will not be a projection with an expr.  I modified get_proj_expr_single to return a LazyPlan ColRefExpression in those cases and needed an extra function to dealing with indices.  Please re-review. @ehsantn @scott-routledge2

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -1043,3 +1043,132 @@ def g(df, a, b=1):
...
+@pytest.mark.parametrize(
+    "op", [operator.eq, operator.ne, operator.gt, operator.lt, operator.ge, operator.le]
+)
+def test_series_filter_distributed(datapath, file_path, op):
+    """Very simple test for series filter for sanity checking."""
+    bodo_df1 = bd.read_parquet(datapath(file_path))
+    py_df1 = pd.read_parquet(datapath(file_path))
+
+    @bodo.jit(spawn=True)
+    def f(df):
```
scott-routledge2, line 1105: This test isn't using f

bodo/pandas/series.py
scott-routledge2: `the plan will not be a projection with an expr` What will the plan be in those cases?

bodo/pandas/utils.py
```diff
@@ -659,10 +671,18 @@ def wrap_plan(plan, res_id=None, nrows=None):
...
+        return proj.args[1][0]
+    else:
+        return make_col_ref_exprs([0], proj)[0]
+
+
+def get_proj_expr_keys(proj: LazyPlan):
+    """Get the single expression from a LogicalProjection node."""
+    if is_single_projection(proj):
+        return proj.args[0]
+    else:
```
scott-routledge2, line 686: If not a single projection, is proj expression?

scott-routledge2 (approved):
Thanks @DrTodd13

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -1043,3 +1043,132 @@ def g(df, a, b=1):
...
+        return df
```
ehsantn, line 1106: `f` is not used here. Also, what is the difference of this test and the one below?

bodo/pandas/utils.py
```diff
@@ -659,10 +671,18 @@ def wrap_plan(plan, res_id=None, nrows=None):
...
 def get_proj_expr_single(proj: LazyPlan):
     """Get the single expression from a LogicalProjection node."""
-    assert is_single_projection(proj), (
-        "get_proj_expr_single: LogicalProjection with a single expr expected"
-    )
-    return proj.args[1][0]
+    if is_single_projection(proj):
+        return proj.args[1][0]
+    else:
+        return make_col_ref_exprs([0], proj)[0]
```
ehsantn, line 679: Can we add some sort of assertion about `proj` being a Series plan?

bodo/pandas/utils.py
```diff
@@ -659,10 +671,18 @@ def wrap_plan(plan, res_id=None, nrows=None):
...
+
+
+def get_proj_expr_keys(proj: LazyPlan):
+    """Get the single expression from a LogicalProjection node."""
+    if is_single_projection(proj):
+        return proj.args[0]
+    else:
+        return proj
```
ehsantn, line 689: Same as above about checking Series plan. Also, the `get_proj_expr_keys` name seems confusing to me. This is just eliminating a trivial projection if present?

ehsantn (approved):
Thanks @DrTodd13. This is certainly an improvement. I think we need to think more about these areas for follow up PRs.

bodo/pandas/series.py
DrTodd13: It could be a source type node but it doesn't really matter because the fallback path applies a column ref expr to it and that can apply to any dataframe or series source.

bodo/tests/test_df_lib/test_end_to_end.py
DrTodd13: Fixed.

bodo/pandas/utils.py
DrTodd13: Could be anything.  Column Ref expr can apply to any source.

bodo/tests/test_df_lib/test_end_to_end.py
DrTodd13: Fixed.

bodo/pandas/utils.py
DrTodd13: Sure.  Done.

bodo/pandas/utils.py
DrTodd13: Done.

PR #555, opened 2025-06-10
Make send recv test deterministic

scott-routledge2:
Previously this test was using `hash` which is non-deterministic across ranks. I think this makes it more clear.

## Checklist

### DIFF
--- a/bodo/tests/test_distributed.py
+++ b/bodo/tests/test_distributed.py
@@ -2785,8 +2785,7 @@ def test_send_recv(val):
 
         return None
 
-    expected_out = val if bodo.get_rank() == recv_rank else None
-    check_func(impl, (val, recv_rank, send_rank), py_output=expected_out)
+    check_func(impl, (val, recv_rank, send_rank))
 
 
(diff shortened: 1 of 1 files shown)


bodo/tests/test_distributed.py
```diff
@@ -2783,7 +2785,8 @@ def impl(send_val, recv_rank, send_rank):
 
         return None
 
-    check_func(impl, (val, recv_rank, send_rank))
```
IsaacWarren: I think this was checking that worked in jit and nonjit right? Maybe it doesn't matter

scott-routledge2: Oh good point, it was hanging for me in the python case locally, which is why I changed it.

ehsantn (approved):

bodo/tests/test_distributed.py
scott-routledge2: I can open a separate ticket for this but basically it doesn't work when cache=True is passed to numba.njit. And I'm running it more than once...

PR #556, opened 2025-06-10
[BSE-4889] Support Series.str.cat

chrisoh7:

Adds support for Series.str.cat, and fallback for cases where `others=None`. 
Updates cases in `check_args_fallback` to handle `BodoStringMethods` fallback correctly.

## Testing strategy

PR CI, adds str.cat test cases. 

## User facing changes

Supports Series.str.cat. 

## Checklist

### DIFF
--- a/bodo/pandas/series.py
+++ b/bodo/pandas/series.py
@@ -34,6 +34,9 @@ from bodo.pandas.utils import (
     get_proj_expr_single,
     get_scalar_udf_result_type,
     get_single_proj_source_if_present,
+    is_arith_expr,
+    is_col_ref,
+    is_scalar_func,
     is_single_colref_projection,
     is_single_projection,
     make_col_ref_exprs,
@@ -613,10 +616,7 @@ class BodoStringMethods:
                 "implemented in Bodo dataframe library for the specified arguments yet. "
                 "Falling back to Pandas (may be slow or run out of memory)."
             )
-            # Prevents internal fallbacks causing a flood of warnings.
-            if not name.startswith("_"):
-                warnings.warn(BodoLibFallbackWarning(msg))
-
+            warnings.warn(BodoLibFallbackWarning(msg))
             return object.__getattribute__(pd.Series(self._series).str, name)
 
     @check_args_fallback("none")
@@ -736,26 +736,14 @@ def _str_partition_helper(s, col):
 
 def _str_cat_helper(df, sep, na_rep):
     """Concatenates df[idx] for idx in idx_pair, separated by sep."""
-    res = []
     if sep is None:
         sep = ""
-    lhs_idx, rhs_idx = 0, 1
 
-    lhs_col = df.iloc[:, lhs_idx]
-    rhs_col = df.iloc[:, rhs_idx]
+    # df is a two-column DataFrame created in zip_series_plan().
+    lhs_col = df.iloc[:, 0]
+    rhs_col = df.iloc[:, 1]
 
-    bitmap_lhs = lhs_col.isnull()
-    bitmap_rhs = rhs_col.isnull()
-
-    for i in df.index:
-        lhs = lhs_col[i] if not bitmap_lhs[i] else na_rep
-        rhs = rhs_col[i] if not bitmap_rhs[i] else na_rep
-        if (bitmap_lhs[i] or bitmap_rhs[i]) and na_rep is None:
-            res.append(pd.NA)
-        else:
-            res.append(f"{lhs}{sep}{rhs}")
-    series = pd.Series(res)
-    return series
+    return lhs_col.str.cat(rhs_col, sep, na_rep)
 
 
 def get_base_plan(plan):
@@ -779,10 +767,14 @@ def validate_str_cat(lhs, rhs):
 
     if lhs_list[0] != rhs_list[0]:
         raise BodoLibNotImplementedException(
-            "self and others are from distinct DataFrames: falling back to Pandas"
+            "str.cat(): self and others are from distinct DataFrames: falling back to Pandas"
         )
+
     # Ensures that at least 1 additional layer is present: single ColRefExpression at the least.
-    assert len(lhs_list) > 1 and len(rhs_list) > 1
+    if not (len(lhs_list) > 1 and len(rhs_list) > 1):
+        raise BodoLibNotImplementedException(
+            "str.cat(): plans should be longer than length 1: falling back to Pandas"
+        )
 
     return lhs_list, rhs_list
 
@@ -793,18 +785,6 @@ def get_list_projections(plan):
         return get_list_projections(plan.args[0]) + [plan]
     else:
         return [plan]
-
-
-def is_col_ref(expr):
-    return expr.plan_class == "ColRefExpression"
-
-
-def is_scalar_func(expr):
-    return expr.plan_class == "PythonScalarFuncExpression"
-
-
-def is_arith_expr(expr):
-    return expr.plan_class == "ArithOpExpression"
 
 
 def get_new_idx(idx, first, side):
@@ -862,6 +842,11 @@ def zip_series_plan(lhs, rhs) -> BodoSeries:
     n_index_arrays = get_n_index_arrays(lhs.index)
     n_cols = len(columns)
 
+    default_schema = pa.field("default", pa.large_string())
+    left_schema, right_schema = default_schema, default_schema
+    left_empty_data, right_empty_data = None, None
+    index = lhs_list[0].empty_data.index
+
     # Pads shorter list with None values.
     for i, (lhs_part, rhs_part) in enumerate(
         itertools.zip_longest(lhs_list[1:], rhs_list[1:], fillvalue=None)
@@ -872,12 +857,13 @@ def zip_series_plan(lhs, rhs) -> BodoSeries:
 
         # Extracts schema and empty_data from first layer of expressions.
         default_schema = pa.field("default", pa.large_string())
-        left_schema = (
-            left_expr.pa_schema[0] if left_expr is not None else default_schema
-        )
-        right_schema = (
-            right_expr.pa_schema[0] if right_expr is not None else default_schema
-        )
+
+        if left_expr is not None:
+            left_schema = left_expr.pa_schema[0]
+
+        if right_expr is not None:
+            right_schema = right_expr.pa_schema[0]
+
         schema = [left_schema, right_schema]
 
         # Create index metadata.
@@ -890,9 +876,16 @@ def zip_series_plan(lhs, rhs) -> BodoSeries:
         left_expr.empty_data.columns = ["lhs"]
         right_expr.empty_data.columns = ["rhs"]
 
-        if lhs_part is not None and rhs_part is not None:
-            empty_data = pd.concat([left_expr.empty_data, right_expr.empty_data])
-            empty_data.index = lhs_part.empty_data.index
+        if left_expr is not None:
+            left_empty_data = left_expr.empty_data
+
+        if right_expr is not None:
+            right_empty_data = right_expr.empty_data
+
+        assert left_empty_data is not None and right_empty_data is not None
+
+        empty_data = pd.concat([left_empty_data, right_empty_data])
+        empty_data.index = index
 
         result = LazyPlan(
             "LogicalProjection",
--- a/bodo/pandas/utils.py
+++ b/bodo/pandas/utils.py
@@ -1072,3 +1072,15 @@ def _get_df_python_func_plan(df_plan, empty_data, func, args, kwargs, is_method=
         (udf_arg,) + index_col_refs,
     )
     return wrap_plan(plan=plan)
+
+
+def is_col_ref(expr):
+    return expr.plan_class == "ColRefExpression"
+
+
+def is_scalar_func(expr):
+    return expr.plan_class == "PythonScalarFuncExpression"
+
+
+def is_arith_expr(expr):
+    return expr.plan_class == "ArithOpExpression"


bodo/pandas/series.py
```diff
@@ -578,9 +578,41 @@ def __getattribute__(self, name: str, /) -> pt.Any:
...
             )
-            warnings.warn(BodoLibFallbackWarning(msg))
+            # Prevents internal fallbacks causing a flood of warnings.
+            if not name.startswith("_"):
+                warnings.warn(BodoLibFallbackWarning(msg))
+
             return object.__getattribute__(pd.Series(self._series).str, name)
 
+    @check_args_fallback("none")
+    def cat(self, others=None, sep=None, na_rep=None, join="left"):
```
scott-routledge2, line 623: Can others be another BodoSeries here? What happens?

bodo/pandas/utils.py
```diff
@@ -92,6 +92,17 @@ def get_series_overloads():
...
 
+@functools.lru_cache
+def get_series_string_overloads():
+    """Return a list of the functions supported on BodoStringMethods objects
+    to some degree by bodo.jit.
+    """
+    from bodo.hiframes.series_str_impl import SeriesStrMethodType
+    from bodo.numba_compat import get_method_overloads
+
+    return get_method_overloads(SeriesStrMethodType)
```
scott-routledge2, line 103: We don't have fallback to jit support yet so this path is untested. I would just double check that the output seems reasonable.

scott-routledge2 (approved):
Couple of suggestions for testing, otherwise LGTM.

bodo/pandas/series.py
chrisoh7: It works the same. Interesting, just added as a test case.

bodo/pandas/utils.py
chrisoh7: Checked just now. Thanks

Output: `['len', 'split', 'get', 'join', 'replace', 'removeprefix', 'casefold', 'removesuffix', 'partition', 'contains', 'fullmatch', 'match', 'cat', 'count', 'find', 'rfind', 'index', 'rindex', 'slice_replace', 'repeat', 'ljust', 'rjust', 'center', 'pad', 'zfill', 'slice', 'startswith', 'endswith', 'encode', 'extract', 'extractall', 'capitalize', 'lower', 'lstrip', 'rstrip', 'strip', 'swapcase', 'title', 'upper', 'isalnum', 'isalpha', 'isdigit', 'isspace', 'islower', 'isupper', 'istitle', 'isnumeric', 'isdecimal', 'translate', 'get_dummies', 'wrap', 'findall', 'rpartition', 'rsplit', 'decode', 'normalize']`

bodo/pandas/series.py
```diff
@@ -578,9 +578,41 @@ def __getattribute__(self, name: str, /) -> pt.Any:
...
+            name=series.name,
+            index=index,
+        )
+
+        return _get_series_python_func_plan(
+            series._plan,
+            new_metadata,
+            "str.cat",
+            (),
+            {"others": others, "sep": sep, "na_rep": na_rep, "join": join},
```
ehsantn: How would this work in parallel without falling back to Pandas? `others` can be a Series/array/DataFrame and would need parallel element-wise computation.

chrisoh7: Yes, also just discovered this morning that others with type BodoSeries can't be supported. I'll work towards creating a new plan.

scott-routledge2: I think we would need to modify the PythonScalarFuncExpression to take in potentially multiple sources? Sort of like our `ArithOpExpression`? And propagate those changes everywhere.

bodo/pandas/series.py
```diff
@@ -690,6 +721,197 @@ def __getattribute__(self, name: str, /) -> pt.Any:
...
+    for i, (lhs_part, rhs_part) in enumerate(
+        itertools.zip_longest(lhs_list[1:], rhs_list[1:], fillvalue=None)
+    ):
+        # Create the plan for the shared part
+        left_expr = None if not lhs_part else lhs_part.args[1][0]
+        right_expr = None if not rhs_part else rhs_part.args[1][0]
+
+        # Extracts schema and empty_data from first layer of expressions.
+        if first:
+            schema = [left_expr.pa_schema[0], right_expr.pa_schema[0]]
```
scott-routledge2: Not sure if this will always work since the schemas might change over the course of executing the plan.

bodo/pandas/series.py
```diff
@@ -690,6 +721,197 @@ def __getattribute__(self, name: str, /) -> pt.Any:
...
+            "PythonScalarFuncExpression",
+            empty_data,
+            plan,
+            expr.args[1],
+            (idx,) + tuple(index_cols),
+        )
+    elif is_arith_expr(expr):
+        return expr
+    else:
+        raise NotImplementedError("Unsupported expr type:", expr.plan_class)
```
scott-routledge2: Raise `BodoLibNotImplementedException` ?

bodo/pandas/series.py
```diff
@@ -690,6 +721,197 @@ def __getattribute__(self, name: str, /) -> pt.Any:
...
+        empty_data = arrow_to_empty_df(pa.schema([expr.pa_schema[0]]))
+        return LazyPlan(
+            "PythonScalarFuncExpression",
+            empty_data,
+            plan,
+            expr.args[1],
+            (idx,) + tuple(index_cols),
+        )
+    elif is_arith_expr(expr):
+        return expr
```
scott-routledge2: This case seems unlikely to me but technically it is possible for the arith expr to be further along in the plan, which would require indices to be updated potentially

bodo/pandas/series.py
```diff
@@ -690,6 +721,197 @@ def __getattribute__(self, name: str, /) -> pt.Any:
...
+    """For first layer of expression, uses idx of itself. Otherwise, left=0 and right=1."""
+    if first:
+        return idx
+    elif side == "right":
+        return 1
+    else:
+        return 0
+
+
+def is_padding(expr):
```
scott-routledge2: Removing this helper and just inlining is more clear IMO.

bodo/pandas/series.py
```diff
@@ -690,6 +721,197 @@ def __getattribute__(self, name: str, /) -> pt.Any:
...
+
+def get_list_projections(plan):
+    """Returns list projections of plan."""
+    if is_single_projection(plan):
+        return get_list_projections(plan.args[0]) + [plan]
+    else:
+        return [plan]
+
+
+def print_plan_list(l):
```
scott-routledge2: Remove if unused?

bodo/pandas/series.py
```diff
@@ -690,6 +721,197 @@ def __getattribute__(self, name: str, /) -> pt.Any:
...
+    result = lhs_list[0]
+    schema, empty_data, first = [], None, True
+
+    # Initializes index columns info.
+    columns = lhs_list[0].empty_data.columns
+    n_index_arrays = get_n_index_arrays(lhs.index)
+    n_cols = len(columns)
+
+    # Ensures that at least 1 additional layer is present.
+    assert len(lhs_list) > 1 and len(rhs_list) > 1
```
scott-routledge2: I think these asserts and the ones above can be folded into the validation check. For example in validate you can do 
``` py
lhs = get_list_projections(left)
rhs = get_list projections(right)
if lhs[0] != rhs[0]:
  raise ...
```
I mention this because `get_list_projections` is doing the same thing as get_base_plan.

bodo/pandas/series.py
```diff
@@ -732,6 +954,44 @@ def _get_series_python_func_plan(series_proj, empty_data, func_name, args, kwarg
     )
 
 
+def _get_df_plan_python_func_plan(
```
scott-routledge2: Maybe we can refactor `_get_df_python_func_plan` to also take in a plan to remove duplicated code here? And refactor to utils.py.

scott-routledge2 (commented):
Thanks @chrisoh7, looks good overall. See my comments about potential edge cases

bodo/pandas/series.py
chrisoh7: Fixed so that if expr is None, replace with default pa.field of large string.

bodo/pandas/series.py
chrisoh7: Done.

bodo/pandas/series.py
chrisoh7: I'll create a follow-up issue to address this. Thanks!

bodo/pandas/plan_optimizer.pyx
```diff
@@ -549,6 +549,7 @@ cdef class PythonScalarFuncExpression(Expression):
         return f"PythonScalarFuncExpression({self.out_schema})"
 
 
+
```
scott-routledge2:
```suggestion
```

bodo/pandas/series.py
```diff
@@ -621,9 +612,50 @@ def __getattribute__(self, name: str, /) -> pt.Any:
...
+    def cat(self, others=None, sep=None, na_rep=None, join="left"):
+        """
+        If others is specified, concatenates the Series and elements of others
+        element-wise and returns a Series. If others is not passed, then falls back to
+        Pandas, and all values in the Series are concatenated into a single string with a given sep.
+        """
+        # Validates others is provided, falls back to Pandas otherwise
+        if others is None:
+            raise BodoLibNotImplementedException(
+                "others is not provided: falling back to Pandas"
```
```suggestion
                "str.cat(): others is not provided: falling back to Pandas"
``` And below as well

File: bodo/pandas/series.py [hunk 13]
```diff
@@ -690,6 +722,196 @@ def __getattribute__(self, name: str, /) -> pt.Any:
...
+    if first:
+        return idx
+    elif side == "right":
+        return 1
+    else:
+        return 0
+
+
+def make_expr(expr, plan, first, schema, index_cols, side="right"):
+    """TODO: add docstring"""
```
scott-routledge2: Commenting so this doesn't get forgot. Will also make it easier for others to review

bodo/pandas/series.py
```diff
@@ -690,6 +722,196 @@ def __getattribute__(self, name: str, /) -> pt.Any:
...
+    lhs_list, rhs_list = validate_str_cat(lhs, rhs)
+    result = lhs_list[0]
+    schema, empty_data, first = [], None, True
+
+    # Initializes index columns info.
+    columns = lhs_list[0].empty_data.columns
+    n_index_arrays = get_n_index_arrays(lhs.index)
+    n_cols = len(columns)
+
+    import itertools
```
scott-routledge2: Move to the top?

bodo/pandas/series.py
scott-routledge2: IMO it would be better to be more conservative here and just raise `BodoLibNotImplementedException` instead of potentially giving wrong results

bodo/tests/test_df_lib/test_series_str_cat.py
```diff
@@ -0,0 +1,123 @@
...
+            "A": [None, 1, 1, 2, 2, 3],
+            "E": [None, "  a", "b  ", "c", "d", "e"],
+            "C": [None, " 1", "     1     ", "h", "2", "3"],
+        },
+        index=["a", "b", "c", "d", "e", "f"],
+    )
+    return [df_noidx, df_idx]
+
+
+@pytest.mark.parametrize(
```
scott-routledge2, line 43: I would suggest adding couple test cases where we need to fallback/raise NotImplementedExeception to make sure that is working properly. You can see examples in test_frontend.py

bodo/tests/test_df_lib/test_series_str_cat.py
```diff
@@ -0,0 +1,123 @@
...
+        rhs_bd = rhs_expr(bdf)
+
+        out_pd = lhs_pd.str.cat(others=rhs_pd, sep="-", **kwargs)
+        out_bd = lhs_bd.str.cat(others=rhs_bd, sep="-", **kwargs)
+        out_bd = out_bd.execute_plan()
+
+        _test_equal(out_bd, out_pd, check_pandas_types=False, check_names=False)
+
+
+@pytest.mark.skip(
```
scott-routledge2, line 134: I am assuming we fallback in this case?

bodo/pandas/series.py
chrisoh7: ok.

bodo/pandas/series.py
chrisoh7: Thanks.

bodo/tests/test_df_lib/test_series_str_cat.py
chrisoh7: Yes.

bodo/pandas/series.py
```diff
@@ -621,9 +613,50 @@ def __getattribute__(self, name: str, /) -> pt.Any:
                 "implemented in Bodo dataframe library for the specified arguments yet. "
                 "Falling back to Pandas (may be slow or run out of memory)."
             )
-            warnings.warn(BodoLibFallbackWarning(msg))
+            # Prevents internal fallbacks causing a flood of warnings.
+            if not name.startswith("_"):
+                warnings.warn(BodoLibFallbackWarning(msg))
```
ehsantn: Why is this necessary? Seems a bit dangerous to me.

bodo/pandas/series.py
```diff
@@ -690,6 +723,196 @@ def __getattribute__(self, name: str, /) -> pt.Any:
...
+
+def _str_cat_helper(df, sep, na_rep):
+    """Concatenates df[idx] for idx in idx_pair, separated by sep."""
+    res = []
+    if sep is None:
+        sep = ""
+    lhs_idx, rhs_idx = 0, 1
+
+    lhs_col = df.iloc[:, lhs_idx]
+    rhs_col = df.iloc[:, rhs_idx]
```
ehsantn: Can we just call `lhs_col.str.cat(rhs_col, sep, na_rep)` instead of implementing from scratch?

bodo/pandas/series.py
```diff
@@ -690,6 +723,196 @@ def __getattribute__(self, name: str, /) -> pt.Any:
...
+    )
+    return series
+
+
+def _str_cat_helper(df, sep, na_rep):
+    """Concatenates df[idx] for idx in idx_pair, separated by sep."""
+    res = []
+    if sep is None:
+        sep = ""
+    lhs_idx, rhs_idx = 0, 1
```
ehsantn: Add a comment about `zip_series_plan` creating a two-column dataframe which is the expected input here?

bodo/pandas/series.py
```diff
@@ -690,6 +723,196 @@ def __getattribute__(self, name: str, /) -> pt.Any:
...
+
+    lhs_list = get_list_projections(lhs._plan)
+    rhs_list = get_list_projections(rhs._plan)
+
+    if lhs_list[0] != rhs_list[0]:
+        raise BodoLibNotImplementedException(
+            "self and others are from distinct DataFrames: falling back to Pandas"
+        )
+    # Ensures that at least 1 additional layer is present: single ColRefExpression at the least.
+    assert len(lhs_list) > 1 and len(rhs_list) > 1
```
ehsantn: Fall back instead of failing here (if this condition is possible)?

bodo/pandas/series.py
```diff
@@ -690,6 +723,196 @@ def __getattribute__(self, name: str, /) -> pt.Any:
...
+def is_col_ref(expr):
+    return expr.plan_class == "ColRefExpression"
+
+
+def is_scalar_func(expr):
+    return expr.plan_class == "PythonScalarFuncExpression"
+
+
+def is_arith_expr(expr):
+    return expr.plan_class == "ArithOpExpression"
```
ehsantn: Move these helpers to utils.py?

bodo/pandas/series.py
```diff
@@ -690,6 +723,196 @@ def __getattribute__(self, name: str, /) -> pt.Any:
...
+    # Pads shorter list with None values.
+    for i, (lhs_part, rhs_part) in enumerate(
+        itertools.zip_longest(lhs_list[1:], rhs_list[1:], fillvalue=None)
+    ):
+        # Create the plan for the shared part
+        left_expr = None if not lhs_part else lhs_part.args[1][0]
+        right_expr = None if not rhs_part else rhs_part.args[1][0]
+
+        # Extracts schema and empty_data from first layer of expressions.
+        default_schema = pa.field("default", pa.large_string())
```
ehsantn, line 859: String works here because we know input to `str.cat` is string? Can we use the left/right data type from the last non-None iteration?

bodo/pandas/series.py
```diff
@@ -690,6 +723,196 @@ def __getattribute__(self, name: str, /) -> pt.Any:
...
+
+        left_expr = make_expr(left_expr, result, first, schema, index_cols, "left")
+        right_expr = make_expr(right_expr, result, first, schema, index_cols)
+
+        left_expr.empty_data.columns = ["lhs"]
+        right_expr.empty_data.columns = ["rhs"]
+
+        if lhs_part is not None and rhs_part is not None:
+            empty_data = pd.concat([left_expr.empty_data, right_expr.empty_data])
+            empty_data.index = lhs_part.empty_data.index
```
ehsantn: What if the data type of the non-None side changes (may not be possible for `str.cat`)? Shouldn't we use the right/left data type of the last non-None iteration?

bodo/pandas/series.py
chrisoh7: Good idea, thanks for catching.

bodo/pandas/series.py
chrisoh7: Done.

bodo/pandas/series.py
chrisoh7: It prints multiple fallback messages for a couple of internal methods that are also falling back apparently. I will remove that line for now.

PR #557, opened 2025-06-11
Enable coverage tracking for df library tests

scott-routledge2:
## Testing strategy
Run CI and check coverage report for DF library info

## Checklist


IsaacWarren:
> > Thanks Scott, any idea why it works here but not for spawn tests?
> 
> I'm not sure what the issue was, I can try enabling in Spawn as well and see.
> 
> Edit: Also, IIRC this coverage tracking only works for code that's run on the spawner? Maybe that's why we turned it off.

Oh yeah that makes sense

PR #560, opened 2025-06-11
Series.map with engine=Bodo

scott-routledge2:
## Checklist

### DIFF
--- a/bodo/pandas_compat.py
+++ b/bodo/pandas_compat.py
@@ -3,8 +3,9 @@ import inspect
 import warnings
 
 import numpy as np
-import pandas as pd
 import pyarrow as pa
+
+import pandas as pd
 
 pandas_version = tuple(map(int, pd.__version__.split(".")[:2]))
 
@@ -62,6 +63,7 @@ if pandas_version < (1, 4):
 # Pandas code: https://github.com/pandas-dev/pandas/blob/ca60aab7340d9989d9428e11a51467658190bb6b/pandas/core/arrays/string_arrow.py#L141
 def ArrowStringArray__init__(self, values):
     import pyarrow as pa
+
     from pandas.core.arrays.string_ import StringDtype
     from pandas.core.arrays.string_arrow import ArrowStringArray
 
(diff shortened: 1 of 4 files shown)


bodo/tests/test_pandas_udf_compat.py
```diff
@@ -6,10 +6,10 @@
 """
 
 import numpy as np
-import pandas as pd
 import pytest
 
 import bodo
+import pandas as pd
```
scott-routledge2: From my precommit hook since building pandas from source, will revert before merging

pixi.lock
scott-routledge2, line 1: TODO: revert before merging

pixi.toml
scott-routledge2, line 1: TODO: revert before merging

bodo/pandas_compat.py
```diff
@@ -335,7 +335,27 @@ def map(
             decorator: Callable | None,
             skip_na: bool,
         ):
-            raise NotImplementedError("BodoExecutionEngine: map not implemented yet.")
+            if not isinstance(data, pd.Series):
+                raise ValueError(
```
scott-routledge2, line 339: Looking Pandas I'm not sure that this path is possible but just in case.

bodo/pandas_compat.py
```diff
@@ -335,7 +335,27 @@ def map(
...
+                    f"BodoExecutionEngine: map() expected input data to be Series, got: {type(data)}"
+                )
+
+            if skip_na:
+                raise ValueError(
+                    "BodoExecutionEngine: na_action not supported other than the default None for map()."
+                )
+
+            if args or kwargs:
```
scott-routledge2, line 348: It might be possible to support kwargs for map UDF case but I didn't see it anywhere in our unit tests, and I think it would require some kind of func_text since we don't support **kwargs?

ehsantn: Pandas API for Series.map doesn't support args/kwargs. But we need to support dict and na_action soon I think which shouldn't be difficult.
https://pandas.pydata.org/docs/reference/api/pandas.Series.map.html

ehsantn: Is it part of Pandas 3.0 APIs? If so, we can support it similar to DataFrame.apply().

scott-routledge2: Engine kwarg is not supported in the dict case (I tried to test it and it raised a Pandas error).

scott-routledge2: To answer your question about the kwargs yes, apparently it was added in Pandas 3.

scott-routledge2: I also think that we do support dict at least in some cases:
``` py
In [5]: x = pd.Series(["A", "B", "A", "B"])

In [6]: f(x)

In [7]: @bodo.jit
   ...: def f(x):
   ...:     return x.map({"A": 1, "B": 2})
   ...: 

In [8]: f(x)
Out[8]: 
0    1
1    2
2    1
3    2
dtype: int64[pyarrow]
```

PR #562, opened 2025-06-12
Support Iceberg write in dataframe library

ehsantn:

Adds support for DataFrame.to_iceberg similar to Pandas 3 APIs: https://github.com/pandas-dev/pandas/pull/61507/files
Reuses as much of the infrastructure as possible.

## Testing strategy

Added a unit test but need to wait for Iceberg CI to be enabled.

## User facing changes

DataFrame.to_iceberg now works.

## Checklist


bodo/io/arrow.cpp
```diff
@@ -103,9 +103,9 @@ PyObject* iceberg_pq_write_py_entry(
     const char* table_data_loc, table_info* in_table,
     array_info* in_col_names_arr, PyObject* partition_spec,
     PyObject* sort_order, const char* compression, bool is_parallel,
-    const char* bucket_region, int64_t row_group_size, char* iceberg_metadata,
-    PyObject* iceberg_arrow_schema_py, PyObject* arrow_fs,
-    UpdateSketchCollection* sketches);
```
IsaacWarren, line 108: Why change the sketches type?

bodo/io/iceberg_parquet_write.cpp
```diff
@@ -584,9 +584,11 @@ std::shared_ptr<arrow::Table> cast_arrow_table_to_iceberg_schema(
  */
 std::vector<PyObject *> iceberg_pq_write_helper(
     const char *fpath, const std::shared_ptr<table_info> table,
-    const std::shared_ptr<array_info> col_names_arr, const char *compression,
-    bool is_parallel, const char *bucket_region, int64_t row_group_size,
-    char *iceberg_metadata, std::shared_ptr<arrow::Schema> iceberg_arrow_schema,
+    const std::variant<std::shared_ptr<array_info>, std::vector<std::string>>
+        col_names_arr,
```
IsaacWarren: IMO it's cleaner to make this a vector of strings and have the caller call array_to_string_vector

bodo/pandas/_bodo_write_function.h
```diff
@@ -0,0 +1,141 @@
...
+    std::shared_ptr<arrow::Schema> in_schema;
+    std::string table_loc;
+    std::string bucket_region;
+    int64_t max_pq_chunksize;
+    std::string compression;
+    PyObject *partition_tuples;
+    PyObject *sort_tuples;
+    std::string iceberg_schema_str;
+    std::shared_ptr<arrow::Schema> iceberg_schema;
+    std::shared_ptr<arrow::fs::FileSystem> fs;
```
IsaacWarren, line 140: These can all be const I think

bodo/pandas/_executor.h
```diff
@@ -36,7 +36,7 @@ class Executor {
     /**
      * @brief Execute the plan and return the result.
      */
-    std::shared_ptr<table_info> ExecutePipelines() {
+    std::variant<std::shared_ptr<table_info>, PyObject*> ExecutePipelines() {
```
IsaacWarren, line 39: Why do we need this?

bodo/pandas/frame.py
```diff
@@ -290,6 +290,101 @@ def to_parquet(
...
+
+        # Register file names, metrics and schema in transaction
+        success = bodo.io.iceberg.write.register_table_write_seq(
+            txn,
+            fnames,
+            file_records,
+            partition_infos,
+            partition_spec,
+            sort_order_id,
+        )
```
IsaacWarren, line 384: IMO this should happen in finalization for the iceberg write node, that way we don't need to return anything and the pipeline execution signature can stay the same

bodo/pandas/physical/write_iceberg.h
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        PyObject* gather_method = PyObject_GetAttrString(comm_world, "gather");
+        if (!gather_method) {
+            Py_DECREF(comm_world);
+            Py_DECREF(mpi_module);
+            throw std::runtime_error("Failed to get gather method");
+        }
+
+        PyObject* args = PyTuple_New(1);
```
IsaacWarren: All of these being pyobjectptrs would simplify the error handling

bodo/pandas/physical/write_iceberg.h
```diff
@@ -0,0 +1,156 @@
...
+                       : OperatorResult::NEED_MORE_INPUT;
+    }
+
+    void Finalize() override {
+        // Gather iceberg_files_info from all ranks using MPI
+        // Equivalent to all_infos = comm.gather(iceberg_files_info_py)
+
+        PyObject* mpi_module = PyImport_ImportModule("bodo.mpi4py.MPI");
+        if (!mpi_module) {
+            throw std::runtime_error("Failed to import bodo.mpi4py.MPI");
```
IsaacWarren: Can we add PhysicalWriteIceberg::Finalize as a prefix to these error messages

bodo/io/iceberg/write.py
```diff
@@ -365,7 +365,16 @@ def generate_data_file_info(
     file_infos = (
         [item for sub in combined_data for item in sub] if combined_data else None
     )
+    return generate_data_file_info_seq(file_infos)
 
+
+def generate_data_file_info_seq(
+    file_infos: list[list[tuple[pt.Any, pt.Any, pt.Any]]],
```
scott-routledge2:
```suggestion
    file_infos: list[tuple[pt.Any, pt.Any, pt.Any]] | None,
``` 
I think the list is flattened by the caller?

bodo/pandas/physical/write_iceberg.h
ehsantn: Done.

bodo/pandas/physical/write_iceberg.h
ehsantn: It's a matter of preference but I think explicit handling is simpler here since the pointers are created and decrefed locally in small code, and also there is `PyTuple_SetItem` that steals a reference.

bodo/io/arrow.cpp
ehsantn: Including the header for `UpdateSketchCollection` in `write_iceberg.h` causes a bunch of compilation issues (didn't seem worth fixing).

bodo/pandas/frame.py
ehsantn: I wanted to go that route, but having initialization and finalization in the same place simplifies understanding the Iceberg write code which is worth the ugly pipeline execution signature I think. The Iceberg code is difficult to understand since it's sprawling so limiting the sprawl is useful here.

bodo/pandas/_executor.h
ehsantn: For returning file lists in Iceberg write.

bodo/io/iceberg_parquet_write.cpp
ehsantn: Done.

bodo/pandas/_bodo_write_function.h
ehsantn: The code using std::move later on them (not sure how much it matters).

bodo/tests/test_df_lib/test_iceberg.py
```diff
@@ -527,3 +528,36 @@ def test_table_read_partitioned_file_pruning(
...
+            "two": ["foo", "bar", "baz", "foo", "bar", "baz", "foo"],
+            "three": [True, False, True, True, True, False, False],
+            "four": [-1.0, 5.1, 2.5, 3.0, 4.0, 6.0, 11.0],
+            "five": ["foo", "bar", "baz", None, "bar", "baz", "foo"],
+        }
+    )
+
+    bdf = bpd.from_pandas(df)
+    catalog_properties = {
+        pyiceberg.catalog.PY_CATALOG_IMPL: "bodo.io.iceberg.catalog.dir.DirCatalog",
```
scott-routledge2, line 548: Maybe we could provide a simpler way of constructing this dictionary if we are going to use it in our test/examples.
Something like `DirCatalog(path)`

scott-routledge2 (approved):
LGTM, thanks @ehsantn

bodo/io/arrow.cpp
IsaacWarren: Add a comment? That's really weird

bodo/pandas/frame.py
IsaacWarren: It's not just the signature, it's also having to handle the different cases in other places too. Generally the Python code just gets any information necessary for the node, it doesn't perform any operations, IMO these should all happen in the node where possible otherwise it's not really lazy. I understand that on write you have to execute after but it becomes inconsistent. Just wanted to give my two cents

bodo/tests/test_df_lib/test_iceberg.py
ehsantn: I'm working on supporting `bdf.to_iceberg("test_table", location="/path/to/iceberg_warehouse")` in follow up PR.

bodo/pandas/frame.py
ehsantn: Thanks for the feedback. I agree with your points but still think simplifying the code here is worth the downsides. We can get feedback from the team moving forward.

bodo/io/arrow.cpp
ehsantn: Done.

bodo/io/iceberg/write.py
ehsantn: Done.

PR #563, opened 2025-06-12
BSE-4853: Pleias Bad English OCR Correction

IsaacWarren:

As title, looking for review before I kick off run on the whole dataset.
A possible extension would be to use all available CPUs for tokenization and have ranks that don't have a GPU assigned send their chunk to gpu ranks but at this point the actual inference is by far the bottleneck so seemed like a low priority.
This should work for instances with multiple GPUs as well if BODO_NUM_WORKERS is set correctly but I haven't tested it.

## Testing strategy
Tested on 10 rows of the input data on a 1 instance cluster.

## User facing changes
N/A

## Checklist

### DIFF
--- a/examples/AI/pleias_ocr_correction.py
+++ b/examples/AI/pleias_ocr_correction.py
@@ -10,9 +10,6 @@ model_name = "LLMDH/pleias_350m_ocr"
 model_len = 2048
 os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"
 
-prompts = pd.read_parquet(
-    "hf://datasets/LLMDH/English-PD-bad-OCR/**/*.parquet"
-)
 tokenizer = None
 
 def tokenize(row):
@@ -74,7 +71,7 @@ def ocr_correction(prompts):
     # Maps batch index and batch entry index to the original prompt indices
     batch_to_prompt_idx = [[]]
     while prompt_idx < len(prompts):
-        split_encoded_prompts = prompts["split_encoded_prompts"][prompt_idx]
+        split_encoded_prompts = prompts["split_encoded_prompts"].iloc[prompt_idx]
 
         split_prompt_idx = 0
         while split_prompt_idx < len(split_encoded_prompts):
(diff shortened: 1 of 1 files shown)


examples/AI/pleias_ocr_correction.py
```diff
@@ -0,0 +1,107 @@
...
+
+    # Split the tokenized row into multiple chunks
+    split_row_encoded = [
+        row_encoded[i : i + num_row_tokens]
+        for i in range(0, len(row_encoded), num_row_tokens)
+    ]
+
+    # Add section markers around each chunk (drop the stop token from the tokenizer output since it's in the text_start)
+    split_row_encoded = [
+        text_start + (split_row_encoded_chunk[1:] if split_row_encoded_chunk[0] == 2 else split_row_encoded_chunk) + text_end + ocr_correction_start
```
scott-routledge2, line 43: Does the tokenizer also leave a text_end token that would need to be replaced for the last chunk?

examples/AI/pleias_ocr_correction.py
```diff
@@ -0,0 +1,107 @@
...
+                batches.append([gen_prompt])
+                batch_to_prompt_idx.append([prompt_idx])
+            split_prompt_idx += 1
+        prompt_idx += 1
+
+    text_results = pd.Series([""] * len(prompts), dtype=ArrowDtype(pa.large_string()))
+    for i, batch in enumerate(batches):
+        batch_result = llm.generate(prompts=batch, sampling_params=sampling_params)
+        for j, result in enumerate(batch_result):
+            text = "".join([output.text for output in result.outputs])
```
scott-routledge2, line 95: The model outputs possibly multiple responses per prompt?

scott-routledge2 (approved):
Thanks Isaac! Couple questions.

strangeloopcanon (approved):

examples/AI/pleias_ocr_correction.py
```diff
@@ -0,0 +1,107 @@
...
+from transformers import AutoTokenizer
+
+batch_size = 350
+model_name = "LLMDH/pleias_350m_ocr"
+model_len = 2048
+os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"
+
+prompts = pd.read_parquet(
+    "hf://datasets/LLMDH/English-PD-bad-OCR/**/*.parquet"
+)
```
ehsantn: Minor: I'd move this to the top-level compute part at the end and move all top-level compute calls inside a main function for clarity.

examples/AI/pleias_ocr_correction.py
```diff
@@ -0,0 +1,107 @@
...
+        max_tokens=model_len,
+    )
+
+    # Organize the split prompts into batches that fit within the batch size
+    batches = [[]]
+    prompt_idx = 0
+    # Maps batch index and batch entry index to the original prompt indices
+    batch_to_prompt_idx = [[]]
+    while prompt_idx < len(prompts):
+        split_encoded_prompts = prompts["split_encoded_prompts"][prompt_idx]
```
ehsantn: Use `iloc` to avoid Pandas Index surprises?

examples/AI/pleias_ocr_correction.py
IsaacWarren: Depending on how you configure it yeah, in our case it will always be one but it's still a list

examples/AI/pleias_ocr_correction.py
IsaacWarren: Not that I can tell from testing

PR #566, opened 2025-06-12
Fix NA handling in groupby to Python semantics

scott-routledge2:
## Checklist

### DIFF
--- a/bodo/libs/streaming/_groupby.cpp
+++ b/bodo/libs/streaming/_groupby.cpp
@@ -2561,7 +2561,8 @@ GroupbyState::GroupbyState(
     bool parallel_, int64_t sync_iter_, int64_t op_id_,
     int64_t op_pool_size_bytes_, bool allow_any_work_stealing,
     std::optional<std::vector<std::shared_ptr<DictionaryBuilder>>>
-        key_dict_builders_)
+        key_dict_builders_,
+    bool use_sql_rules)
     :  // Create the operator buffer pool
       op_pool(std::make_unique<bodo::OperatorBufferPool>(
           op_id_,
@@ -2712,9 +2713,7 @@ GroupbyState::GroupbyState(
     // safely set skip_na_data to true for all SQL aggregations. There is an
     // issue to fix this behavior so that use_sql_rules trumps the value of
     // skip_na_data: https://bodo.atlassian.net/browse/BSE-841
-    bool skip_na_data = true;
-    bool use_sql_rules = true;
-
+    bool skip_na_data = use_sql_rules;
     if (!ftypes.empty() && ftypes[0] == Bodo_FTypes::window) {
         // Handle a collection of window functions.
         this->accumulate_before_update = true;
--- a/bodo/pandas/physical/aggregate.h
+++ b/bodo/pandas/physical/aggregate.h
@@ -109,7 +109,8 @@ class PhysicalAggregate : public PhysicalSource, public PhysicalSink {
             std::make_unique<bodo::Schema>(*in_table_schema_reordered), ftypes,
             std::vector<int32_t>(), f_in_offsets, f_in_cols, this->keys.size(),
             std::vector<bool>(), std::vector<bool>(), cols_to_keep_vec, nullptr,
-            get_streaming_batch_size(), true, -1, -1, -1);
+            get_streaming_batch_size(), true, -1, -1, -1, false, std::nullopt,
+            /*use SQL rules*/ false);
     }
 
     virtual ~PhysicalAggregate() = default;
--- a/bodo/tests/test_groupby.py
+++ b/bodo/tests/test_groupby.py
@@ -197,6 +197,7 @@ def test_df_int_no_null(request):
 
 
 @pytest.mark.slow
+@pytest.mark.df_lib
 def test_nullable_int(memory_leak_check):
     def impl(df):
         A = df.groupby("A").sum()
@@ -253,6 +254,7 @@ def test_nullable_int(memory_leak_check):
     check_func(impl_select_colH, (df,), sort_output=True, check_dtype=False)
 
 
+@pytest.mark.df_lib
 def test_groupby_sum_integer_upcast(memory_leak_check):
     """
     Test that we upcast output of groupby sum on integers to their
@@ -355,6 +357,7 @@ def test_groupby_sum_integer_upcast(memory_leak_check):
 
 
 @pytest.mark.slow
+@pytest.mark.df_lib
 def test_groupby_nullable_float(memory_leak_check):
     def impl(df):
         A = df.groupby("A").sum()
@@ -444,6 +447,7 @@ def test_groupby_df_numpy_bool(memory_leak_check):
     check_func(impl, (), sort_output=True, reset_index=True)
 
 
+@pytest.mark.df_lib  # TODO: count not implemented yet for DF Library
 @pytest_mark_pandas
 def test_all_null_keys(memory_leak_check):
     """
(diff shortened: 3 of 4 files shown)


bodo/libs/streaming/_groupby.cpp
```diff
@@ -4229,6 +4229,52 @@ uint64_t GroupbyState::op_pool_bytes_allocated() const {
...
+ *
+ * @param in_table input batch to groupby.
+ * @param n_keys number of key columns in input
+ * @return std::shared_ptr<table_info> input table with NAs filtered out
+ */
+std::shared_ptr<table_info> filter_na_values(
+    std::shared_ptr<table_info> in_table, uint64_t n_keys) {
+    bodo::vector<bool> not_na(in_table->nrows(), true);
+    bool can_have_na = false;
+    for (uint64_t i = 0; i < n_keys; i++) {
```
scott-routledge2, line 4245: Since coming from Arrow I think we can remove this optimization since they will always be nullable?

scott-routledge2:
Forgot to update frame.py/the test. It should be ready for review now.

bodo/tests/test_groupby.py
```diff
@@ -444,6 +447,7 @@ def impl():
     check_func(impl, (), sort_output=True, reset_index=True)
 
 
+@pytest.mark.df_lib  # TODO: count not implemented yet for DF Library
```
IsaacWarren: Should this be skipped? Or remove the todo?

scott-routledge2: I'll remove for now since I don't think we are gaining anything from including it

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -982,24 +982,28 @@ def test_series_sort(datapath):
...
+def test_basic_groupby(dropna):
     """
     Test a simple groupby operation.
     """
     df1 = pd.DataFrame(
         {
             "B": ["a1", "b11", "c111"] * 2,
             "E": [1.1, 2.2, 13.3] * 2,
-            "A": pd.array([2, 2, 3] * 2, "Int64"),
+            "A": pd.array([pd.NA, 2, 3] * 2, "Int64"),
```
ehsantn: Add NAs to column E to test NAs in value columns too?

bodo/pandas/physical/aggregate.h
```diff
@@ -98,6 +99,18 @@ class PhysicalAggregate : public PhysicalSource, public PhysicalSink {
...
+            BodoAggFunctionData& bind_info =
+                agg_expr.bind_info->Cast<BodoAggFunctionData>();
+
+            // NOTE: drop_na must be consistent accross all expressions
+            // This is a little awkward but AFAICT there is no bind_info for
+            // the LogicalAggregate.
+            if (!dropna.has_value()) {
+                dropna = bind_info.dropna;
+            }
+            assert(dropna.value() == bind_info.dropna);
```
ehsantn: Maybe a runtime exception here just in case?

bodo/libs/streaming/_groupby.cpp
```diff
@@ -2712,9 +2714,7 @@ GroupbyState::GroupbyState(
     // safely set skip_na_data to true for all SQL aggregations. There is an
     // issue to fix this behavior so that use_sql_rules trumps the value of
     // skip_na_data: https://bodo.atlassian.net/browse/BSE-841
-    bool skip_na_data = true;
-    bool use_sql_rules = true;
-
+    bool skip_na_data = use_sql_rules;
```
ehsantn: Looks like `skip_na_data` should be true for Pandas as well. We set it to true in JIT path here: https://github.com/bodo-ai/Bodo/blob/302e4f06f4f6ff3746f1f113fcae83ab2dc6e6dc/bodo/ir/aggregate.py#L522

Here is Pandas example code skipping NAs:
```python
   ...: df = pd.DataFrame(
   ...:     {
   ...:         "A": pd.array([pd.NA, 2, 3, 7] * 3, "Int64"),
   ...:         "B": ["A1", "B1", "C1", "Abc"] * 3,
   ...:         "C": pd.array([6, pd.NA, 4, 4, 1, 3] * 2, "Int64"),
   ...:     }
   ...: )
In [10]: df.groupby("A")["C"].sum()
Out[10]:
A
2     7
3    11
7     7
Name: C, dtype: Int64
```

bodo/libs/streaming/_groupby.cpp
```diff
@@ -4229,6 +4229,52 @@ uint64_t GroupbyState::op_pool_bytes_allocated() const {
...
+    for (size_t i = 0; i < in_table->nrows(); i++) {
+        if (not_na[i]) {
+            idx_list.emplace_back(i);
+        }
+    }
+    if (idx_list.size() == in_table->nrows()) {
+        // No NA values, skip the copy.
+        return in_table;
+    } else {
+        return RetrieveTable(std::move(in_table), std::move(idx_list));
```
ehsantn, line 4272: Minor optimization: generate `idx_list` in this code path only to avoid overheads in fast path above?

bodo/libs/streaming/_groupby.cpp
```diff
@@ -4229,6 +4229,52 @@ uint64_t GroupbyState::op_pool_bytes_allocated() const {
...
+                           col_not_na.begin(), not_na.begin(),
+                           std::logical_and<>());
+        }
+    }
+    if (!can_have_na) {
+        // No NA values, just return.
+        return in_table;
+    }
+
+    // Retrieve table takes a list of columns. Convert the boolean array.
```
ehsantn: Let's open an issue to improve performance of this later.

bodo/libs/streaming/_groupby.cpp
```diff
@@ -4267,6 +4313,11 @@ bool groupby_agg_build_consume_batch(GroupbyState* groupby_state,
     MPI_Comm_size(MPI_COMM_WORLD, &n_pes);
     MPI_Comm_rank(MPI_COMM_WORLD, &myrank);
 
+    // Filter NA keys in Pandas case.
+    if (groupby_state->pandas_drop_na) {
+        in_table = filter_na_values(std::move(in_table), groupby_state->n_keys);
```
ehsantn: `filter_na_keys` is more descriptive to me but not a big deal.

ehsantn (commented):
Thanks @scott-routledge2 . See comments especially about `skip_na_data` and testing.

bodo/libs/streaming/_groupby.cpp
scott-routledge2: Thanks! Will fix. Also, wouldn't this be an example of dropna=True? The skip_na_data doesn't do anything in this particular case in my testing.

bodo/libs/streaming/_groupby.cpp
scott-routledge2: But don't we use `idx_list` to determine whether we can take the fast path?

bodo/libs/streaming/_groupby.cpp
ehsantn: What do you mean? `dropna=True` is on by default but NA data are different rows than keys here. Not sure why `skip_na_data` doesn't do anything.

bodo/libs/streaming/_groupby.cpp
ehsantn: Isn't sum of `not_na` equivalent to `idx_list.size()`?

bodo/libs/streaming/_groupby.cpp
scott-routledge2: If you look at where `skip_na_data` is used, it's only used for certain functions like nunique, not sum. From testing it looks like we do the correct behavior. See https://github.com/bodo-ai/Bodo/blob/302e4f06f4f6ff3746f1f113fcae83ab2dc6e6dc/bodo/libs/groupby/_groupby_col_set.cpp#L2102

bodo/libs/streaming/_groupby.cpp
scott-routledge2: Yep, you're right. I rearranged things slightly just to add a bool contains_na and set it in previous loop.

ehsantn (approved):
Thanks for the fixes.

PR #572, opened 2025-06-17
Convert relative Iceberg paths to absolute paths in directory catalog

ehsantn:
Fixes a bug where relative paths are written to Iceberg metadata, breaking reads from other working directories. Tested it manually since not easy to test in current infrastructure.


scott-routledge2 (approved):
Thanks @ehsantn

PR #575, opened 2025-06-17
[BSE-4904] Support DataFrameGroupby.sum()

scott-routledge2:

- Adds DataFrameGroupby.sum
- Adds fallback for DataFrameGroubpy and SeriesGroupby methods
- Adds upcasting for sum to upcast int8/16/32 -> int64, uint8,... -> uint64.. etc, also checks for unsupported columns cases. Currently the strategy is to fallback if unsupported columns, but it would ideally raise ValueError once we support all the same cases as Pandas. To that point, we are currently missing sum for list and decimal. Decimal not working is surprising to me, will followup on that and investigate.

There are a few corner cases which are not fully supported (can be followup in next PR):
- Fix corner case: selecting the same column we group on in the aggregate.
- Fix corner case: throwing away agg values keeps dropna behavior the same.
- Decimal sum is broken. I know the issue, but it may require refactoring a lot of our C++ groupby code to pass precision/scale around so I will leave it as a followup.

## Testing strategy
 PR CI + local testing

## User facing changes

groupby.sum(), more fallbacks,

## Checklist

### DIFF
--- a/bodo/pandas/groupby.py
+++ b/bodo/pandas/groupby.py
@@ -45,7 +45,7 @@ class DataFrameGroupBy:
         self._selection = (
             selection
             if selection is not None
-            else [col for col in obj.columns if col not in keys]
+            else list(filter(lambda col: col not in keys, obj.columns))
         )
 
     def __getitem__(self, key) -> DataFrameGroupBy | SeriesGroupBy:
@@ -208,7 +208,7 @@ def _groupby_sum(
 def _get_agg_output_type(func: str, pa_type: pa.DataType, col_name: str) -> pa.DataType:
     """Gets the output type of an aggregation or raise ValueError for
     unsupported pa_type/func combinations. Should closely match
-    get_groupby_output_dtype.
+    https://github.com/bodo-ai/Bodo/blob/d1133e257662348cc7b9ef52cf445633036133d2/bodo/libs/groupby/_groupby_common.cpp#L562
     """
     new_type = pa_type
     if func == "sum":


bodo/pandas/utils.py
```diff
@@ -92,6 +92,17 @@ def get_series_overloads():
...
 
+@functools.lru_cache
+def get_groupby_overloads():
+    """Return a list of the functions supported on DataFrameGroupby objects
+    to some degree by bodo.jit.
+    """
+    from bodo.hiframes.pd_groupby_ext import DataFrameGroupByType
+    from bodo.numba_compat import get_method_overloads
+
+    return get_method_overloads(DataFrameGroupByType)
```
scott-routledge2, line 114: Overload methods:
```
['value_counts', 'groups', 'indices', 'rank', 'cov', 'tail', 'backfill', 'pct_change', 'cumcount', '__iter__', 'tshift', 'take', 'resample', 'corr', 'cummin', 'plot', 'any', 'pad', 'fillna', 'nth', 'bfill', 'all', 'filter', 'describe', 'cummax', 'get_group', 'cumprod', 'ohlc', 'sample', 'diff', 'sem', 'quantile', 'ffill', 'hist', 'is_monotonic_increasing', 'is_monotonic_decreasing', 'unique', 'nlargest', 'nsmallest', 'corrwith', 'boxplot', 'rolling']
``` 
I'm not sure if all of these apply to Series as well?
Either way we can't test this path because of no JIT fallback support.

bodo/pandas/groupby.py
```diff
@@ -163,3 +142,115 @@ def __getattribute__(self, name: str, /) -> Any:
             warnings.warn(BodoLibFallbackWarning(msg))
             gb = pd.DataFrame(self._obj).groupby(self._keys)[self._selection[0]]
             return object.__getattribute__(gb, name)
+
+
+def _groupby_sum(
```
scott-routledge2, line 151: subsequent PRs can generalize this more to apply a function or mapping of function->column names to columns in a DataFrame/Series

scott-routledge2:
Looks like there are still a couple tests failing, will fix first before requesting review..

scott-routledge2:
> Looks like there are still a couple tests failing, will fix first before requesting review..

Everything should be passing now

bodo/pandas/groupby.py
```diff
@@ -37,7 +42,11 @@ def __init__(
         self._keys = keys
         self._as_index = as_index
         self._dropna = dropna
-        self._selection = selection
+        self._selection = (
+            selection
+            if selection is not None
+            else [col for col in obj.columns if col not in keys]
```
DrTodd13: Maybe using filter is clearer?

bodo/pandas/groupby.py
```diff
@@ -163,3 +146,115 @@ def __getattribute__(self, name: str, /) -> Any:
...
+            exprs,
+        )
+
+    return wrap_plan(plan)
+
+
+def _get_agg_output_type(func: str, pa_type: pa.DataType, col_name: str) -> pa.DataType:
+    """Gets the output type of an aggregation or raise ValueError for
+    unsupported pa_type/func combinations. Should closely match
+    get_groupby_output_dtype.
```
DrTodd13: What is that and where is it?

bodo/pandas/groupby.py
```diff
@@ -163,3 +146,115 @@ def __getattribute__(self, name: str, /) -> Any:
...
+    """
+    new_type = pa_type
+    if func == "sum":
```
DrTodd13, line 214: Is there a way to simulate this on zero data in Pandas or something else?  Feels like we ought to be able to get this from somewhere else.

scott-routledge2: Pandas might optimize for the empty_data case and not do casting? I will look into it.

bodo/pandas/groupby.py
scott-routledge2: It's in C++, I'll include a link, it's how we create the output buffers for aggregate

PR #576, opened 2025-06-17
First batch of changes from NYC taxi and TPCH work.

DrTodd13:

1) Source plan caching fix.
2) Implement bound_between.
3) Support rename.
4) Add test for setitem of arithmetic expressions.
5) Fix setitem to allow arithmetic expressions.

## Testing strategy

run_ci - source plan caching and setitem arithmetic expressions are by the tests.  bound_between and rename should only be triggered by nyc taxi and tpch q5 at this point and have been tested locally.

## User facing changes

None

## Checklist

### DIFF
--- a/bodo/pandas/frame.py
+++ b/bodo/pandas/frame.py
@@ -277,6 +277,12 @@ class BodoDataFrame(pd.DataFrame, BodoLazyWrapper):
         errors: IgnoreRaise = "ignore",
     ) -> BodoDataFrame | None:
         orig_plan = self._plan
+        # TODO: The value of copy here is ignored.
+        # Most cases of copy=False we have are the idiom A=A.rename(copy=False) where there is still
+        # only one possible reference to the data so we can treat this case like copy=True since for us
+        # we only materialize as needed and so copying isn't an overhead.
+        if copy == False:
+            warnings.warn("BodoDataFrame::rename copy=False argument ignored assuming A=A.rename(copy=False) idiom.")
         renamed_plan = LazyPlan(
             orig_plan.plan_class,
             orig_plan.empty_data.rename(columns=columns),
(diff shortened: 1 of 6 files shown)


bodo/pandas/physical/expression.cpp
```diff
@@ -329,6 +332,39 @@ std::shared_ptr<PhysicalExpression> buildPhysicalExprTree(
                     buildPhysicalExprTree(bce.child, col_ref_map, no_scalars),
                     bce.return_type));
         } break;  // suppress wrong fallthrough error
+        case duckdb::ExpressionClass::BOUND_BETWEEN: {
```
ehsantn, line 335: Is it possible/practical to add unit tests for this at this point?

bodo/pandas/frame.py
```diff
@@ -243,6 +263,28 @@ def shape(self):
             case ExecState.COLLECTED:
                 return super().shape
 
+    @check_args_fallback(supported=["columns", "copy"])
```
ehsantn, line 266: `copy` parameter isn't handled.

ehsantn: Also a unit test would be great.

bodo/pandas/physical/expression.cpp
DrTodd13: done.

bodo/pandas/frame.py
DrTodd13: copy was intentional.  Both TPCH and nyc  uses rename but rename only uses the A=A.rename(...., copy=False) idiom which still means there is only one copy but if copy=False not in that idiom then it is tricky.  With the idiom NYC uses, we can treat it like a copy=True.  There's no way to detect this idiom that I know of.

PR #580, opened 2025-06-18
[BSE-4908] Support Series.str.extract

chrisoh7:

Implements `Series.str.extract`. 
Adds support for  `Series.str.extract`. 
Refactors helper functions used in extract and partition to correctly propagate index column data. 

## Testing strategy

PR CI, adds test cases for extract in test case generator. 

## User facing changes

Support for Series.str.extract. 

## Checklist

### DIFF
--- a/bodo/pandas/series.py
+++ b/bodo/pandas/series.py
@@ -691,6 +691,11 @@ class BodoStringMethods:
 
         pattern = re.compile(pat, flags=flags)
         n_cols = pattern.groups
+
+        # Like Pandas' implementation, raises ValueError when there are no capture groups.
+        if n_cols == 0:
+            raise ValueError("pattern contains no capture groups")
+
         group_names = pattern.groupindex
         is_series_output = not expand and n_cols == 1  # In this case, returns a series.
 
@@ -717,6 +722,7 @@ class BodoStringMethods:
             is_method=False,
         )
 
+        # expand=False and n_cols=1: returns series
         if is_series_output:
             return series_out
 
@@ -878,21 +884,27 @@ def gen_partition(name):
         if not expand:
             return series_out
 
+        n_index_arrays = get_n_index_arrays(series_out.index)
+        index_cols = tuple(range(1, 1 + n_index_arrays))
+        index_col_refs = tuple(make_col_ref_exprs(index_cols, series_out._plan))
+
         # Create schema for output DataFrame with 3 columns
         arrow_schema = pa.schema(
             [pa.field(f"{idx}", pa.large_string()) for idx in range(3)]
         )
         empty_data = arrow_to_empty_df(arrow_schema)
-        empty_series = pd.Series([], dtype=pd.ArrowDtype(pa.large_string()))
+        empty_data.index = series_out._plan.empty_data.index
 
-        expr = tuple(create_expr(idx, empty_series, series_out) for idx in range(3))
+        expr = tuple(
+            create_expr(idx, empty_data, series_out, index_cols) for idx in range(3)
+        )
 
         # Creates DataFrame with 3 columns
         df_plan = LazyPlan(
             "LogicalProjection",
             empty_data,
             series_out._plan,
-            expr,
+            expr + index_col_refs,
         )
 
         return wrap_plan(plan=df_plan)
--- a/bodo/pandas/series.py
+++ b/bodo/pandas/series.py
@@ -580,7 +580,7 @@ class BodoSeries(pd.Series, BodoLazyWrapper):
         )
 
 
-def _str_partition_helper(s, col):
+def _get_col_as_series(s, col):
     """Extracts column col from list series and returns as Pandas series."""
     series = pd.Series(
         [
@@ -589,6 +589,27 @@ def _str_partition_helper(s, col):
         ]
     )
     return series
+
+
+def _str_extract_helper(s, pattern, flags=0, series=False):
+    res = []
+
+    for i in range(len(s)):
+        expr = s.iloc[i]
+        # Case 1: expand=False and n_cols=1
+        if series:
+            if expr is not pd.NA and (match := pattern.search(expr)):
+                res.append(match.groups()[0])
+            else:
+                res.append(pd.NA)
+            continue
+        # Case 2: expand=True or n_cols>1
+        match_list = []
+        if expr is not pd.NA and (match := pattern.search(expr)):
+            match_list = list(match.groups())
+        match_list.extend([pd.NA] * (pattern.groups - len(match_list)))
+        res.append(match_list)
+    return pd.Series(res)
 
 
 class BodoStringMethods:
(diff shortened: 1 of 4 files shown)


bodo/pandas/utils.py
```diff
@@ -766,6 +766,7 @@ def _reconstruct_pandas_index(df, arrow_schema):
     for descr in arrow_schema.pandas_metadata.get("index_columns", []):
         if isinstance(descr, str):
             index_name = None if _is_generated_index_name(descr) else descr
+            # print(df, descr)
```
scott-routledge2: Remove
```suggestion
```

bodo/tests/test_df_lib/test_series_generator.py
```diff
@@ -128,6 +128,31 @@ def test_func(args, kwargs):
     }
 )
 
+extract_res_0 = pd.DataFrame(
```
scott-routledge2: Can we just use Pandas to check the results here?

bodo/tests/test_df_lib/test_series_str_methods.py
```diff
@@ -246,6 +257,8 @@ def _install_series_str_tests():
 
 df_decode = pd.DataFrame({"A": [b"hi", b"()", b"hello my name is chris.", None]})
 
+df_extract = pd.DataFrame({"A": ["a1", "b2", "c3"]}, index=["X", "Y", "Z"])
```
scott-routledge2: Add some na values?

bodo/pandas/series.py
```diff
@@ -909,7 +1019,29 @@ def zip_series_plan(lhs, rhs) -> BodoSeries:
     return result
 
 
-def _get_series_python_func_plan(series_proj, empty_data, func_name, args, kwargs):
+def create_expr(idx, empty_data, series_out, index_cols):
```
scott-routledge2: Use more descriptive name e.g. `get_col_as_series_expr`, "create_expr" is a little vague

bodo/pandas/series.py
```diff
@@ -696,6 +728,84 @@ def join_list(l):
...
+            (pattern,),
+            {"flags": flags, "series": is_series_output},
+            is_method=False,
+        )
+
+        # expand=False and n_cols=1: returns series
+        if is_series_output:
+            return series_out
+
+        n_index_arrays = get_n_index_arrays(series_out.index)
```
scott-routledge2: This will materialize the entire series, use `head(0)` or empty_data

bodo/pandas/series.py
```diff
@@ -982,37 +1114,27 @@ def partition(self, sep=" ", expand=True):
         if not expand:
             return series_out
 
+        n_index_arrays = get_n_index_arrays(series_out.index)
```
scott-routledge2: Same here

bodo/tests/test_df_lib/test_series_generator.py
chrisoh7: The column names are strings in bodo and ints in pandas, so _test_equal raises error for equivalent outputs.

scott-routledge2: Maybe leave a comment? In the future if this is a common issue we could add a flag to `_test_equal`

bodo/pandas/series.py
```diff
@@ -589,6 +589,38 @@ def sort_values(
...
+    series = pd.Series(
+        [
+            None if not isinstance(s.iloc[i], list) else s.iloc[i][col]
+            for i in range(len(s))
+        ]
+    )
+    return series
+
+
+def _str_extract_helper(s, pattern, flags=0, series=False):
```
scott-routledge2: Maybe a comment about how this does the same thing as str.extract except it returns matches in a list

scott-routledge2 (approved):
LGTM

bodo/tests/test_df_lib/test_series_generator.py
ehsantn: I agree, let's leave a comment at least.

bodo/pandas/series.py
```diff
@@ -982,37 +1117,30 @@ def partition(self, sep=" ", expand=True):
...
-        empty_series = pd.Series([], dtype=pd.ArrowDtype(pa.large_string()))
-
-        # Create scalar function expression for each column: extract value at index idx from each row
-        def create_expr(idx):
-            return LazyPlan(
-                "PythonScalarFuncExpression",
-                empty_series,
-                series_out._plan,
-                (
-                    "bodo.pandas.series._str_partition_helper",
```
ehsantn, line 999: Remove dead `_str_partition_helper` above.

bodo/pandas/series.py
```diff
@@ -749,6 +829,39 @@ def _str_cat_helper(df, sep, na_rep):
...
+        ]
+    )
+    return series
+
+
+def _str_extract_helper(s, pattern, flags=0, is_series=False):
+    """Performs row-wise pattern matching, returns a series of match lists."""
+    res = []
+
+    for i in range(len(s)):
```
ehsantn: Can we reuse Pandas instead of a loop like this that can be slow and potentially miss corner cases? Would `A.str.extract(patt, flags).apply(lambda a: a.tolist(), axis=1)` work (assuming performance is ok)?

chrisoh7: This works with patterns that have group names (i.e., pat = r'(?P<letter>[ab]'), but for those that don't, doesn't work. The reason for that is when I run s.str.extract() inside the helper, it's actually getting _str_extract() from ArrowExtensionArray class (.pixi/envs/default/lib/python3.13/site-packages/pandas/core/arrays/arrow/array.py) and has line that checks: 
```
groups = re.compile(pat).groupindex.keys()
if len(groups) == 0:
     raise ValueError(f"{pat=} must contain a symbolic group name.")
```
So when pat has no group names, what looks better:
1. cast type to object, do the job, and cast back?
2. use my old code
3. fallback to Pandas

chrisoh7: 1 would look roughly like this? 
```
def _str_extract_helper(s, pat, flags=0, is_series=False):
    """Performs row-wise pattern matching, returns a series of match lists."""
    res = []
    import re

    pattern = re.compile(pat, flags=flags)
    s = s.astype("string[python]")
    return s.str.extract(pat).apply(lambda a: a.tolist(), axis=1)
```

bodo/pandas/series.py
```diff
@@ -737,16 +834,25 @@ def _str_partition_helper(s, col):
     return series
 
 
-def _str_cat_helper(df, sep, na_rep):
-    """Concatenates df[idx] for idx in idx_pair, separated by sep."""
-    if sep is None:
-        sep = ""
+def _str_extract_helper(s, pat, expand, n_cols, flags):
```
chrisoh7, line 837: @ehsantn Is this better?

ehsantn: Yes, thanks.

PR #585, opened 2025-06-20
[BSE-4921] Groupby Agg Support

scott-routledge2:

Adds support for groupby and series agg by using _get_agg_plan helper function, which was expanded to generally handle cases like grouped.sum() but also grouped.agg("sum").

Functions added:
* count
* mean

There are many ways to [specify func for agg](https://pandas.pydata.org/pandas-docs/stable/user_guide/groupby.html#applying-multiple-functions-at-once), this implementation relies on pandas agg on a zero sized df for validation checks on things like column renaming, which can be messy. To create an agg plan, the func is normalized into a list of tuples (input column name, func) tuples in the order that they will appear in the output. There are additional validation checks performed during func normalizing/dtype fixing since there are some additional cases Bodo does not support yet. 

follow up PRs:
- Enable all agg funcs that bodo supports
- Fix multi index column names and aggregating on key columns
- Possibly porting more compiler tests to DataFrame library

## Testing strategy
New tests added, PR CI

## User facing changes
DataFrame/Series groupby().agg support plus mean and count.

## Checklist


bodo/pandas/_util.h
```diff
@@ -132,17 +132,19 @@ struct BodoPythonScalarFunctionData : public duckdb::FunctionData {
...
  */
 struct BodoAggFunctionData : public duckdb::FunctionData {
-    BodoAggFunctionData(bool dropna) : dropna(dropna) {}
+    BodoAggFunctionData(bool dropna, std::string name)
+        : dropna(dropna), name(name) {}
 
     bool Equals(const FunctionData &other_p) const override {
         const BodoAggFunctionData &other = other_p.Cast<BodoAggFunctionData>();
-        return (other.dropna == this->dropna);
+        return (other.dropna == this->dropna && other.name == this->name);
```
scott-routledge2, line 140: Agg funcs can have different names but still be considered equal in the optimizer, adding an extra field to data ensures different funcs are treated differently.

bodo/pandas/groupby.py
```diff
@@ -147,35 +224,70 @@ def __getattribute__(self, name: str, /) -> Any:
...
         grouped._selection[0]
         if isinstance(grouped, SeriesGroupBy)
         else grouped._selection
-    ].sum()
+    ].agg(func, *args, **kwargs)
 
-    empty_data = _cast_groupby_agg_columns("sum", empty_data_pandas, grouped._selection)
+    func = grouped._normalize_agg_func(func, kwargs)
+
+    n_key_cols = 0 if grouped._as_index else len(grouped._keys)
```
scott-routledge2, line 282: Technically this is not fully correct since you can aggregate over key columns. For example:
``` py
df1.groupby("C", as_index=False, dropna=True)[["C", "E", "D"]].agg("sum")
```
Wasn't sure how common this case was, but thought it could be addressed in followup.

bodo/pandas/groupby.py
```diff
@@ -220,41 +344,88 @@ def _get_agg_output_type(func: str, pa_type: pa.DataType, col_name: str) -> pa.D
...
+        new_type = _get_agg_output_type(
+            func, in_data.dtype.pyarrow_dtype, out_data.name
+        )
+        out_data = out_data.astype(pd.ArrowDtype(new_type))
+        return out_data
+
+    for i, (in_col_name, func_) in enumerate(func):
+        out_col_name = out_data.columns[i + n_key_cols]
+
+        if not isinstance(out_data[out_col_name], pd.Series):
```
scott-routledge2, line 425: I am not sure if this case is possible but just to be safe.

bodo/pandas/groupby.py
```diff
@@ -220,41 +344,88 @@ def _get_agg_output_type(func: str, pa_type: pa.DataType, col_name: str) -> pa.D
...
+            raise BodoLibNotImplementedException(
+                "GroupBy.agg(): detected duplicate output column name in output columns: ",
+                out_col_name,
+            )
+
+        in_col = in_data[in_col_name]
+        if not isinstance(out_data[out_col_name], pd.Series):
+            raise BodoLibNotImplementedException(
```
scott-routledge2, line 433: example: `bdf.groupby("C", as_index=True, dropna=True)[["A", "A"]].agg(["sum"])`

bodo/pandas/_util.h
```diff
@@ -132,17 +132,19 @@ struct BodoPythonScalarFunctionData : public duckdb::FunctionData {
...
     }
     duckdb::unique_ptr<duckdb::FunctionData> Copy() const override {
-        return duckdb::make_uniq<BodoAggFunctionData>(this->dropna);
+        return duckdb::make_uniq<BodoAggFunctionData>(this->dropna, this->name);
     }
 
     bool dropna;
+    std::string name;
```
DrTodd13: dropna and name can be const?

bodo/pandas/groupby.py
```diff
@@ -83,6 +84,36 @@ def __getattribute__(self, name: str, /) -> Any:
...
+            ]
+        elif is_dict_like(func):
+            normalized_func = [
+                (col, _get_aggfunc_str(func_)) for col, func_ in func.items()
+            ]
+        elif is_list_like(func):
+            normalized_func = [
+                (col, _get_aggfunc_str(func_))
+                for col in self._selection
+                for func_ in func
```
DrTodd13, line 118: Just want to make sure this part is tested.  Not sure I know the semantics well enough to say this is wrong but just want to make sure this shouldn't be "for col, func_ in zip(self._selection, func)".

scott-routledge2: In this case, it is more than just a zip since we want the cross product of columns, funcs.
For example, if the columns in selection were `["A", "B"]` and the funcs list was `["sum", "count"]`, the output columns would be: `("A", "sum"), ("A", "count"), ("B", "sum), ("B", "count")` as a MultiIndex.

bodo/pandas/_util.h
ehsantn: Let's add a comment in the code about this since very important.

bodo/pandas/groupby.py
```diff
@@ -83,6 +84,39 @@ def __getattribute__(self, name: str, /) -> Any:
...
+    agg = aggregate
+
+    def _normalize_agg_func(self, func, kwargs: dict) -> list[tuple[str, str]]:
+        """
+        Convert func and kwargs into a list of (column, function) tuples.
+        """
+        # list of (input column name, function) pairs
+        normalized_func: list[tuple[str, str]] = []
+
+        if func is None and kwargs:
```
ehsantn, line 100: Let's add comments with example codes for each case.

bodo/pandas/groupby.py
ehsantn: Let's add a comment with example code.

bodo/pandas/groupby.py
ehsantn: Let's add a comment in the code (and maybe some simple assertion).

bodo/pandas/groupby.py
ehsantn: Add comment to the code.

bodo/pandas/groupby.py
ehsantn: Add comment to the code.

bodo/pandas/groupby.py
```diff
@@ -220,41 +342,82 @@ def _get_agg_output_type(func: str, pa_type: pa.DataType, col_name: str) -> pa.D
...
+    for i, (in_col_name, func_) in enumerate(func):
+        out_col_name = out_data.columns[i + n_key_cols]
+
+        if not isinstance(out_data[out_col_name], pd.Series):
+            raise BodoLibNotImplementedException(
+                f"GroupBy.agg(): detected duplicate output column name in output columns: '{out_col_name}'"
+            )
+
+        in_col = in_data[in_col_name]
+        if not isinstance(out_data[out_col_name], pd.Series):
```
ehsantn: Duplicate check? Should be in_col?
```suggestion
        if not isinstance(in_col, pd.Series):
```

bodo/pandas/groupby.py
scott-routledge2: Added a fallback for now

bodo/pandas/groupby.py
scott-routledge2: Thanks for the catch! In the example I tested there were both duplicate input and output columns so it didn't actually matter

PR #588, opened 2025-06-20
TPCH related fixes.

DrTodd13:

1. Added dataframe library version of tpch to benchmarks/tpch directory.
2. Pass cardinality from Python to duckdb source operators.
3. Make sort arg supported by groupby but assert if it isn't false.
4. Use more direct method in setitem that avoids collections.
5. Print stack traceback cause of plan execution when dump plans option turned on.

## Testing strategy

run_ci

## User facing changes

None

## Checklist


benchmarks/tpch/dataframe_lib.py
```diff
@@ -0,0 +1,1061 @@
+# Original from https://gist.githubusercontent.com/UranusSeven/55817bf0f304cc24f5eb63b2f1c3e2cd/raw/796dbd2fce6441821fc0b5bc51491edb49639c55/tpch.py
```
ehsantn, line 1: Why not use translations from a well-known source like this to be more confident?
https://github.com/coiled/benchmarks/blob/main/tests/tpch/dask_queries.py

bodo/pandas/frame.py
```diff
@@ -866,6 +866,11 @@ def groupby(
         Provides support for groupby similar to Pandas:
         https://github.com/pandas-dev/pandas/blob/0691c5cf90477d3503834d983f69350f250a6ff7/pandas/core/frame.py#L9148
         """
+        if sort != False:
+            raise BodoLibNotImplementedException(
+                "groupby: only sort=False currently supported"
+            )
```
ehsantn: Requiring the user to always set sort=False to use Bodo is a significant burden I believe. This is a well-known issue of Pandas APIs when parallelized since users rarely require the sort but is expensive in parallel. As a pragmatic solution, we can just set `sort=False` by default same as JIT: https://docs.bodo.ai/latest/api_docs/pandas/dataframe/groupby/

bodo/pandas/frame.py
```diff
@@ -1033,7 +1039,12 @@ def _update_setitem_internal_state(
             # Update internal data manager (e.g. insert a new block or update an
             # existing one). See:
             # https://github.com/pandas-dev/pandas/blob/0691c5cf90477d3503834d983f69350f250a6ff7/pandas/core/frame.py#L4481
-            super().__setitem__(key, head_val)
+            if new_column:
+                self._mgr.insert(
+                    len(self._info_axis) - 1, key, *self._sanitize_column(head_val)
+                )
+            else:
+                super().__setitem__(key, head_val)
```
ehsantn, line 1042: What was the problem with `super().__setitem__` in this case? Do you have a reproducer? Using Pandas is desirable if possible to make sure all the internal details are set and we are more future proof.

bodo/pandas/plan_optimizer.pyx
```diff
@@ -714,10 +714,10 @@ cdef class LogicalGetParquetRead(LogicalOperator):
 
     def __cinit__(self, object out_schema, object parquet_path, object storage_options):
         self.out_schema = out_schema
-        cdef unique_ptr[CLogicalGet] c_logical_get = make_parquet_get_node(parquet_path, out_schema, storage_options)
-        self.c_logical_operator = unique_ptr[CLogicalOperator](<CLogicalGet*> c_logical_get.release())
         self.path = parquet_path
         self.storage_options = storage_options
+        cdef unique_ptr[CLogicalGet] c_logical_get = make_parquet_get_node(parquet_path, out_schema, storage_options, self.getCardinality())
```
ehsantn, line 719: Getting the exact number of rows during planning is probably too much overhead. Some approximation is probably enough. What do we do in BodoSQL @knassre-bodo ?

bodo/pandas/utils.py
```diff
@@ -465,6 +467,7 @@ def _exec_plan(plan):
             bodo.dataframe_library_dump_plans
             and bodo.libs.distributed_api.get_rank() == 0
         ):
+            traceback.print_stack(file=sys.stdout)
```
ehsantn, line 477: Let's add a comment about why this is useful since may not be obvious from the code.

ehsantn (changes requested):
I think there are fundamental issues here that need addressed. I'd be happy to chat.

bodo/pandas/frame.py
DrTodd13: __setitem__ has a check that requires length which causes a premature execution of the plan.

[Function that does len check.](https://github.com/pandas-dev/pandas/blob/0f437949513225922d851e9581723d82120684a6/pandas/core/frame.py#L4108)

benchmarks/tpch/dataframe_lib.py
DrTodd13: I spot checked query 5 and for sure I didn't check all of them but one thing suspect about both of our versions is the use of "TMP" column rather than something meaningful like in the one you posted.  However, I just looked a bit more and for query 2, the original bodo version and the one I checked in match but the one you referenced has an extra head(100) as the very last step.  I don't know the original source of the bodo one or the one I checked in but the formatting seems identical so I suspect they come from the same source.

bodo/pandas/frame.py
DrTodd13: Have we already accepted other deviations from Pandas standards for default values?  If so, I think this breaks the "one line change" mantra.  It is a small thing for sure but not something I agree with.  Nevertheless, I've reverted it based on your preference.

bodo/pandas/utils.py
DrTodd13: Done.

benchmarks/tpch/dataframe_lib.py
ehsantn: I don't know where our previous version comes from, but `head(100)` is not in the original SQL query. We will have to double check these before publishing benchmarks.
https://github.com/bodo-ai/Bodo/blob/b5031029f08a3df6c1313fe087bb7ebeb103f131/benchmarks/tpch/tpch.ipynb
https://github.com/bodo-ai/Bodo/blob/b5031029f08a3df6c1313fe087bb7ebeb103f131/BodoSQL/bodosql/tests/test_tpch_first_half.py#L107

bodo/pandas/frame.py
ehsantn: This is "lesser of two evils" I believe. I've never seen any Pandas code rely on this sorting behavior and the original reason was just nicer printing if I remember correctly (after talking to Pandas core developers).

bodo/pandas/frame.py
ehsantn: The `else` branch doesn't have this problem? Let's add a comment.

benchmarks/tpch/dataframe_lib.py
```diff
@@ -0,0 +1,1061 @@
...
+    print("Done loading orders")
+    return df
+
+
+@functools.lru_cache
+def load_customer(data_folder: str) -> pd.DataFrame:
+    print("Loading customer")
+    data_path = data_folder + "/customer.parquet"
+    df = pd.read_parquet(data_path)
+    print("Done oading customer")
```
IsaacWarren:
```suggestion
    print("Done loading customer")
```

IsaacWarren (approved):
Thanks @DrTodd13!

bodo/pandas/frame.py
DrTodd13: resolved

bodo/pandas/utils.py
DrTodd13: done.

bodo/pandas/frame.py
DrTodd13: Else may have the problem.  It is untested.  I'll try to add a test.

benchmarks/tpch/dataframe_lib.py
DrTodd13: Check all the queries.  Two versions now equivalent except bodo.jit version uses apply to get year whereas new version uses .dt.year.

bodo/pandas/frame.py
DrTodd13: There already were tests of setting existing columns and that code path returns before the length check in the nested function that was causing problem.  So, as far as I know, it is safe to call the regular routine for existing columns.

PR #589, opened 2025-06-22
Support Series.min/max

ehsantn:
## Changes included in this PR

As title. Uses Arrow compute in the backend.

## Testing strategy

Added unit test.

## User facing changes

Series.min/max now work in the dataframe library.

## Checklist
- [x] Pipelines passed before requesting review. To run CI you must include `[run CI]` in your commit message.
- [x] I am familiar with the [Contributing Guide](https://github.com/bodo-ai/Bodo/blob/main/CONTRIBUTING.md) 
- [x] I have installed + ran pre-commit hooks.

PR #592, opened 2025-06-24
Add decimal test to Series.min/max

ehsantn:
Small follow up on previous PR:
https://github.com/bodo-ai/Bodo/pull/589


scott-routledge2 (approved):
Thanks!

PR #593, opened 2025-06-24
Add groupby aggfuncs

scott-routledge2:

Adds all aggfuncs that are currently supported by the BodoSQL infrastructure and match Pandas semantics

Notable functions missing include:
all() (all elements in group are Truthy, no equivalent in SQL)
any() (at least one element in group is truthy, no equivalent in SQL)
cov() (requires two Series, which are aligned internally)
quantile (maybe could use percentile?)
sem (??)

## Testing strategy

New test added

## User facing changes
Added support for:
    "first",
    "idxmax",
    "idxmin",
    "last",
    "max",
    "median",
    "min",
    "prod",
    "size",
    "skew",
    "std",
    "var"

## Checklist


bodo/pandas/groupby.py
```diff
@@ -343,47 +533,91 @@ def _get_aggfunc_str(func):
...
+
+    Returns:
+        pa.DataType: The output type from applying func to col_name.
     """
-    new_type = pa_type
-    if func == "sum":
+    new_type = None
+    fallback = False
+
+    # TODO: Enable more fallbacks where the operation is supported in Pandas and not in Bodo
```
scott-routledge2, line 486: This function may be a little conservative but what we say we support is covered by tests. Some functions may raise a type error when there is a fallback available.

DrTodd13 (approved):
LGTM. Thanks.

bodo/pandas/groupby.py
```diff
@@ -119,7 +120,18 @@ def _normalize_agg_func(self, func, kwargs: dict) -> list[tuple[str, str]]:
             ]
         else:
             func = _get_aggfunc_str(func)
-            normalized_func = [(col, func) for col in self._selection]
+            # Size is a special case that only produces 1 column, since it doesn't
+            # depend on input column given.
+            if func == "size":
+                # Getting the size of each groups without any input column.
+                # e.g. df.groupby("B")[[]].size()
```
ehsantn, line 127: Isn't something like `df.groupby("B").size()` very common? Is `df.groupby("B")[[]].size()` different?

bodo/pandas/groupby.py
```diff
@@ -155,6 +167,97 @@ def count(self):
...
+        Compute the idxmin of each group.
+        """
+        return _groupby_agg_plan(self, "idxmin")
+
+    @check_args_fallback(supported="none")
+    def idxmax(self, axis=lib.no_default, skipna=True, numeric_only=False):
+        """
+        Compute the idxmax of each group.
+        """
+        return _groupby_agg_plan(self, "idxmax")
```
ehsantn: I thought first/last/idxmin/idxmax aren't supported yet?

bodo/pandas/groupby.py
```diff
@@ -155,6 +167,97 @@ def count(self):
...
+
+    @check_args_fallback(supported="none")
+    def nunique(self, dropna=True):
+        """
+        Compute the nunique of each group.
+        """
+        return _groupby_agg_plan(self, "nunique")
+
+    @check_args_fallback(supported="none")
+    def prod(self, numeric_only=False, min_count=0):
```
ehsantn: Same.

bodo/pandas/groupby.py
```diff
@@ -256,6 +323,129 @@ def _normalize_agg_func(self, func, kwargs):
...
+        Compute the idxmin of each group.
+        """
+        return _groupby_agg_plan(self, "idxmin")
+
+    @check_args_fallback(supported="none")
+    def idxmax(self, axis=lib.no_default, skipna=True):
+        """
+        Compute the idxmax of each group.
+        """
+        return _groupby_agg_plan(self, "idxmax")
```
ehsantn: Same.

bodo/pandas/groupby.py
```diff
@@ -343,47 +533,89 @@ def _get_aggfunc_str(func):
...
+    Returns:
+        pa.DataType: The output type from applying func to col_name.
     """
-    new_type = pa_type
-    if func == "sum":
+    new_type = None
+    fallback = False
+
+    # TODO: Enable more fallbacks where the operation is supported in Pandas and not in Bodo
+    if func in ("sum"):
```
ehsantn: Subtle tuple vs. scalar value bug:
```suggestion
    if func in ("sum",):
```

ehsantn (approved):
Thanks @scott-routledge2 . See comments on small issues.

bodo/pandas/groupby.py
scott-routledge2: Under the hood `df.groupby("B")` initializes the `_selection` to be all the columns that were not keys (to match pandas behavior), so this should work in the common case.

bodo/pandas/groupby.py
scott-routledge2: Oh I removed them from other places but must've forgot this part, thanks!

bodo/pandas/groupby.py
scott-routledge2: Thanks!

bodo/pandas/groupby.py
ehsantn: Ok, thanks.

PR #601, opened 2025-06-25
Basic concat support.

DrTodd13:

Basic concat support.

## Testing strategy

run_ci

## User facing changes

concat with no optional arguments works.

## Checklist


bodo/tests/test_lazy/test_bodo_series.py
```diff
@@ -281,6 +281,7 @@ def test_slice(single_pandas_managers, head_s, collect_func, del_func):
     assert lam_s._lazy
     assert lam_sliced_head_s.tolist() == (head_s[1:3]).tolist()
 
+    """
```
IsaacWarren: Is this broken now?

bodo/pandas/plan_optimizer.pyx
```diff
@@ -413,6 +419,20 @@ cdef class LogicalComparisonJoin(LogicalOperator):
         return f"LogicalComparisonJoin({join_type})"
 
 
+cdef class LogicalSetOperation(LogicalOperator):
+    """Wrapper around DuckDB's LogicalSetOperation to provide access in Python.
+    """
+
+    def __cinit__(self, out_schema, LogicalOperator lhs, LogicalOperator rhs, str setop):
```
IsaacWarren, line 426: Maybe include a comment for possible values of setop?

bodo/pandas/_physical_conv.cpp
```diff
@@ -174,6 +175,71 @@ void PhysicalPlanBuilder::Visit(duckdb::LogicalComparisonJoin& op) {
     this->active_pipeline->AddOperator(physical_join);
 }
 
+bool arrowSchemaTypeEquals(const ::arrow::Schema& s1,
```
IsaacWarren, line 185: Why can't we use schema.Equals in this case?

bodo/pandas/_physical_conv.cpp
```diff
@@ -174,6 +175,71 @@ void PhysicalPlanBuilder::Visit(duckdb::LogicalComparisonJoin& op) {
...
+        }
+    }
+    return true;
+}
+
+void PhysicalPlanBuilder::Visit(duckdb::LogicalSetOperation& op) {
+    if (op.type == duckdb::LogicalOperatorType::LOGICAL_UNION) {
+        // UNION ALL
+        if (op.setop_all) {
+            // Rightt-child will feed into a table.
```
IsaacWarren:
```suggestion
            // Right-child will feed into a table.
```

bodo/pandas/_physical_conv.cpp
```diff
@@ -174,6 +175,71 @@ void PhysicalPlanBuilder::Visit(duckdb::LogicalComparisonJoin& op) {
...
+                                            build_pipelines.end());
+
+            // Left-child will feed into the same table.
+            this->Visit(*op.children[1]);
+            std::shared_ptr<bodo::Schema> lhs_table_schema =
+                this->active_pipeline->getPrevOpOutputSchema();
+            ::arrow::Schema lhs_arrow = *(lhs_table_schema->ToArrowSchema());
+            if (!arrowSchemaTypeEquals(rhs_arrow, lhs_arrow)) {
+                std::cout << "lhs " << lhs_arrow.ToString() << std::endl;
+                std::cout << "rhs " << rhs_arrow.ToString() << std::endl;
```
IsaacWarren: Move this to the runtime error message?

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -1739,3 +1739,41 @@ def test_read_csv(datapath):
         bodo_out,
         py_out,
     )
+
+
```
IsaacWarren: Can we add a fallback test for non-all union?

bodo/pandas/_util.cpp
```diff
@@ -328,6 +375,32 @@ std::shared_ptr<arrow::Scalar> convertDuckdbValueToArrowScalar(
     return scalar_res.ValueOrDie();
 }
 
+std::shared_ptr<arrow::DataType> duckdbValueToArrowType(
```
IsaacWarren, line 379: Can't we just call convertDuckDbValueToArrowScalar and get the type from the scalar?

bodo/pandas/_util.cpp
```diff
@@ -60,6 +60,53 @@ extractValue(const duckdb::Value &value) {
     }
 }
 
+std::variant<int8_t, int16_t, int32_t, int64_t, uint8_t, uint16_t, uint32_t,
+             uint64_t, bool, std::string, float, double,
+             std::shared_ptr<arrow::Scalar>>
+getNullValue(const duckdb::Value &value) {
```
IsaacWarren: IMO the name of this doesn't make sense, int8_t{} isn't a null value, it's the default value

bodo/pandas/base.py
```diff
@@ -421,6 +425,102 @@ def to_datetime(
...
+    sort: bool = False,
+    copy: bool | None = None,
+) -> BodoDataFrame | BodoSeries:
+    if isinstance(objs, Mapping):
+        raise BodoLibNotImplementedException(
+            "concat does not current support objs of Mapping type"
+        )
+
+    if len(objs) == 0:
+        raise ValueError("No objects to concatenate")
```
IsaacWarren, line 470: Is this the same error pandas raises in this case?

bodo/pandas/frame.py
```diff
@@ -1028,6 +1031,16 @@ def __setitem__(self, key, value) -> None:
             self._update_setitem_internal_state(new_plan, key, value)
             return
 
+        if (
```
IsaacWarren, line 1088: I think a comment explaining what this is doing would be helpful

bodo/pandas/physical/expression.h
```diff
@@ -230,6 +230,89 @@ std::shared_ptr<arrow::Array> ScalarToArrowArray(
...
+    using BuilderType = arrow::NumericBuilder<ArrowType>;
+
+    BuilderType builder;
+    arrow::Status status;
+    for (size_t i = 0; i < num_elements; ++i) {
+        status = builder.AppendNull();
+        if (!status.ok()) {
+            throw std::runtime_error("builder.Append failed.");
+        }
+    }
```
IsaacWarren:
```suggestion
      status = builder.AppendNulls(num_elements);
      if (!status.ok()) {
           throw std::runtime_error("builder.Append failed.");
      }
```

bodo/pandas/physical/expression.h
```diff
@@ -230,6 +230,89 @@ std::shared_ptr<arrow::Array> ScalarToArrowArray(
...
+
+    friend std::ostream &operator<<(std::ostream &os,
+                                    const PhysicalNullExpression<T> &obj) {
+        os << "PhysicalNullExpression " << obj.constant << std::endl;
+        return os;
+    }
+
+   private:
+    const T constant;  // holds no real value, only for type
```
IsaacWarren, line 310: Use a typedef instead?

bodo/pandas/physical/project.h
```diff
@@ -29,7 +29,8 @@ class PhysicalProjection : public PhysicalSourceSink {
 
         // Create the output schema from expressions
         this->output_schema = std::make_shared<bodo::Schema>();
-        for (auto& expr : exprs) {
+        for (size_t i = 0; i < exprs.size(); ++i) {
+            auto& expr = exprs[i];
```
IsaacWarren: What is this change for?

bodo/pandas/physical/expression.cpp
```diff
@@ -62,6 +62,58 @@ std::shared_ptr<arrow::Array> ScalarToArrowArray(bool value,
     return array;
 }
 
+// String specialization
+std::shared_ptr<arrow::Array> NullArrowArray(const std::string& value,
+                                             size_t num_elements) {
+    arrow::StringBuilder builder;
+    arrow::Status status;
+    for (size_t i = 0; i < num_elements; ++i) {
+        status = builder.AppendNull();
```
IsaacWarren: Same as in the header, use AppendNulls instead of a loop

bodo/pandas/physical/expression.cpp
```diff
@@ -62,6 +62,58 @@ std::shared_ptr<arrow::Array> ScalarToArrowArray(bool value,
...
+    return array_result.ValueOrDie();
+}
+
+std::shared_ptr<arrow::Array> NullArrowArray(bool value, size_t num_elements) {
+    arrow::BooleanBuilder builder;
+    arrow::Status status;
+
+    for (size_t i = 0; i < num_elements; ++i) {
+        // Append boolean value
+        status = builder.AppendNull();
```
IsaacWarren: AppendNulls here too

bodo/pandas/physical/union_all.h
```diff
@@ -0,0 +1,80 @@
...
+   public:
+    explicit PhysicalUnionAll(std::shared_ptr<bodo::Schema> input_schema)
+        : output_schema(input_schema) {}
+
+    virtual ~PhysicalUnionAll() = default;
+
+    void Finalize() override {}
+
+    /**
+     * @brief Do limit.
```
IsaacWarren: Update docstring

bodo/pandas/physical/union_all.h
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    /**
+     * @brief GetResult - just for API compatability but should never be called
+     */
+    std::variant<std::shared_ptr<table_info>, PyObject*> GetResult() override {
+        // Limit should be between pipelines and act alternatively as a sink
+        // then source but there should never be the need to ask for the result
+        // all in one go.
+        throw std::runtime_error("GetResult called on a limit node.");
```
IsaacWarren:
```suggestion
        throw std::runtime_error("GetResult called on a union node.");
```

bodo/pandas/physical/union_all.h
```diff
@@ -0,0 +1,80 @@
...
+        return (prev_op_result == OperatorResult::FINISHED)
+                   ? OperatorResult::FINISHED
+                   : OperatorResult::NEED_MORE_INPUT;
+    }
+
+    /**
+     * @brief GetResult - just for API compatability but should never be called
+     */
+    std::variant<std::shared_ptr<table_info>, PyObject*> GetResult() override {
+        // Limit should be between pipelines and act alternatively as a sink
```
IsaacWarren:
```suggestion
        // Union should be between pipelines and act alternatively as a sink
```

bodo/pandas/plan_optimizer.pyx
DrTodd13: done.

bodo/pandas/_physical_conv.cpp
DrTodd13: That checks column names too right?  I don't think column names were matching here for some reason and my recollection it wasn't an easy fix.

bodo/pandas/_physical_conv.cpp
DrTodd13: done

bodo/pandas/_physical_conv.cpp
DrTodd13: done

bodo/pandas/_physical_conv.cpp
IsaacWarren: check_metadata is false by default but maybe column names don't count as metadata

bodo/tests/test_df_lib/test_end_to_end.py
DrTodd13: Will handle in dropduplicates PR coming up.

bodo/pandas/_util.cpp
DrTodd13: You can't use that function if the duckdb value is NULL.

IsaacWarren: Can we add support for nulls to that function?

bodo/pandas/_util.cpp
DrTodd13: Done.

bodo/pandas/base.py
DrTodd13: Yes.

bodo/pandas/frame.py
DrTodd13: Done.

bodo/pandas/physical/expression.h
DrTodd13: Done.

bodo/pandas/physical/project.h
DrTodd13: That was part of a failed experiment that I removed.  I reverted now.

bodo/pandas/physical/expression.cpp
DrTodd13: Done.

bodo/pandas/physical/expression.cpp
DrTodd13: Done.

bodo/pandas/physical/union_all.h
DrTodd13: Done.

bodo/pandas/physical/union_all.h
DrTodd13: Done.

bodo/pandas/physical/union_all.h
DrTodd13: Done.

bodo/pandas/physical/expression.h
DrTodd13: You can almost do this purely type based except for the arrow::Scalar case where you need the type inside the arrow::Scalar.  Unfortunately, that means none of the cases can be purely type based.

bodo/tests/test_lazy/test_bodo_series.py
ehsantn: Same question.

bodo/pandas/base.py
```diff
@@ -421,6 +425,102 @@ def to_datetime(
...
+            ignore_index=ignore_index,
+            keys=keys,
+            levels=levels,
+            names=names,
+            sort=sort,
+            copy=copy,
+        )
+
+        orig_a = py_copy.deepcopy(a._plan)
+        orig_b = py_copy.deepcopy(b._plan)
```
ehsantn: This should use the approach here: https://github.com/bodo-ai/Bodo/pull/630

bodo/tests/test_lazy/test_bodo_series.py
DrTodd13: Don't remember doing this but it seems to work so reverted back.

bodo/pandas/base.py
```diff
@@ -443,6 +447,102 @@ def to_datetime(
     )
 
 
+@check_args_fallback("none")
```
ehsantn: I assume we don't support all arguments. Let's check for supported arguments properly.

bodo/pandas/_plan.cpp
```diff
@@ -616,6 +651,7 @@ std::pair<int64_t, PyObject *> execute_plan(
     PyObject *pyarrow_schema =
         arrow::py::wrap_schema(output_table->schema()->ToArrowSchema());
 
+    // Clear out the table_index to schema mapping for the next plan.
```
ehsantn: Outdated comment?

bodo/pandas/_plan.cpp
```diff
@@ -77,6 +77,15 @@ duckdb::unique_ptr<duckdb::LogicalOperator> optimize_plan(
     return out_plan;
 }
 
+duckdb::unique_ptr<duckdb::Expression> make_const_null(PyObject *out_schema_py,
+                                                       int64_t field_idx) {
+    std::shared_ptr<arrow::Schema> arrow_schema = unwrap_schema(out_schema_py);
+    const std::shared_ptr<arrow::Field> &field = arrow_schema->field(field_idx);
+    auto [_, out_type] = arrow_field_to_duckdb(field);
+    return duckdb::make_uniq<duckdb::BoundConstantExpression>(
+        duckdb::Value(out_type));
```
ehsantn, line 88: Kind of unintuitive that `duckdb::Value` sets the is_null flag when the constructor argument is a data type. Let's add a comment.

bodo/pandas/_physical_conv.cpp
ehsantn: Let's add a comment about this function exists because we don't want to check column names.

bodo/pandas/_physical_conv.cpp
```diff
@@ -174,6 +175,70 @@ void PhysicalPlanBuilder::Visit(duckdb::LogicalComparisonJoin& op) {
...
+    return true;
+}
+
+void PhysicalPlanBuilder::Visit(duckdb::LogicalSetOperation& op) {
+    if (op.type == duckdb::LogicalOperatorType::LOGICAL_UNION) {
+        // UNION ALL
+        if (op.setop_all) {
+            // Right-child will feed into a table.
+            PhysicalPlanBuilder rhs_builder;
+            rhs_builder.Visit(*op.children[0]);
```
ehsantn: Make `children[1]` the build side to be consistent with join?
```suggestion
            rhs_builder.Visit(*op.children[1]);
```

bodo/pandas/_physical_conv.cpp
```diff
@@ -174,6 +175,70 @@ void PhysicalPlanBuilder::Visit(duckdb::LogicalComparisonJoin& op) {
...
+                    "PhysicalPlanBuilder::Visit(LogicalSetOperation lhs and "
+                    "rhs schemas not identical. " +
+                    lhs_arrow.ToString() + " versus " + rhs_arrow.ToString());
+            }
+
+            finished_pipelines.emplace_back(
+                this->active_pipeline->Build(physical_union_all));
+            // Downstream pipelines pull from the singular table.
+            this->active_pipeline =
+                std::make_shared<PipelineBuilder>(physical_union_all);
```
ehsantn: Why is this different from join pipelines? This doesn't look correct to me.

bodo/pandas/physical/union_all.h
```diff
@@ -0,0 +1,80 @@
...
+     *
+     * @return std::pair<std::shared_ptr<table_info>, OperatorResult>
+     * The output table from the current operation and whether there is more
+     * output.
+     */
+    OperatorResult ConsumeBatch(std::shared_ptr<table_info> input_batch,
+                                OperatorResult prev_op_result) override {
+        if (!collected_rows) {
+            collected_rows = std::make_unique<ChunkedTableBuilderState>(
+                input_batch->schema(), input_batch->nrows());
```
ehsantn: `input_batch` may have a different size for whatever reason (filters etc).
```suggestion
                input_batch->schema(), get_streaming_batch_size());
```

bodo/pandas/physical/union_all.h
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * @brief ProduceBatch - act as a data source
+     *
+     * returns std::pair<std::shared_ptr<table_info>, OperatorResult>
+     */
+    std::pair<std::shared_ptr<table_info>, OperatorResult> ProduceBatch()
+        override {
+        auto next_batch = collected_rows->builder->PopChunk(true);
+        std::shared_ptr<table_info> out_batch = std::get<0>(next_batch);
```
ehsantn: Minor: `out_batch ` unused.

bodo/pandas/_physical_conv.cpp
DrTodd13: Done.

bodo/pandas/base.py
DrTodd13: Done.

bodo/pandas/_plan.cpp
DrTodd13: Did have some code there but removed it but forgot to remove corresponding comment.

bodo/pandas/_plan.cpp
DrTodd13: Done.

bodo/pandas/_physical_conv.cpp
DrTodd13: If you concat(A, B) then A becomes children[0] and B becomes children[1].  I think we should maintain the normal order pandas users expect of A followed by B which means that A has to go first which means that A has to go in the rhs_builder since rhs_builder is added to the completed pipeline list first and will therefore be run first.

bodo/pandas/_physical_conv.cpp
DrTodd13: There's two ways to do this.  First, add dfA to table builder in first pipeline, add dfB to same table builder in second pipeline, pop from table builder in third pipeline.  Second, add dfA to table builder in first pipeline, first pop all entries from table builder upon receiving first batch from dfB before then passing batches dfB straight through without adding to table builder.  Approach 2 is slightly more complicated so I went with the simpler approach to start with.  With approach 2 there's also the complication of knowing how to know when you've switched to the second pipeline but I assume that is solvable.  Let me know if you really want to switch approaches.

bodo/pandas/physical/union_all.h
DrTodd13: Fixed.

bodo/pandas/_physical_conv.cpp
ehsantn: We can maintain the SQL convention by building the plan with (B, A) order on the frontend side. But not a big deal.

bodo/pandas/base.py
DrTodd13: Done.

bodo/pandas/_physical_conv.cpp
ehsantn: I think following SQL convention can be important so the optimizer switches sides properly to avoid unnecessary spilling (move smaller table to build side, stream the larger table).

bodo/pandas/_physical_conv.cpp
ehsantn: Approach 1 breaks streaming for the "probe" side by making it "build", leading to unnecessary spilling for large tables. For approach 2, the overhead of appending to buffer and returning the same data (extra copies) is probably a lot less than spilling to disk. I'd be happy to discuss this.

bodo/pandas/_physical_conv.cpp
DrTodd13: @ehsantn Is this fundamentally compatible with pandas if pandas orders based on what order you give to concat whereas dukdb orders based on table size?

ehsantn: Good point. Is it big deal if we change the order and not match Pandas here? Doesn't sound important to me.

bodo/pandas/base.py
```diff
@@ -443,6 +446,100 @@ def to_datetime(
     )
 
 
+@check_args_fallback(unsupported="all")
```
ehsantn, line 449: `unsupported="all"` ignores the first argument?

bodo/pandas/physical/union_all.h
```diff
@@ -0,0 +1,96 @@
...
+     *
+     * returns std::pair<std::shared_ptr<table_info>, OperatorResult>
+     */
+    std::pair<std::shared_ptr<table_info>, OperatorResult> ProcessBatch(
+        std::shared_ptr<table_info> input_batch,
+        OperatorResult prev_op_result) override {
+        if (!first_processed_batch) {
+            first_processed_batch = input_batch;
+        }
+        if (collected_rows && !collected_rows->builder->empty()) {
```
ehsantn, line 64: Doesn't this basically discard probe side batches (except the first) until the build batches are consumed?

ehsantn (commented):
See comment.

bodo/pandas/base.py
DrTodd13: check_args_fallback only applies to arguments with default values.

bodo/pandas/physical/union_all.h
DrTodd13: I don't think so.  It records the first one and returns have_more_output and subsequent input_batches will be empty so we need to remember the first one.  Once the build batches are empty then we process the original first probe batch we got.  After that, I suppose it saves each input batch and restores immediately.

ehsantn: Makes sense. It just looks a bit scary to rely on `HAVE_MORE_OUTPUT` for correctness but should be fine if it works in all cases.

PR #602, opened 2025-06-25
Iceberg Table Cardinality Estimate

IsaacWarren:

Adds a method to get the length of a table from a variety of metadata and passes the length to LogicalGet for read iceberg nodes in the df lib.

## Testing strategy
Added test for method and ran it locally since Iceberg tests are disabled on CI.

## User facing changes
Better plans in the dataframe library with Iceberg reads.

## Checklist


bodo/pandas/base.py
```diff
@@ -184,6 +185,17 @@ def read_iceberg(
...
+    # Get the table length estimate, if there's not a filter it will be exact
+    table_len_estimate = get_table_length(table, snapshot_id or -1)
+
+    # If there's a row filter, we need to estimate the selectivity
+    # and adjust the table length estimate accordingly.
+    if row_filter is not None and table_len_estimate > 0:
+        # TODO: do something smarter here like sampling or turn the filter into a
+        # separate node so the planner can handle it
+        filter_selectivity_estimate = 0.1
+        table_len_estimate = int(table_len_estimate * filter_selectivity_estimate)
```
scott-routledge2, line 199: What do we do in BodoSQL for estimating selectivity?

IsaacWarren: I'm pretty sure calcite does the same thing and just multiplies by a constant, (I thought I remembered it being 0.1)

IsaacWarren: getSelectivity in BodoRelMdSelectivity.kt has the rules

ehsantn: Same question. But ok to skip if too hard to find out since `row_filter` is optional.

ehsantn (approved):
Looks good to me. Thanks @IsaacWarren .

bodo/pandas/base.py
IsaacWarren: Yeah I think duckdb and calcite look at the filter type and change the estimate based on the filter type, ideally we'd just turn it into its own filter node so we don't have to worry about it. DuckDB's default filter selectivity is 0.2, I'll change to match

PR #618, opened 2025-06-30
Avoid Arrow data types for UDF calls

ehsantn:

This makes the example code below with df.apply >4x faster on my local M1 Mac. Works around a Pandas performance issue when using Arrow dtypes in UDFs:
https://github.com/pandas-dev/pandas/issues/61747
Also tested Pandas main branch and the issue persists so the Pandas 3 release won't solve it.

```python
import bodo.pandas as pd
import numpy as np
import time

NUM_ROWS = 500_000
df = pd.DataFrame({
    "A": np.arange(NUM_ROWS) % 30,
    "B": np.arange(NUM_ROWS)+1.0
})

t0 = time.time()
df["C"] = df.apply(lambda r: 0 if r.A == 0 else (r.B // r.A), axis=1)
print(df)
t1 = time.time()
print(f"Time taken: {t1 - t0:.2f} seconds")
```

## Testing strategy

Measured performance locally (obvious).

## Checklist

### DIFF
--- a/bodo/pandas/utils.py
+++ b/bodo/pandas/utils.py
@@ -57,12 +57,12 @@ def get_lazy_single_manager_class() -> type[
     )
 
 
-def cpp_table_to_df(cpp_table, arrow_schema=None):
+def cpp_table_to_df(cpp_table, arrow_schema=None, use_arrow_dtypes=True):
     """Convert a C++ table (table_info) to a pandas DataFrame."""
     from bodo.ext import plan_optimizer
 
     arrow_table = plan_optimizer.cpp_table_to_arrow(cpp_table)
-    df = arrow_table_to_pandas(arrow_table, arrow_schema)
+    df = arrow_table_to_pandas(arrow_table, arrow_schema, use_arrow_dtypes)
     return df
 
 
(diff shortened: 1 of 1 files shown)


bodo/pandas/utils.py
```diff
@@ -614,9 +614,14 @@ def run_func_on_table(cpp_table, result_type, in_args):
     """Run a user-defined function (UDF) on a DataFrame created from C++ table and
     return the result as a C++ table and column names.
     """
-    input = cpp_table_to_df(cpp_table)
     func, is_series, is_attr, args, kwargs = in_args
 
+    # Arrow dtypes can be very slow for UDFs in Pandas:
+    # https://github.com/pandas-dev/pandas/issues/61747
+    # TODO[BSE-4948]: Use Arrow dtypes when Bodo engine is specified
+    use_arrow_dtypes = not (is_attr and func in ("apply", "str"))
```
scott-routledge2: Does map also have the same performance issue?

scott-routledge2 (approved):
LGTM, Thanks @ehsantn

bodo/pandas/utils.py
ehsantn: This was a typo. Thanks for the catch.

ehsantn: Actually, left map out for now since I can't reproduce the problem with map right now and apply's problem may be due to packaging the row into a Series which map doesn't do.

ehsantn:
Looks like there are CI errors in _dt_month_name that don't seem relevant to this PR.
https://github.com/bodo-ai/Bodo/runs/45093447711

PR #622, opened 2025-07-01
Support na_action and kws in Series.map()

ehsantn:

Adds support for na_action and kws in Series.map JIT to enhance Pandas UDF support.

## Testing strategy

Added unit test. Tested Pandas 3 changes locally since not released yet.

## User facing changes

`na_action` and kws will work in Series.map with Bodo engine when Pandas 3 is out.

## Checklist

### DIFF
--- a/bodo/tests/test_series_part1.py
+++ b/bodo/tests/test_series_part1.py
@@ -2859,7 +2859,6 @@ def test_series_map_args(memory_leak_check):
 
 
 @pytest.mark.slow
-@pytest.mark.df_lib
 def test_series_map_na_action_kws(memory_leak_check):
     """Test Series.map with na_action and kw arguments"""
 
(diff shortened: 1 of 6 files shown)


DrTodd13 (approved):
LGTM.  Thanks.

bodo/tests/test_series_part1.py
```diff
@@ -2863,6 +2858,19 @@ def test_wrong_func(S):
         bodo.jit(test_wrong_func)(S)
 
 
+@pytest.mark.slow
+@pytest.mark.df_lib
```
scott-routledge2: Wouldn't we have to change the map frontend code in order for this test to pass in df_lib?

scott-routledge2 (approved):
LGTM! Although we should probably remove the df library test until Pandas 3.0 is out?

bodo/tests/test_series_part1.py
ehsantn: Good catch. Done.

PR #631, opened 2025-07-03
[BSE-4968] Support Series.mean

chrisoh7:

Adds support for Series.mean. 
Note: mean couldn't be plugged into `_compute_series_reduce` like the other reduce methods because parallel mean calculation `_compute_series_reduce("mean")` yields incorrect values. 
Below is comparison of running s.mean() with pandas and bodo.
```
s = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
Name: A, dtype: int64
pandas: 4.5 
bodo: 5.25

s = rev([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
Name: B, dtype: int64
pandas: 4.5 
bodo: 3.75
```
Workaround was computing reduce with sum and count, and calculating sum(s)/count(s), since `_compute_series_reduce` materializes the plan anyways. Null array check prevents div error.

Changes implementation of `PhysicalReduce` to take in a list of function names, to support evaluating multiple reductions on single run. 
## Testing strategy

PR CI, adds end-to-end test case.

## User facing changes

Series.mean().

## Checklist

### DIFF
--- a/bodo/pandas/_plan.cpp
+++ b/bodo/pandas/_plan.cpp
@@ -138,7 +138,6 @@ duckdb::unique_ptr<duckdb::Expression> make_agg_expr(
 
     // Get DuckDB output type
     auto field = out_schema->field(0);
-    // printf("%s\n", out_schema->ToString().c_str());
     auto [_, out_type] = arrow_field_to_duckdb(field);
 
     // Get arguments and their types for the aggregate function.
--- a/bodo/pandas/series.py
+++ b/bodo/pandas/series.py
@@ -1530,10 +1530,7 @@ def _split_internal(self, name, pat, n, expand, regex=None):
         is_method=False,
     )
 
-    # TODO: Implement Series.max()
     n_cols = length_series.max()
-
-    print(n_cols)
 
     n_index_arrays = get_n_index_arrays(index)
     index_cols = tuple(range(1, 1 + n_index_arrays))


bodo/pandas/series.py
```diff
@@ -673,6 +673,12 @@ def product(
     def count(self):
         return _compute_series_reduce(self, "count")
 
+    @check_args_fallback(unsupported="all")
+    def mean(self, axis=0, skipna=True, numeric_only=False, **kwargs):
+        if (n := _compute_series_reduce(self, "count")) <= 0:
```
chrisoh7: Feel free to let me know if this is a good way to tackle this or there are better ways.

chrisoh7: One thing I wanted to do was to compute reduce for sum and count at the same time, using something like `_compute_series_reduce(self, ["sum", "count"])`. @ehsantn does the parallel reduce infra have UDF registering functionalities or anything that could support these?

DrTodd13: Yes, you should probably extend compute_series_reduce to take one func or a list of funcs and that way there is only one LogicalAggregate and only one pass over the data.  Then you can do the divide as you are doing on the two results returned by that function.

bodo/pandas/series.py
```diff
@@ -1098,8 +1104,8 @@ def _compute_series_reduce(bodo_series: BodoSeries, func_name: str):
 
     from bodo.pandas.base import _empty_like
 
-    # TODO: support other functions like mean, etc.
-    assert func_name in ("min", "max", "sum", "product", "count"), (
+    # TODO: expand coverage for other reduction functions
+    assert func_name in ("min", "max", "sum", "product", "count", "mean"), (
```
DrTodd13: Why mean here?  I think you are correctly doing this as a sum/count so mean never gets to this point?

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -1698,6 +1698,12 @@ def test_series_sum_product_count():
         assert np.isclose(bdf[c].sum(), df[c].sum(), rtol=1e-6)
         assert np.isclose(bdf[c].product(), df[c].product(), rtol=1e-6)
         assert bdf[c].count() == df[c].count()
+        out_pandas, out_bodo = df[c].mean(), bdf[c].mean()
+        assert (
+            pd.isna(out_pandas)
```
DrTodd13: Switch the order here to put the common case np.isclose as the primary value and pd.isna(out_pandas) as the fallback?

DrTodd13 (changes requested):
It is nice the infrastructure is closer for this to be working than I thought but just one main suggestion.  Thanks.

bodo/pandas/series.py
ehsantn: I agree. I think it's worth it to support multiple reduction functions so there is only a single pass over the data. The backend needs a bit of work but shouldn't be too difficult. For example here:
https://github.com/bodo-ai/Bodo/blob/e97341bd2c9be8e05707e0b111b770324e53f5f5/bodo/pandas/_physical_conv.cpp#L82

ehsantn (commented):
See comment.

bodo/pandas/series.py
chrisoh7: Thanks, sounds good. I'll create a separate ticket on that assigned to myself and work on it shortly.

bodo/pandas/series.py
chrisoh7: You're right, I missed that one.

bodo/pandas/_physical_conv.cpp
```diff
@@ -79,17 +80,26 @@ void PhysicalPlanBuilder::Visit(duckdb::LogicalAggregate& op) {
...
+                    "LogicalAggregate with no groups must have BOUND_AGGREGATE "
+                    "expression types for reduction.");
+            }
         }
-        auto& agg_expr =
-            op.expressions[0]->Cast<duckdb::BoundAggregateExpression>();
-
-        if (agg_expr.function.name == "count_star") {
+        std::vector<std::string> function_names;
+        std::vector<duckdb::BoundAggregateExpression*> agg_expressions;
```
DrTodd13: Why a pointer here instead of a memory managed version like unique_ptr?

bodo/pandas/_physical_conv.cpp
```diff
@@ -103,15 +113,20 @@ void PhysicalPlanBuilder::Visit(duckdb::LogicalAggregate& op) {
...
+        auto bodo_schema = std::make_shared<bodo::Schema>();
         BodoAggFunctionData& bind_info =
-            agg_expr.bind_info->Cast<BodoAggFunctionData>();
-
-        auto out_schema = bind_info.out_schema;
-        auto bodo_schema = bodo::Schema::FromArrowSchema(out_schema);
-
-        auto physical_op = std::make_shared<PhysicalReduce>(
-            bodo_schema, agg_expr.function.name);
+            agg_expressions[0]->bind_info->Cast<BodoAggFunctionData>();
```
DrTodd13: Maybe a comment here why you only have to look at agg_expression[0] to get the whole output schema.

bodo/pandas/physical/reduce.h
```diff
@@ -48,36 +48,40 @@ class PhysicalReduce : public PhysicalSource, public PhysicalSink {
...
+                arrow::compute::CallFunction(function_name, {in_arrow_array});
+            CHECK_ARROW(cmp_res.status(), "Error in Arrow compute kernel");
+            std::shared_ptr<arrow::Scalar> out_scalar_batch =
+                cmp_res.ValueOrDie().scalar();
+            output_scalar_batches.push_back(out_scalar_batch);
+            ReductionType reduction_type = getReductionType(function_name);
+
+            // Update reduction result
+            if (iter == 0) {
+                output_scalars = output_scalar_batches;
```
scott-routledge2: I think it would be a bit simpler to push back here instead:
```suggestion
                output_scalars.push_back(out_scalar_batch);
```

bodo/pandas/series.py
```diff
@@ -1072,51 +1094,56 @@ def validate_reduce(func_name, pa_type):
...
-            return 0
-        if func_name == "product":
-            return 1
+        return generate_null_reduce(func_names)
 
-    # Check for supported types
-    if output_type := validate_reduce(func_name, pa_type):
-        zero_size_self = zero_size_self.astype(output_type)
+    new_arrow_schema = map_validate_reduce(func_names, pa_type)
+    zero_size_self = arrow_to_empty_df(new_arrow_schema)
```
scott-routledge2, line 1137: technically the `return_type` of `duckdb::AggregateFunction` would be different if passing the entire `zero_size_self` for each expression.

For example mean on a float column, the first agg expression would have output type float (since it's a sum), but then the second output type would also have type float (since we only look at the first type in the schema see _plan.cpp:140). 

I don't think this does anything right now since we don't use this value, @DrTodd13 do you know how it's used in duckdb?

scott-routledge2 (approved):
LGTM, thanks @chrisoh7! See my comments.

bodo/pandas/physical/reduce.h
```diff
@@ -97,14 +101,28 @@ class PhysicalReduce : public PhysicalSource, public PhysicalSink {
...
-
-        std::shared_ptr<array_info> result =
-            arrow_array_to_bodo(array, bodo::BufferPool::DefaultPtr());
-        std::vector<std::shared_ptr<array_info>> cvec = {result};
+        // Create a vector of Arrow arrays from output_scalars
+        std::vector<std::shared_ptr<arrow::Array>> arrow_arrays;
+        for (const auto& output_scalar : output_scalars) {
+            if (output_scalar->type->id() != arrow::Type::INT64 &&
+                output_scalar->type->id() != arrow::Type::DOUBLE &&
+                output_scalar->type->id() != arrow::Type::BOOL) {
```
scott-routledge2: There can be other types because of min/max, I think that's why CI is failing.

bodo/pandas/_physical_conv.cpp
chrisoh7: Thanks for pointing out, turns out I don't need the vector at all.

bodo/pandas/_physical_conv.cpp
chrisoh7: Done.

bodo/pandas/physical/reduce.h
chrisoh7: Good point, thanks.

bodo/pandas/physical/reduce.h
chrisoh7: Removed.

bodo/pandas/series.py
```diff
@@ -1499,6 +1533,8 @@ def _split_internal(self, name, pat, n, expand, regex=None):
     # TODO: Implement Series.max()
     n_cols = length_series.max()
 
+    print(n_cols)
```
ehsantn: Leftover print?

bodo/pandas/_plan.cpp
```diff
@@ -138,6 +138,7 @@ duckdb::unique_ptr<duckdb::Expression> make_agg_expr(
 
     // Get DuckDB output type
     auto field = out_schema->field(0);
+    // printf("%s\n", out_schema->ToString().c_str());
```
ehsantn: Leftover print

bodo/pandas/series.py
DrTodd13: @scott-routledge2 I can't say we know enough at this point to definitely say it isn't a problem.  However, I would say most likely it isn't a problem for the duckdb optimizer not to know the correct type of something.  I can't say it isn't going to generate a cast in such cases that would do something to our physical backend potentially resulting in an output type in the result not matching something we computed on the frontend.

PR #637, opened 2025-07-05
Support Python 3.9 for PyIceberg

ehsantn:

PyIceberg's Poetry setup requires Python 3.9: https://github.com/apache/iceberg-python/pull/2167

## Checklist

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -116,7 +116,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python_version: ["3.10", "3.11", "3.12", "3.13"]
+        python_version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
     needs: build_bodosql_wheels
     steps:
       - name: Setup Python
--- a/BodoSQL/pyproject.toml
+++ b/BodoSQL/pyproject.toml
@@ -7,7 +7,7 @@ name = "bodosql"
 dynamic = ["version"]
 description = "Bodo's Vectorized SQL execution engine for clusters"
 readme = "README_pypi.md"
-requires-python = ">=3.10,<3.14"
+requires-python = ">=3.9,<3.14"
 keywords = ["data", "analytics", "cluster"]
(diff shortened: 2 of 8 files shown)


bodo/pandas/array_manager.py
```diff
@@ -87,59 +89,58 @@ def __init__(
...
+                )
+            elif isinstance(head_axis0, pd.PeriodIndex):
+                assert index_data is not None
+                new_axis0 = index_data
+            elif isinstance(head_axis0, pd.TimedeltaIndex):
+                assert index_data is not None
+                new_axis0 = pd.TimedeltaIndex(
+                    index_data, name=head_axis0.name, unit=head_axis0.unit
+                )
+            elif isinstance(head_axis0, pd.Index):
```
ehsantn, line 138: `pd.Index` needs to be moved last in the class matching case since it's the super class of others.

scott-routledge2 (approved):
LGTM, thanks @ehsantn !

buildscripts/bodo/conda-recipe/meta.yaml
```diff
@@ -64,7 +64,7 @@ requirements:
     - pandas >=2.2
     - fsspec >=2021.09
     - pyarrow =19.0
-    - numba 0.61.2
+    - numba >=0.60
```
IsaacWarren, line 67: Should this be `numba >=0.60,<0.62` here and the other places we specify numba version?

IsaacWarren (approved):
Thanks Ehsan!

buildscripts/bodo/conda-recipe/meta.yaml
ehsantn: Specifying upper version limits causes trouble for dependency solvers and is not great practice. We are less sensitive to Numba versions at this point and will upgrade quickly if necessary.

IsaacWarren: Yeah, I'd just be worried that a numba upgrade would break our old package versions forcing users to upgrade their bodo version but if new numba versions are unlikely to cause issues, makes sense

PR #643, opened 2025-07-07
Todd/tpch3

DrTodd13:

1) Rewrite of join code to support non-equi joins.
2) Fixes for groupby.

## Testing strategy

run_ci

## User facing changes

None.

## Checklist


bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -1012,6 +1012,42 @@ def test_merge_switch_side():
     )
 
 
+def test_merge_filter():
+    """Simple test for DataFrame merge."""
```
ehsantn:
```suggestion
def test_merge_non_equi_cond():
    """Simple test for non-equi join conditions."""
```

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -1012,6 +1012,42 @@ def test_merge_switch_side():
...
+    bdf1 = bd.from_pandas(df1)
+    bdf2 = bd.from_pandas(df2)
+
+    df3 = df1.merge(df2, how="inner", left_on=["A"], right_on=["Cat"])
+    bdf3 = bdf1.merge(bdf2, how="inner", left_on=["A"], right_on=["Cat"])
+
+    df4 = df3[df3.B < df3.Dog]
+    bdf4 = bdf3[bdf3.B < bdf3.Dog]
+    # Make sure bdf3 is unevaluated at this point.
+    assert bdf4.is_lazy_plan()
```
ehsantn, line 1040: Check the plan to make sure the filter is pushed to a join condition?

bodo/pandas/groupby.py
```diff
@@ -44,11 +44,7 @@ def __init__(
         self._keys = keys
         self._as_index = as_index
         self._dropna = dropna
-        self._selection = (
```
ehsantn, line 46: What do changes in this file fix? Can we add a unit test for it?

bodo/libs/_bodo_common.cpp
```diff
@@ -475,6 +475,81 @@ std::shared_ptr<::arrow::Field> DataType::ToArrowType(std::string& name) const {
     return std::make_shared<::arrow::Field>(name, dtype, is_nullable);
 }
 
+std::shared_ptr<::arrow::DataType> DataType::ToArrowDataType() const {
```
ehsantn, line 478: This has a lot of code duplication with above. Let's refactor it (maybe even just call the above with a dummy name and get dtype to minimize changes).

bodo/pandas/physical/expression.cpp
```diff
@@ -468,3 +510,28 @@ std::shared_ptr<ExprResult> PhysicalUDFExpression::ProcessBatch(
...
+    }
+    if (res.scalar()->type->id() != arrow::Type::BOOL) {
+        throw std::runtime_error("join_expr_internal did not return bool.");
+    }
+    auto bool_scalar =
+        std::dynamic_pointer_cast<arrow::BooleanScalar>(res.scalar());
+    if (bool_scalar && bool_scalar->is_valid) {
+        return bool_scalar->value;
+    } else {
+        throw std::runtime_error("join_expr_internal bool is null or invalid.");
```
ehsantn, line 533: What happens if one side of an expression is null? Arrow always returns false?

bodo/pandas/_util.cpp
```diff
@@ -546,3 +561,57 @@ PyObject *duckdbFilterSetToPyicebergFilter(
 
     return ret;
 }
+
+arrow::Datum ConvertToDatum(void *raw_ptr,
+                            std::shared_ptr<arrow::DataType> type) {
+    using arrow::Type;
+
+    switch (type->id()) {
```
ehsantn, line 569: Types like datetime/timedelta are missing? Let's open an issue to test this feature with more data types and fill out the gaps.

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -1012,6 +1012,42 @@ def test_merge_switch_side():
...
+    )
+
+    bdf1 = bd.from_pandas(df1)
+    bdf2 = bd.from_pandas(df2)
+
+    df3 = df1.merge(df2, how="inner", left_on=["A"], right_on=["Cat"])
+    bdf3 = bdf1.merge(bdf2, how="inner", left_on=["A"], right_on=["Cat"])
+
+    df4 = df3[df3.B < df3.Dog]
+    bdf4 = bdf3[bdf3.B < bdf3.Dog]
```
ehsantn, line 1038: I think this needs more tests to cover various column index calculations and data types. Could be a follow up issue.

bodo/pandas/series.py
```diff
@@ -672,6 +672,10 @@ def mean(self, axis=0, skipna=True, numeric_only=False, **kwargs):
     def ndim(self) -> int:
         return super().ndim
 
+    @property
```
scott-routledge2: I think this is duplicated above?

bodo/pandas/physical/join.h
```diff
@@ -92,16 +92,53 @@ class PhysicalJoin : public PhysicalSourceSink, public PhysicalSink {
         std::shared_ptr<bodo::Schema> probe_table_schema_reordered =
             probe_table_schema->Project(probe_col_inds);
 
+        for (const duckdb::JoinCondition& cond : conditions) {
```
ehsantn, line 105: I don't this needs passing equality conditions and should only be non-equality ones.

bodo/pandas/physical/join.h
```diff
@@ -92,16 +92,53 @@ class PhysicalJoin : public PhysicalSourceSink, public PhysicalSink {
...
+                                            left_bce.binding.column_index}]],
+                        left_bce.GetName(), true),
+                    std::make_shared<PhysicalColumnRefExpression>(
+                        this->build_col_inds_rev[getColRefMap(
+                            right_bindings)[{right_bce.binding.table_index,
+                                             right_bce.binding.column_index}]],
+                        right_bce.GetName(), false),
+                    cond.comparison));
+            if (physExprTree) {
+                physExprTree = std::static_pointer_cast<PhysicalExpression>(
```
ehsantn, line 134: Add a check to make sure the data types are numeric and throw an error otherwise?

Technically we can support non-numerics like string too since we have table pointers (I think that's the reason for adding table points to the APIs). But can be follow up since it's more work.

bodo/pandas/physical/join.h
```diff
@@ -92,16 +92,53 @@ class PhysicalJoin : public PhysicalSourceSink, public PhysicalSink {
...
+
+        // ---------------------------------------------------
+
         // TODO[BSE-4813]: handle outer joins properly
         bool build_table_outer = false;
         bool probe_table_outer = false;
 
+        cond_expr_fn_t join_func = nullptr;
+        if (has_non_equi_cond) {
+            join_func = PhysicalExpression::join_expr;
```
ehsantn, line 151: Calling the expression evaluator on each row can be expensive. Let's open an issue for performance optimization. Ideally, we should JIT the expression and pass in batches of rows. Let's see how TPC-H queries using this perform and decide on priority.

ehsantn (approved):
Thanks @DrTodd13. This is a great addition. See comments.

bodo/libs/_bodo_common.cpp
DrTodd13: Done.

bodo/tests/test_df_lib/test_end_to_end.py
DrTodd13: Done.

bodo/pandas/physical/join.h
DrTodd13: Not necessary to JIT I think.  If you make the API have the infra to process a batch of rows (push the iteration into the callback) then you can do the conversion to arrow only once and then I think it should be fine.

bodo/pandas/groupby.py
DrTodd13: TPCH Q8 groupby.apply. Added test.

bodo/pandas/_util.cpp
DrTodd13: Added some.

bodo/pandas/physical/join.h
DrTodd13: Yes, I was going to ask you about that.  Made the change.

scott-routledge2 (approved):
LGTM, thanks @DrTodd13

bodo/pandas/physical/join.h
DrTodd13: It is going to throw an error later somewhere else if the type isn't supported.  Is there a real advantage of doing it here?  I added some datetime stuff so we'd have to keep that set and any check here in sync which seems like too much work.

bodo/pandas/series.py
DrTodd13: Done.

PR #651, opened 2025-07-09
[BSE-4946] Suppress subsequent fallback warnings

chrisoh7:

Fixes behavior of Series method fallback raising warnings for every unsupported internal/subsequent call. 

Edit: it turns out a lot of inplace editing methods (like `fillna` and `reset_index`) does not work when `self` is casted to a Pandas Series in getattr. Wrapping the initial fallback method with a warning-silencing wrapper works safely. 

## Testing strategy

N/A

## User facing changes

Users are not spammed with warnings. 

## Checklist

### DIFF
--- a/bodo/pandas/series.py
+++ b/bodo/pandas/series.py
@@ -118,16 +118,22 @@ class BodoSeries(pd.Series, BodoLazyWrapper):
     def __getattribute__(self, name: str):
         """Custom attribute access that triggers a fallback warning for unsupported attributes."""
 
+        ignore_fallback_attrs = ["dtype", "name", "to_string", "attrs", "flags"]
+
         cls = object.__getattribute__(self, "__class__")
         base = cls.__mro__[0]
 
-        if name not in base.__dict__ and not name.startswith("_"):
+        if (
+            name not in base.__dict__
+            and name not in ignore_fallback_attrs
+            and not name.startswith("_")
+        ):
             msg = (
                 f"{name} is not implemented in Bodo Dataframe Library yet. "
                 "Falling back to Pandas (may be slow or run out of memory)."
             )
             warnings.warn(BodoLibFallbackWarning(msg))
-            return object.__getattribute__(pd.Series(self._obj), name)
+            return fallback_wrapper(object.__getattribute__(self, name))
 
         return object.__getattribute__(self, name)
 
@@ -678,6 +684,9 @@ class BodoSeries(pd.Series, BodoLazyWrapper):
         Descriptive statistics include those that summarize the central tendency, dispersion and
         shape of a dataset's distribution, excluding NaN values.
         """
+        if not isinstance(self.dtype, pd.ArrowDtype):
+            raise BodoLibNotImplementedException("self is not a BodoSeries.")
+
         pa_type = self.dtype.pyarrow_dtype
 
         if pa.types.is_null(pa_type):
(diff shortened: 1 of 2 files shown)


DrTodd13 (commented):
Are we sure this isn't too aggressive?  Are there cases where we fallback for one operation but then have a whole bunch of operations we can parallelize?  With the approach in this PR, once you fallback then you stay in regular Pandas unless you explicitly convert back to Bodo?

DrTodd13:
@scott-routledge2 What about fillna followed some arithmetic stuff?  I'm sure we can construct such an example but the big question as always is how common is this going to be in practice.  Anyway, my spidey-sense is tingling on this one so would like to discuss with the larger group tomorrow before committing to this approach.  Did we have any ideas for how to limit the messages using some other approach?  Could we route messages through the decorator that we have so that there is only one fallback per decorator invocation at most?

chrisoh7:
> rns type B then bodo operation A on our equivalent of type B should also return the bodo equivalent of type B. Randomly getting B or the bodo equivalent of B from a bodo operation I can imagine would make programming against the API for performa

I have tried using decorators / global flags / etc., the tricky part is that there isn't a good way to distinguish a getattr call that's made by the User from a call within another method. Decorators and toggling flags ended up not working because I couldn't find a way to tell when the warnings should be enabled again. 

I do understand and agree with what you pointed out, and on a similar note I thought it might be good for the users if we returned a Bodo Series for even the methods that we don't support (our current library returns a Pandas Series), via overriding methods or decorators slapped onto getattr's output. 

As for this PR though, I think its fallback behavior is identical to what we have at the moment, so it is kind of a separate discussion.

bodo/pandas/series.py
```diff
@@ -1151,6 +1151,29 @@ def tz_localize(self, tz=None, ambiguous="NaT", nonexistent="NaT"):
...
+    # Avoid wrapping indexers & accessors
+    if (
+        callable(attr)
+        and not hasattr(attr, "__getitem__")
+        and not hasattr(attr, "__getattr__")
+    ):
+
+        def silenced_method(*args, **kwargs):
+            with warnings.catch_warnings():
+                warnings.simplefilter("ignore")
```
scott-routledge2: Can we filter for specific warnings i.e. BodoLibFallbackWarning (and keep internal Pandas warnings)?

scott-routledge2 (approved):
LGTM, thanks @chrisoh7 !

bodo/pandas/series.py
```diff
@@ -687,6 +684,9 @@ def describe(self, percentiles=None, include=None, exclude=None):
         Descriptive statistics include those that summarize the central tendency, dispersion and
         shape of a dataset's distribution, excluding NaN values.
         """
+        if not isinstance(self.dtype, pd.ArrowDtype):
+            raise BodoLibNotImplementedException("self is not a BodoSeries.")
```
DrTodd13: Message doesn't match the if condition?

bodo/pandas/series.py
```diff
@@ -1151,6 +1151,29 @@ def tz_localize(self, tz=None, ambiguous="NaT", nonexistent="NaT"):
         )
 
 
+def fallback_wrapper(attr):
```
DrTodd13, line 1156: Thanks Chris.  I like the direction this approach is taking much better.   Is it possible to add a test for this?  Do we already have tests that check for an expected fallback message?  Can we add another that would generate multiple warnings if not for this change and then make sure we only get one?

chrisoh7: Sure, just added a frontend test. Thanks!

PR #653, opened 2025-07-09
[BSE-4949] Support column selection in DataFrameGroupby getattr

chrisoh7:

As title.
```
df = pd.DataFrame({'Animal': ['Falcon', 'Falcon',
                              'Parrot', 'Parrot'],
                   'S': [380., 370., 24., 26.]})

print(df.groupby(['Animal']).S.mean())
```
Adds an additional try-except in getattr that runs getitem with `name` as key. On failure, raise original fallback warning and falls back. 

## Testing strategy

N/A

## User facing changes

Users can access groupby columns using simple attribute syntax. 

## Checklist

### DIFF
--- a/bodo/tests/test_df_lib/test_end_to_end.py
+++ b/bodo/tests/test_df_lib/test_end_to_end.py
@@ -1999,6 +1999,23 @@ def test_groupby_getattr_fallback_behavior():
         _ = grouped.not_a_column
 
 
+def test_series_agg():
+    import pandas as pd
+
+    import bodo.pandas as bd
+
+    df = pd.DataFrame({"A": [1, 2, 3, 4, 5]})
+    bdf = bd.from_pandas(df)
+
+    bodo_out = bdf.A.aggregate("sum")
+    pd_out = df.A.aggregate("sum")
+    assert bodo_out == pd_out
+
+    bodo_out = bdf.A.aggregate(["min", "max", "count", "product"])
+    pd_out = df.A.aggregate(["min", "max", "count", "product"])
+    _test_equal(bodo_out, pd_out, check_pandas_types=False)
+
+
 def test_groupby_apply():
     """Test for a groupby.apply from TPCH Q8."""
 
--- a/bodo/pandas/series.py
+++ b/bodo/pandas/series.py
@@ -56,6 +56,7 @@ from bodo.pandas.utils import (
     BodoLibNotImplementedException,
     arrow_to_empty_df,
     check_args_fallback,
+    fallback_wrapper,
     get_lazy_single_manager_class,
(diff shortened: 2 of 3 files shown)


bodo/pandas/groupby.py
```diff
@@ -72,6 +72,11 @@ def __getattribute__(self, name: str, /) -> Any:
         try:
             return object.__getattribute__(self, name)
         except AttributeError:
+            pass
+        try:
```
scott-routledge2: We need to also validate that `attr` is a valid column in the DataFrame, otherwise, if someone does `grouped.apply`, we don't support apply yet, so it will think that "apply" is a column in the DataFrame as opposed to a method of DataFrameGroupBy. (And an edge case would be what happens if the DataFrame actually does have a column named "apply")

chrisoh7: Done.

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -1970,6 +1972,39 @@ def test_series_describe():
...
+
+    df = pds.DataFrame({"apply": [1], "B": [1], "C": [2]})
+    bdf = bd.from_pandas(df)
+
+    grouped = bdf.groupby("B")
+
+    # Accessing a column: should not raise a warning
+    with warnings.catch_warnings(record=True) as record:
+        warnings.simplefilter("always")
+    assert not record, f"Unexpected warning when accessing column: {record}"
```
ehsantn: This doesn't run any code. Also, you need to use `pytest.warns`.

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -1970,6 +1972,39 @@ def test_series_describe():
...
+
+    # Accessing an implemented Pandas GroupBy method: should raise fallback warning
+    with warnings.catch_warnings(record=True) as record:
+        warnings.simplefilter("always")
+    fallback_warnings = [
+        w for w in record if issubclass(w.category, BodoLibFallbackWarning)
+    ]
+    assert len(fallback_warnings) == 1
+    assert "apply" in str(fallback_warnings[0].message)
```
ehsantn: Same.

bodo/tests/test_df_lib/test_end_to_end.py
chrisoh7: Oops, must have gotten deleted by accident.
Also I think pytest.warns can't be used when the intended behavior is not raising a warning. I did make the changes to use `pytest.warns` for the cases after this one.

scott-routledge2 (approved):
LGTM, thanks @chrisoh7 !

PR #655, opened 2025-07-10
BSE-4975: Change parquet readahead

IsaacWarren:
Previously each rank would try to read 16 batches (16 * 128k rows) ahead and 4 fragments (4 files) a head due to Arrow's default settings which are meant for a single reader. This PR adjusts those defaults based on the number of ranks on the node.

## Testing strategy
PR CI + streaming Iceberg read into write benchmark + local nyc taxi head OOM.

## User facing changes
Lower memory consumption during parquet/iceberg reads.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,6 +56,11 @@ bodo/libs/memory.cpp
 bodo/io/pyarrow_wrappers.cpp
 bodo/io/pyarrow_wrappers.h
 bodo/io/pyarrow_wrappers_api.h
+bodo/memory.cpp
+bodo/pandas/plan_optimizer.cpp
+bodo/tests/memory_tester.cpp
+bodo/transforms/type_inference/native_typer.cpp
+bodo/transforms/type_inference/native_typer.h
 
 # Iceberg Connector
 iceberg_db
--- a/bodo/io/arrow_reader.h
+++ b/bodo/io/arrow_reader.h
@@ -341,29 +341,6 @@ class ArrowReader {
         }
     }
 
-    /*
-     * @brief determine how many batches to read ahead. Arrows defaults are
-     * based on a single reader and we need to adjust for the number of ranks on
-     * the node while always ensuring at least one batch is read ahead.
-     */
-    static int32_t batch_readahead() {
-        int ranksOnNode = std::get<0>(dist_get_ranks_on_node());
-        return std::max(arrow::dataset::kDefaultBatchReadahead / ranksOnNode,
-                        1);
-    }
-
-    /*
-     * @brief determine how many batches to read ahead. Arrows defaults are
-     * based on a single reader and we need to adjust for the number of ranks on
-     * the node. We don't make sure this is at least one because fragments can
-     * be large, and if we have a lot of ranks a full extra fragment can be too
-     * much memory.
-     */
-    static int32_t frag_readahead() {
-        int ranksOnNode = std::get<0>(dist_get_ranks_on_node());
-        return arrow::dataset::kDefaultFragmentReadahead / ranksOnNode;
-    }
-
     /**
      * @brief Report Init Stage metrics if they haven't already been reported.
      * Note that this will only report the metrics to the QueryProfileCollector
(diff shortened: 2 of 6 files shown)


scott-routledge2 (approved):
LGTM, thanks @IsaacWarren

bodo/io/iceberg_parquet_reader.cpp
```diff
@@ -1180,8 +1180,7 @@ void IcebergParquetReader::init_scanners() {
             dataset_py, dataset_expr_filter_py, this->selected_fields,
             this->batch_size != -1 ? this->batch_size
                                    : arrow::dataset::kDefaultBatchSize,
-            /*use_threads*/ true, arrow::dataset::kDefaultBatchReadahead,
-            arrow::dataset::kDefaultFragmentReadahead, this->cpu_executor()));
+            /*use_threads*/ true, this->batch_readahead, this->frag_readahead));
```
ehsantn: No need to pass `this->cpu_executor()?` Looks like our `SingleThreadedCpuThreadPool` may be used here?

bodo/io/parquet_reader.cpp
```diff
@@ -182,11 +182,13 @@ void ParquetReader::init_pq_scanner() {
     // get_scanner_batches returns a tuple with the record batch reader and the
     // updated offset for the first batch.
     PyObjectPtr scanner_batches_tup = PyObject_CallMethod(
-        pq_mod, "get_scanner_batches", "OOOdiOOLLOOO", fnames_list_py.get(),
+        pq_mod, "get_scanner_batches", "OOOdiOOLLOOLll", fnames_list_py.get(),
```
ehsantn: Let's avoid `long` for Windows compatibility. It seems fine in this case but hard to know when reading the code.

bodo/io/parquet_reader.cpp
```diff
@@ -182,11 +182,13 @@ void ParquetReader::init_pq_scanner() {
     // get_scanner_batches returns a tuple with the record batch reader and the
     // updated offset for the first batch.
     PyObjectPtr scanner_batches_tup = PyObject_CallMethod(
```
ehsantn, line 182: Delete dead `batch_size_py` code above.

.gitignore
```diff
@@ -56,6 +56,11 @@ bodo/libs/memory.cpp
 bodo/io/pyarrow_wrappers.cpp
 bodo/io/pyarrow_wrappers.h
 bodo/io/pyarrow_wrappers_api.h
+bodo/memory.cpp
+bodo/pandas/plan_optimizer.cpp
+bodo/tests/memory_tester.cpp
+bodo/transforms/type_inference/native_typer.cpp
+bodo/transforms/type_inference/native_typer.h
```
ehsantn: This is added to main already.

bodo/io/arrow_reader.h
```diff
@@ -543,6 +545,40 @@ class ArrowReader {
...
+     * the node while always ensuring at least one batch is read ahead.
+     */
+    int32_t init_batch_readahead() {
+        if (!this->parallel) {
+            // If not parallel, we don't need to adjust the readahead
+            return arrow::dataset::kDefaultBatchReadahead;
+        }
+        int ranksOnNode = std::get<0>(dist_get_ranks_on_node());
+        return std::max(arrow::dataset::kDefaultBatchReadahead / ranksOnNode,
+                        1);
```
ehsantn, line 563: Do we need to reduce batch readahead at all? The default is 8M rows which is independent of file size and row groups.
https://github.com/apache/arrow/blob/3f95bc18828d3649fe0c4a437bd70150819952cb/cpp/src/arrow/dataset/scanner.h#L52

bodo/io/arrow_reader.h
```diff
@@ -543,6 +545,40 @@ class ArrowReader {
...
+     * be large, and if we have a lot of ranks a full extra fragment can be too
+     * much memory.
+     */
+    int32_t init_frag_readahead() {
+        if (!this->parallel) {
+            // If not parallel, we don't need to adjust the readahead
+            return arrow::dataset::kDefaultFragmentReadahead;
+        }
+        int ranksOnNode = std::get<0>(dist_get_ranks_on_node());
+        return arrow::dataset::kDefaultFragmentReadahead / ranksOnNode;
```
ehsantn, line 579: This looks a bit scary in terms of performance. This will be 0 on any machine with more than 4 cores? Is 1 a better minimum? Have you measured performance vs main? Is there a way to be more adaptive and adjust based on fragment size?

ehsantn (commented):
Thanks @IsaacWarren. See comments.

bodo/io/iceberg_parquet_reader.cpp
IsaacWarren: Forgot to revert this from a previous experiment, thanks!

bodo/io/arrow_reader.h
IsaacWarren: If you have 8 columns of int64, that's ~500MiB, not to mention string/other variable size columns, on a 100 + rank node like c6i.32xlarge that's 50GB of ram which is ~20% of the system ram, it seems quite high to me in addition to the 1 open fragment per rank that's actively being read, especially if there are other operations in a pipeline that need RAM like join or groupby

bodo/io/arrow_reader.h
IsaacWarren: Same as the other comment 1 extra fragment could still be really big on a big cluster. We would need to scan the whole dataset ahead of time to know fragment sizes so not sure this is feasible, I'll compare performance of this PR on a c6i.32xlarge and r6i.32xlarge without this PR, it's the same number of ranks r6i just has more RAM so shouldn't OOM

IsaacWarren: I wrote 100 25MB files to s3 and read and summed them on my laptop, the E2E time difference was a couple percent and didn't seem consistent in either direction with this PR vs main

bodo/io/arrow_reader.h
ehsantn: Ok, doesn't matter if performance is ok in general.

bodo/io/arrow_reader.h
ehsantn: Ok, thanks for testing.

PR #661, opened 2025-07-11
[BSE-4979] Expand Series binop

chrisoh7:

Expand Series binary operations (e.g., `__add__`) to support operations between string Series and strings, datetime Series and DateOffset objects, etc. 

* NOTE: upon discovering that operations such as the one below, where fallback to Pandas raises a TypeError due to Arrow lacking support, I added a try-catch inside the fallback wrapper that attempts a deeper fallback, i.e., converting individual elements from Arrow to Pandas. Currently it only handles the case where dtype=timestamp. If a series doesn't fall into any existing cases, it raises the TypeError caught earlier. 
```
import bodo.pandas as pd
import pandas

date = pd.Series(pd.date_range("20130101 09:10:12", periods=10, freq="MS"))
offset = pd.DateOffset(years=+25)
s = pandas.Series([offset] * 10)

print(date + s)
```

## Checklist


bodo/pandas/series.py
```diff
@@ -341,6 +340,49 @@ def _arith_binop(self, other, op, reverse):
...
+                and op in ("__add__", "__radd__")
+            ):
+                if op == "__add__":
+                    return self.str.cat(other)
+                if op == "__radd__":
+                    return other.str.cat(self)
+
+            # TODO: implement Series.add()
+            if op == "__add__":
+                return self.add(other)
```
chrisoh7, line 354: Currently falls back to Pandas for these cases (`add`, `radd`). I'll open an issue for implementing Series.add(). 
But also please do let me know if implementing add from scratch would be necessary or we're good with routing to str.cat and arith_binop.

bodo/pandas/series.py
```diff
@@ -341,6 +340,49 @@ def _arith_binop(self, other, op, reverse):
...
+                and isinstance(other.dtype, pd.ArrowDtype)
+                and other.dtype in allowed_types_map["str_default"]
+                and op in ("__add__", "__radd__")
+            ):
+                if op == "__add__":
+                    return self.str.cat(other)
+                if op == "__radd__":
+                    return other.str.cat(self)
+
+            # TODO: implement Series.add()
```
DrTodd13: can you document what kinds of operations make it to this point?

bodo/pandas/series.py
DrTodd13: string+string routing to cat is fine.  Maybe make arith_binop call numeric_binop for the numeric stuff and non_numeric_binop for the non-numeric stuff (instead of non_arith_binop).  That routine is supposed to provide the implementation for __add__ and __sub__ so then calling a series method named add or sub doesn't seem right to me.  I also didn't understand something you said about writing those in a different PR.  Seems like they need to be present and fallback if nothing else.  Can none of these cases like date/time stuff go through ArithOpExpression?  That might just require extending a few things in the backend but not be that bad.  Anyway, most curious about this section from 358-366 and what the plans are for it.

chrisoh7: So I was referring to these: https://pandas.pydata.org/docs/reference/api/pandas.Series.add.html, my thought from last Friday was calling add() and sub() for series-series operations. But looking into their implementation however, I'm realizing it should be the other way around (add and sub routing to binop).

bodo/pandas/utils.py
```diff
@@ -817,9 +818,36 @@ def fallback_wrapper(attr):
...
         def silenced_method(*args, **kwargs):
+            msg = ""
             with warnings.catch_warnings():
                 warnings.simplefilter("ignore", category=BodoLibFallbackWarning)
-                return attr(*args, **kwargs)
+                try:
+                    return attr(*args, **kwargs)
+                except TypeError as e:
+                    msg = e
+                    pass
```
chrisoh7, line 842: @scott-routledge2 @DrTodd13  Could you provide thoughts on the code below this? It's a 'deeper fallback', there's a note on this in the PR description. Things to note are: 1. it currently only covers case where self.dtype=timestamp, expanding it as we go (discover other fallback failures in the future) was my initial intention, 2. cases when args or kwargs also has a BodoSeries or Series with ArrowDTypes are not handled yet.

bodo/pandas/utils.py
```diff
@@ -817,9 +818,36 @@ def fallback_wrapper(attr):
...
+
+            # In some cases, fallback fails and raises TypeError due to some operations being unsupported between PyArrow types.
+            # Below logic processes deeper fallback that converts problematic PyArrow types to their Pandas equivalents.
+            warnings.warn(
+                BodoLibFallbackWarning(
+                    "TypeError triggering deeper fallback. Converting PyarrowDtype elements in self to Pandas dtypes."
+                )
+            )
+            # TODO: generalize this. Currently, this method is only used for BodoSeries.
+            pd_self = pd.Series(self)
```
DrTodd13: Wrap this in an if checking self for BodoSeries and in the else raise exception.

bodo/pandas/utils.py
```diff
@@ -817,9 +818,36 @@ def fallback_wrapper(attr):
...
+
+            # self.dtype = pd.ArrowDtype(pa.timestamp("ns")): map to_datetime.
```
DrTodd13: Dead code?

chrisoh7: Nope, comments. I'll change it so that it's more obvious.

bodo/pandas/series.py
```diff
@@ -341,6 +336,39 @@ def _arith_binop(self, other, op, reverse):
...
+                    return self.str.cat(other)
+                if op == "__radd__":
+                    return other.str.cat(self)
+
+        # If other is an iterable, fall back to Pandas.
+        # TODO: strengthen this check.
+        elif isinstance(other, allowed_types_map["binop_scalar"]):
+            lhs, rhs = (self, other) if not reverse else (other, self)
+            if op == "__add__":
+                return lhs.map(lambda x: x + rhs)
```
ehsantn: Have you measured the performance of using `map`? It may lead to significant slow down. If so, we can call custom Python functions for add/sub.

chrisoh7: Yes, it gets pretty slow. By 'custom Python functions for add/sub' do you mean a new implementation of a separate method that processes add/sub with scalars?
```
Bodo:
0     2038-01-01 09:10:12
1     2038-02-01 09:10:12
             ...         
98    2046-03-01 09:10:12
99    2046-04-01 09:10:12
Name: A, Length: 100, dtype: timestamp[ns][pyarrow]
total time: 2.14217209815979

Pandas:
0    2038-01-01 09:10:12
1    2038-02-01 09:10:12
             ...        
98   2046-03-01 09:10:12
99   2046-04-01 09:10:12
Length: 100, dtype: datetime64[ns]
total time: 0.003282785415649414
```

ehsantn: Just a simple function like `def add(a, b): a+b` that we call from a Python scalar function.

chrisoh7: Without map and element-wise addition I get
```
TypeError: operation 'add' not supported for dtype 'timestamp[ns][pyarrow]' with object 
of type <class 'pandas._libs.tslibs.offsets.DateOffset'>
```
.

chrisoh7: We could however only use`map` for datetime series and route to python scalar addition in other cases.

ehsantn:
Could you run CI and fix issues before review?

ehsantn:
CI is failing.

bodo/pandas/series.py
```diff
@@ -134,7 +135,7 @@ def _plan(self):
     def __getattribute__(self, name: str):
         """Custom attribute access that triggers a fallback warning for unsupported attributes."""
 
-        ignore_fallback_attrs = ["dtype", "name", "to_string", "attrs", "flags"]
+        ignore_fallback_attrs = ["dtype", "name", "to_string", "attrs", "flags", "iloc"]
```
scott-routledge2: Why suppress the warning for iloc?

bodo/pandas/utils.py
```diff
@@ -817,9 +828,36 @@ def fallback_wrapper(attr):
...
+            # In some cases, fallback fails and raises TypeError due to some operations being unsupported between PyArrow types.
+            # Below logic processes deeper fallback that converts problematic PyArrow types to their Pandas equivalents.
+            if isinstance(self, bodo.pandas.BodoSeries):
+                pd_self = pd.Series(self)
+
+                # When self.dtype is pd.ArrowDtype(pa.timestamp("ns")), apply to_datetime elementwise.
+                if isinstance(self.dtype, pd.ArrowDtype) and pa.types.is_timestamp(
+                    self.dtype.pyarrow_dtype
+                ):
+                    warnings.warn(
```
scott-routledge2, line 853: Could there be some cases where the operation is also a type error in Pandas, e.g adding a string to a timestamp series?

bodo/pandas/series.py
```diff
@@ -1861,6 +1922,37 @@ def _is_pd_pa_timestamp_no_tz(dtype):
     )
 
 
+def gen_arith(self, other, name):
+    """Generates Series.add/radd/sub/rsub."""
+    if isinstance(
```
scott-routledge2, line 1926: Can we make adding,subtracting, etc two Series a followup? For numeric series I think this would be pretty easy to do since we already have the code for it (provided those series come from the same dataframe) in the case where the two Series have different indices, we allow it but the answer is wrong:
``` py
>>> s = bd.Series([1,2,3,4,5], index=[5,4,3,2,1])
>>> s2 = bd.Series([1,2,3,4,5], index=[1, 2, 3, 4, 5])
>>> s + s2
5     2
4     4
3     6
2     8
1    10
```

bodo/pandas/series.py
```diff
@@ -1861,6 +1922,37 @@ def _is_pd_pa_timestamp_no_tz(dtype):
...
+        )
+    if (
+        name
+        in (
+            "sub",
+            "rsub",
+        )
+        and self.dtype in allowed_types_map["str_default"]
+    ):
+        raise TypeError("Unsupported operand type(s) for -: 'str' and 'str'")
```
scott-routledge2, line 1944: Could we generalize this check to make sure that we can "op" self and other. For example things that would be TypeError could be adding a string to a timestamp series, adding a timestamp to a string series, etc.

bodo/pandas/series.py
```diff
@@ -293,20 +294,14 @@ def __invert__(self):
 
     def _arith_binop(self, other, op, reverse):
         """Called when a BodoSeries is element-wise arithmetically combined with a different entity (other)"""
-        from bodo.pandas.base import _empty_like
+        if is_numeric(other):
+            return self._numeric_binop(other, op, reverse)
 
-        if not (
-            (
```
scott-routledge2, line 299: I think we should have some kind of check like this in `numeric_binop` (and an equivalent one in non-numeric binop) that returns are type error if the types are not op-able (to avoid execution, fallback, or unexpected errors).

bodo/pandas/utils.py
chrisoh7: Yes. Adding a string doesn't make it to that stage, but adding a string Series raises a type error and is routed to deeper fallback. Then the logic below also raises a type error, which ends up being the final result. I think it shouldn't be that bad since we end up raising errors correctly?
```
unit = self.dtype.pyarrow_dtype.unit
converted = pd_self.astype(f"datetime64[{unit}]")

return getattr(converted, attr.__name__)(*args[1:], **kwargs)
```

bodo/pandas/series.py
chrisoh7: Sure, I'll create an issue.

bodo/pandas/series.py
chrisoh7: There was a reason behind that at some point but honestly I can't remember. Just removed it.

bodo/pandas/utils.py
```diff
@@ -432,7 +432,17 @@ def run_func_on_table(cpp_table, result_type, in_args):
     # Arrow dtypes can be very slow for UDFs in Pandas:
     # https://github.com/pandas-dev/pandas/issues/61747
     # TODO[BSE-4948]: Use Arrow dtypes when Bodo engine is specified
-    use_arrow_dtypes = not (is_attr and func == "apply")
+    use_arrow_dtypes = not (
+        is_attr
+        and func
+        in (
+            "apply",
```
ehsantn, line 443: Let's add comments about why these methods don't use Arrow.

bodo/pandas/series.py
```diff
@@ -2118,38 +2210,62 @@ def method(self, *args, **kwargs):
...
+    ),
+    "dt.round": (pd.ArrowDtype(pa.timestamp("ns")),),
+    "dt_default": (
         pd.ArrowDtype(pa.timestamp("ns")),
         pd.ArrowDtype(pa.date64()),
         pd.ArrowDtype(pa.time64("ns")),
         pd.ArrowDtype(pa.duration("ns")),
-    ],
+    ),
+    "binop_scalar": (
```
ehsantn, line 2242: What about using `pd.api.types.is_scalar` like elsewhere?

bodo/pandas/utils.py
```diff
@@ -817,9 +828,36 @@ def fallback_wrapper(attr):
...
+                    msg = e
+                    pass
+
+            # In some cases, fallback fails and raises TypeError due to some operations being unsupported between PyArrow types.
+            # Below logic processes deeper fallback that converts problematic PyArrow types to their Pandas equivalents.
+            if isinstance(self, bodo.pandas.BodoSeries):
+                pd_self = pd.Series(self)
+
+                # When self.dtype is pd.ArrowDtype(pa.timestamp("ns")), apply to_datetime elementwise.
+                if isinstance(self.dtype, pd.ArrowDtype) and pa.types.is_timestamp(
```
ehsantn, line 850: Needs to check for timestamps with time zones since `datetime64` cast below wouldn't be valid.

bodo/pandas/utils.py
```diff
@@ -817,9 +828,36 @@ def fallback_wrapper(attr):
...
+                    self.dtype.pyarrow_dtype
+                ):
+                    warnings.warn(
+                        BodoLibFallbackWarning(
+                            "TypeError triggering deeper fallback. Converting PyarrowDtype elements in self to Pandas dtypes."
+                        )
+                    )
+                    unit = self.dtype.pyarrow_dtype.unit
+                    converted = pd_self.astype(f"datetime64[{unit}]")
```
ehsantn: What about using `pd_self.array._pa_array.to_pandas()` instead?

ehsantn (approved):
Looks good to me. See comments.

PR #664, opened 2025-07-12
Handle read Parquet filter columns not used elsewhere in the query

ehsantn:

As title. Fixes a bug in TPC-H Q6.

## Testing strategy

Added unit test.

## User facing changes

Bug fix.

## Checklist


bodo/pandas/_physical_conv.cpp
```diff
@@ -21,8 +21,18 @@ void PhysicalPlanBuilder::Visit(duckdb::LogicalGet& op) {
...
+    // filters that are not used anywhere else in the query.
+    auto& column_ids = op.GetColumnIds();
+    if (!op.projection_ids.empty()) {
+        for (const auto& col : op.projection_ids) {
+            selected_columns.push_back(column_ids[col].GetPrimaryIndex());
+        }
+    } else {
+        for (auto& ci : column_ids) {
+            selected_columns.push_back(ci.GetPrimaryIndex());
+        }
```
DrTodd13, line 35: Are we sure that there are no other unhandled optimizations that duckdb could be doing to us at this point?  We don't have to implement them but perhaps generate a warning or exception if they occur to make it easier to determine something unexpected happened and where we need to go to fix it?  There are a bunch more LogicalGet fields that we don't currently look at, e.g., dynamic_filters.

DrTodd13 (approved):
Thanks for the fix.  IMO, we should have been requiring all fields that duckdb's physical side looked at in LogicalOperators to be checked on our side and either implemented or asserted null/default value/etc.  Of course, that isn't even very clear because I believe not all fields are used by the backend and it isn't marked in any way which are which.  Going forward we should keep that in mind and try to do it when we make changes to other LogicalOperator descendants.

bodo/pandas/_physical_conv.cpp
ehsantn: I don't we know what optimizations could be happening and what all the fields mean since DuckDB isn't documented much. But I agree it's important. Let's open an issue and look at the optimizations and the fields some more (hopefully we know more of DuckDB now making it possible to learn more).

IsaacWarren (approved):
Thanks Ehsan

PR #668, opened 2025-07-14
NYC taxi: Avoid calling cast when the offset is non-zero

scott-routledge2:

Avoid calling `Cast` when the offset is nonzero (casting with nonzero offsets is known to be expensive, since we are copying anyways in this case, we can just perform any casts there).

Timings (1 parquet file, locally): 
```
main -> this branch (jit: time in func/e2e time with cache=True)
1 worker:
50.75s -> 25.34s (jit: 7.78s/9.07s)

4 workers:
 22.50s -> 10.28s (jit: 2.95s/4.38s)

10 workers:
 19.31s -> 12.22s (jit: 4.78s/7.02s)
```

Updated pipeline level timings (1 worker)
```
19PhysicalReadParquet => 0.758594
0 18PhysicalProjection => 2.0828
1 18PhysicalProjection => 2.00395
2 18PhysicalProjection => 1.87779
3 18PhysicalProjection => 1.86219
4 18PhysicalProjection => 2.00889
5 12PhysicalJoin => 1.54852
6 18PhysicalProjection => 0.005274
7 18PhysicalProjection => 0.131971
8 18PhysicalProjection => 5.17249
17PhysicalAggregate => 5.71854
```

## Testing strategy
Existing unit tests, NYC taxi

## Checklist


ehsantn (approved):

bodo/libs/_bodo_to_arrow.cpp
```diff
@@ -1232,35 +1232,76 @@ std::shared_ptr<array_info> arrow_boolean_array_to_bodo(
...
+    auto arrow_bin_arr = static_pointer_cast<ArrowArrayType>(arrow_arr);
     int64_t n = arrow_bin_arr->length();
+    in_offset_t *in_offsets = (in_offset_t *)arrow_bin_arr->raw_value_offsets();
+    in_offset_t first_offset = in_offsets[0];
+
+    if (array_id < 0) {
+        array_id = generate_array_id(n);
+    }
+
+    if (first_offset == 0) {
```
IsaacWarren, line 1265: Add a comment on why we can 0 copy only if first_offset is 0?

scott-routledge2: I am actually not sure what the exact reason is other than the comment below:
```
// Arrow arrays that are outputs of slicing may have a non-zero first
// offset, which is not supported in Bodo. Copying data explicitly in
// this case.
```
@ehsantn did you add this comment?

IsaacWarren: Oh ok, nevermind then, that makes sense

ehsantn: Yes, this is the reason.

PR #675, opened 2025-07-15
Add jit wrapper around map in DataFrame Library

scott-routledge2:

Add an `_engine` parameter which defaults to "bodo" for wrapping the call to map in a `bodo.jit`. If the UDF is not compile-able, then the "python" engine will be used (e.g. calling Pandas map method). 

Followups:
Add jit wrapper for df.apply
investigate overheads in run_func_on_table

## Testing strategy
It was harder to see benefit in the actual taxi example due to it being a small improvement + other operator times varying between executions, so I mostly used this minimal example derived from NYC taxi:

``` py
import bodo.pandas as bd
from tqdm import tqdm
import random
import bodo
import time
import pandas as pd

N = 20_000_000
random.seed(1)

S = pd.Series(
     [random.randint(0, 23) for _ in tqdm(range(N))]
)

def get_time_bucket(t):
    bucket = "other"
    if t in (8, 9, 10):
        bucket = "morning"
    elif t in (11, 12, 13, 14, 15):
        bucket = "midday"
    elif t in (16, 17, 18):
        bucket = "afternoon"
    elif t in (19, 20, 21):
        bucket = "evening"
    return bucket

start = time.perf_counter()
bodo_S = bd.Series(S)
new_S = bodo_S.map(get_time_bucket)
print(new_S)
print("Df took: ", time.perf_counter() - start, "s")
```
### No Jit breakdown of run_func_on_table:
```
Total cpp_table_to_arrow: 0.017 seconds
Total arrow_table_to_pandas: 0.446 seconds
Total call: 3.505 seconds
Total cast_time: 0.977 seconds
Total from_pandas: 0.373 seconds
Total df_to_cpp: 0.483 seconds
Total: 5.802 seconds
```

### Jit breakdown of run_func_on_table:
```
Total cpp_table_to_arrow: 0.017 seconds
Total arrow_table_to_pandas: 0.441 seconds
Total call: 1.691 seconds
Total cast_time: 0.302 seconds
Total from_pandas: 0.371 seconds
Total df_to_cpp: 0.410 seconds
Total: 3.231 seconds
```

## Checklist


bodo/tests/test_series_part2.py
```diff
@@ -227,7 +227,8 @@ def test_impl(S):
     check_func(test_impl, (S,))
 
 
-@pytest.mark.df_lib
+# TODO (df_lib): Support with _engine="bodo"
```
scott-routledge2: Jit doesn't return the right type in this case because we don't support returning tuples (due to arrow not supporting them) and instead we use structs, which causes unexpected results.

bodo/tests/test_series_part2.py
```diff
@@ -52,7 +52,8 @@
         ),
     ],
 )
-def test_series_map_dict_arg(S, d, memory_leak_check):
+# TODO: reenable memory_leak_check
```
scott-routledge2: Not sure why the memory leak check failed here, I tried locally/on platform and didn't see it

bodo/pandas/series.py
```diff
@@ -563,11 +564,51 @@ def dt(self):
...
+                return _get_series_python_func_plan(
+                    self._plan, empty_series, map_wrapper, (), {}, is_method=False
+                )
+            else:
+                msg = (
+                    "Compiling user defined function failed or encountered an unsupported result type."
+                    "Falling back to Python engine, which will"
+                    "run on a small sample of data to determine output types."
+                    "This may hurt performance:"
+                    f"{error_msg}."
```
ehsantn: I'd make this message less scary since not a big deal. Do we even need a warning for users? For cases like AI libraries this may be too annoying.

bodo/pandas/series.py
```diff
@@ -563,11 +564,51 @@ def dt(self):
...
+            try:
+                empty_series = _get_empty_series_arrow(map_wrapper(self.head(0)))
+            except BodoError as e:
+                empty_series = None
+                error_msg = str(e)
+
+            assert empty_series is None or isinstance(empty_series.dtype, pd.ArrowDtype)
+
+            # Jit failed to determine dtypes, likely from gaps in our Arrow support.
+            if empty_series is not None and pa.types.is_null(
```
ehsantn, line 591: When can the output have null dtype? Can you share example codes that show this?

DrTodd13 (approved):
LGTM.  Thanks Scott.

bodo/pandas/series.py
scott-routledge2: Yeah makes sense since this has been the default behavior. My only concern was that someone might think a function is being compiled when it actually isn't. If someone was bothered by the warnings they could just set `engine="python"`, which could be documented in the warning?

bodo/pandas/series.py
scott-routledge2: Yeah, I added this check for this test:
``` py
@pytest.mark.smoke
@pytest.mark.df_lib
def test_series_map_full_pipeline(memory_leak_check):
    """make sure full Bodo pipeline is run on UDFs, including untyped pass."""

    # datetime.date.today() requires untyped pass
    def test_impl(S):
        return S.map(lambda a: datetime.date.today())

    S = pd.Series(pd.date_range(start="2018-04-24", end="2019-04-29", periods=5))
    check_func(test_impl, (S,))
```
    
```
  0    2025-07-17
1    2025-07-17
2    2025-07-17
3    2025-07-17
4    2025-07-17
dtype: object
```
Could be a quick followup to support this case as well.

bodo/pandas/series.py
ehsantn: Up to you. But I'd remove `This may hurt performance` to make it less scary at least.

bodo/pandas/series.py
ehsantn: Still not clear to me why the output would be null here but can be a follow up.

scott-routledge2: The output type is `object`, but there are no rows, so the convert_to_arrow function returns type null. We can either add sampling to the Jit path too or fix the return type

ehsantn: Let's open an issue to fix the return type.

PR #678, opened 2025-07-16
Check and match filter and expression source plans

ehsantn:

Makes sure filter and binary expression arguments have the same source plans. It matches the source plans when one side is just a projection on the other. Fixes an issue in TPC-H Q22.

## Testing strategy

Added unit tests.

## User facing changes

Avoids some segfaults and other issues.

## Checklist

### DIFF
--- a/bodo/pandas/base.py
+++ b/bodo/pandas/base.py
@@ -506,7 +506,7 @@ def concat(
                     if x in old_schema:
                         exprs.extend(make_col_ref_exprs([old_schema.index(x)], plan))
                     else:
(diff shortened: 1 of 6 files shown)


DrTodd13 (approved):
Is it possible to reorganize the hierarchy a bit to introduce an intermediate superclass so that a lot of this repeated code can be eliminated?  Type names can be abstracted.  The code coverage on this patch is very low.  If the reorg above was done that could help with code coverage but in general I'd like to see a higher coverage.

scott-routledge2 (approved):
LGTM, thanks @ehsantn

PR #679, opened 2025-07-16
[BSE-4985] Add TPCH Queries to end-to-end tests

chrisoh7:

Add TPCH Queries to end-to-end tests with skip marker.

## Testing strategy

PR CI, adds unit tests for each tpch query. 

## Checklist


bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -2205,3 +2205,81 @@ def udf(df):
...
+            1 - lineitem_filtered.L_DISCOUNT
+        )
+        lineitem_filtered["CHARGE"] = (
+            lineitem_filtered.L_EXTENDEDPRICE
+            * (1 - lineitem_filtered.L_DISCOUNT)
+            * (1 + lineitem_filtered.L_TAX)
+        )
+        gb = lineitem_filtered.groupby(
+            ["L_RETURNFLAG", "L_LINESTATUS"], as_index=False
+        )[
```
chrisoh7: I copied the code for q01 directly from [dataframe_lib.py](https://github.com/bodo-ai/Bodo/blob/fd0983741adf9e4b4dc663a38c3fddd042bc87da/benchmarks/tpch/dataframe_lib.py), but this test is resulting in an error due to incorrect groupby subset syntax (needs an additional pair of square brackets). But I also noticed that this is marked 'good' on the [spreadsheet](https://docs.google.com/spreadsheets/d/14hzbkCEt6NaaZhlT1fG6NE3yZriYL-zOLdulbt2un0U/edit?gid=0#gid=0), so I'm wondering if I'm missing out on something. @ehsantn ++ It's the same with q02, maybe I shouldn't be running this with the small sample data that I created, and instead use the actual data?

chrisoh7: @DrTodd13

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -2392,3 +2392,99 @@ def test_fallback_wrapper_deep_fallback():
...
+
+    if is_dataframe:
+        bd_result.is_lazy_plan()
+        _test_equal(bd_result, pd_result, check_pandas_types=False, reset_index=True)
+    else:
+        # For scalar or numeric results
+        assert isinstance(pd_result, (int, float)) and isinstance(
+            bd_result, (int, float)
+        )
+        assert abs(pd_result - bd_result) < 1e-6
```
DrTodd13: Use np.isclose.

benchmarks/tpch/dataframe_lib.py
```diff
@@ -10,7 +10,11 @@
 
 
 @functools.lru_cache
-def load_lineitem(data_folder: str) -> pd.DataFrame:
+def load_lineitem(data_folder: str, use_bodo=True) -> pd.DataFrame:
```
DrTodd13: I think this could be cleaner if instead of use_bodo=True, you just passed in the module and I suppose that can default to bodo.pandas, i.e., pd=bodo.pandas.

DrTodd13: I might just remove the return type annotation since it will depend on the value of the pd parameter.

benchmarks/tpch/dataframe_lib.py
```diff
@@ -971,7 +1049,139 @@ def q22(customer, orders):
     )
     total = agg1.merge(agg2, on="CNTRYCODE", how="inner")
     total = total.sort_values(by=["CNTRYCODE"], ascending=[True])
-    print(total)
+    return total
+
+
+@timethis
+@collect_datasets
+def q01(lineitem: pd.DataFrame):
```
DrTodd13: With pd as an argument then remove this type annotation.  That is consistent because the other queries below don't have type annotations.

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -2392,3 +2392,99 @@ def test_fallback_wrapper_deep_fallback():
     assert "TypeError triggering deeper fallback" in warning_msg, (
         f"Unexpected warning message: {warning_msg}"
     )
+
+
+import benchmarks.tpch.dataframe_lib as tpch
```
DrTodd13: Maybe we should have an entirely separate test file for this.  I'm not sure.  Let's get others' opinions.

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -2392,3 +2392,99 @@ def test_fallback_wrapper_deep_fallback():
...
+
+datapath = "bodo/tests/data/tpch-test_data/parquet"
+
+
+@pytest.fixture
+def lineitem():
```
DrTodd13: Will also need the pd argument.

DrTodd13: Repeat for others below.

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -2392,3 +2392,99 @@ def test_fallback_wrapper_deep_fallback():
...
+def region():
+    return tpch.load_region(datapath, use_bodo=False)
+
+
+@pytest.fixture
+def partsupp():
+    return tpch.load_partsupp(datapath, use_bodo=False)
+
+
+@pytest.mark.parametrize(
```
DrTodd13: I want to be able to invoke a specific tpch query through testing from the command-line.  Can I do that with this approach?  I was thinking that this should probably be one test per query and that way you can do per test skips rather than comments in this data structure.  You can probably keep this function in a slightly modified form but as a utility function and not a test entry point.  Each individual tpch query entry point could call this function providing the same information you have in each entries in the below list.

benchmarks/tpch/dataframe_lib.py
DrTodd13: Repeat for all others below.

bodo/tests/test_df_lib/test_end_to_end.py
chrisoh7: Done, can you check it out and let me know if it's good enough? @DrTodd13

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -2392,3 +2392,195 @@ def test_fallback_wrapper_deep_fallback():
...
+
+import benchmarks.tpch.dataframe_lib as tpch
+from benchmarks.tpch.dataframe_lib import _query_to_datasets, collect_datasets
+
+datapath = "bodo/tests/data/tpch-test_data/parquet"
+
+
+def load_tpch_data(name, datapath):
+    if name == "lineitem":
+        return tpch.load_lineitem(datapath, pd=pd)
```
DrTodd13: Can replace the whole function with:
` return getattr(tpch, f"load_{name}")(datapath, pd=pd)`

DrTodd13: I wouldn't worry about the raise on an invalid tpch source name.

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -2392,3 +2392,195 @@ def test_fallback_wrapper_deep_fallback():
...
+    else:
+        # For scalar or numeric results
+        assert isinstance(pd_result, (int, float)) and isinstance(
+            bd_result, (int, float)
+        )
+        assert np.isclose(pd_result, bd_result)
+
+
+def test_tpch_q01():
+    collect_datasets(tpch.q01)
```
DrTodd13: I think this line isn't needed.  It is just doing the query_to_dataset that your run_tpch_query_test is already doing.  Repeat for all tests below.

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -2392,3 +2392,195 @@ def test_fallback_wrapper_deep_fallback():
...
+
+@pytest.mark.skip(
+    reason="RuntimeError: PhysicalReduce::ConsumeBatch: Error in Arrow compute kernel NotImplemented: Function 'sum' has no kernel matching input"
+)
+def test_tpch_q22():
+    collect_datasets(tpch.q22)
+    run_tpch_query_test(tpch.tpch_q22, 22, True)
+
+
+@pytest.mark.skip(reason="Currently only a subset of queries run correctly.")
```
DrTodd13: Don't think this function is needed.

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -2392,3 +2392,195 @@ def test_fallback_wrapper_deep_fallback():
...
+
+def run_tpch_query_test(query_func, num, is_dataframe):
+    pd_args = [load_tpch_data(key, datapath) for key in _query_to_datasets[num]]
+    bd_args = [bd.from_pandas(df) for df in pd_args]
+
+    pd_kwargs = {"pd": pd}
+    pd_result = query_func(*pd_args, **pd_kwargs)
+    bd_result = query_func(*bd_args)
+
+    if is_dataframe:
```
DrTodd13: Can we check if the pandas output is a dataframe or series and if so use this branch, else use the other branch?  Then, you don't need that field that says whether the expected output is a dataframe.

DrTodd13 (commented):
Thanks Chris.  Moving in the right direction.  Also, before the final merge of this PR, please merge with main and run again to finalize the skip list because I notice some of the skip reasons you show shouldn't exist anymore.

DrTodd13:
> > Thanks Chris. Moving in the right direction. Also, before the final merge of this PR, please merge with main and run again to finalize the skip list because I notice some of the skip reasons you show shouldn't exist anymore.
> 
> Just updated it. I'm noticing a lot of `"RuntimeError: PhysicalSort from LogicalOrder with non-empty projection map unimplemented."`, although there are some tests that didn't used to pass that are now passing.

Okay.  Wow, we didn't realize this was happening before and I put in the check as part of yesterday's PR.  I will implement this today.

bodo/tests/test_df_lib/test_end_to_end.py
DrTodd13: I thought that we agreed in the meeting to move these to a separate file?

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -2512,3 +2512,164 @@ def test_fallback_wrapper_deep_fallback():
...
+    bd_result = query_func(*bd_args)
+
+    if isinstance(
+        pd_result,
+        (
+            pd.DataFrame,
+            pd.Series,
+        ),
+    ):
+        bd_result.is_lazy_plan()
```
DrTodd13:
```suggestion
```

DrTodd13 (approved):
Thanks Chris.  Looks good.  Just move to a separate file.

bodo/tests/test_df_lib/test_end_to_end.py
chrisoh7: Yes, I completely missed that, thanks for the heads up.

benchmarks/tpch/dataframe_lib.py
```diff
@@ -1030,5 +1118,5 @@ def main():
 
 
 if __name__ == "__main__":
-    print(f"Running TPC-H against bodo.pandas v{pd.__version__}")
```
scott-routledge2: Keep as bodo.pandas?

PR #680, opened 2025-07-16
BSE-4937: Add bodo iceberg benchmark

IsaacWarren:

Adds code for bodo iceberg benchmark

## Testing strategy
Run on platform, triggered locally

## Checklist


benchmarks/iceberg_io/bodo/copy_orders_bodo.py
```diff
@@ -0,0 +1,51 @@
...
+
+# Get credentials from IMDS
+session = boto3.Session()
+credentials = session.get_credentials().get_frozen_credentials()
+
+# Set credential-related environment variables
+# If we don't do this all workers will try to connect to IMDS
+# and potentially get throttled. The downside is they expire after six hours
+# and won't be refetched automatically
+os.environ["AWS_ACCESS_KEY_ID"] = credentials.access_key
```
scott-routledge2, line 27: Is this necessary to run the benchmark on the platform? Noticed we weren't doing it for pyiceberg.

benchmarks/nyc_taxi/bodo/run_bodo.py
```diff
@@ -24,7 +24,7 @@ def run_bodo_benchmark(num_workers):
     bodo_workspace = BodoWorkspaceClient()
     benchmark_cluster = bodo_workspace.ClusterClient.create(
         name="Benchmark Bodo",
-        instance_type="r6i.16xlarge",
+        instance_type="c6i.32xlarge",
```
scott-routledge2: I think we used `r6i.16xlarge` everywhere for the NYC taxi benchmark?

benchmarks/iceberg_io/bodo/run_bodo.py
```diff
@@ -0,0 +1,68 @@
+"""This script runs the full Bodo benchmark on the Bodo Platform.
```
scott-routledge2, line 1: Would it be possible to reuse this script for pyiceberg as well and add another argument "file path"?

scott-routledge2 (approved):
Thanks @IsaacWarren

benchmarks/iceberg_io/bodo/copy_orders_bodo.py
IsaacWarren: Yeah, pyiceberg doesn't need it because they use threads instead of processes I think

benchmarks/nyc_taxi/bodo/run_bodo.py
IsaacWarren: Thanks, meant to make sure it was this instance type for iceberg, edited the wrong file

benchmarks/iceberg_io/bodo/run_bodo.py
IsaacWarren: We could add an option to use pyiceberg instead but IMO it makes sense for it to be a separate script like the rest. 
The others don't have an option to change the table so for consistency I think I'll leave it without

PR #681, opened 2025-07-16
[BSE-4993] Implement plan execution warning

chrisoh7:

We check is_lazy_plan in the output Bodo dataframe but frontend code may execute a plan unexpectedly and create another output plan. The goal of this PR is to make sure there is no unintended plan execution in intermediate steps of unit tests.

## Checklist

### DIFF
--- a/bodo/pandas/plan.py
+++ b/bodo/pandas/plan.py
@@ -15,7 +15,6 @@ from bodo.pandas.utils import (
     cpp_table_to_df,
     cpp_table_to_series,
     get_n_index_arrays,
-    increment_exec_counter,
     wrap_plan,
 )
 
@@ -832,3 +831,21 @@ def match_binop_expr_source_plans(lhs, rhs):
         return lhs, new_rhs
 
     return None, None
+
+
+# Plan execution counter
+plan_exec_count = 0
+
+
+def increment_exec_counter():
+    global plan_exec_count
+    plan_exec_count += 1
+
+
+def get_exec_counter():
+    return plan_exec_count
+
+
+def reset_exec_counter():
+    global plan_exec_count
+    plan_exec_count = 0
--- a/bodo/pandas/plan.py
+++ b/bodo/pandas/plan.py
@@ -846,7 +846,10 @@ class PlanExecutionCounter:
 
 @contextmanager
 def assert_executed_plan_count(n: int):
-    start = PlanExecutionCounter.get()
-    yield
-    end = PlanExecutionCounter.get()
-    assert end - start == n, f"Expected {n} plan executions, but got {end - start}"
+    try:
+        start = PlanExecutionCounter.get()
+        yield
+        end = PlanExecutionCounter.get()
+        assert end - start == n, f"Expected {n} plan executions, but got {end - start}"
+    finally:
+        PlanExecutionCounter.reset()
--- a/bodo/pandas/plan.py
+++ b/bodo/pandas/plan.py
@@ -15,6 +15,7 @@ from bodo.pandas.utils import (
     cpp_table_to_df,
     cpp_table_to_series,
     get_n_index_arrays,
+    increment_exec_counter,
     wrap_plan,
 )
 
@@ -520,6 +521,8 @@ def execute_plan(plan: LazyPlan):
     if bodo.dataframe_library_debug:
         msg = f"Execution of the following plan is triggered: \n{plan}."
         warnings.warn(BodoPlanExecutionWarning(msg))
+
+    increment_exec_counter()
 
     def _exec_plan(plan):
         import bodo
--- a/bodo/pandas/plan.py
+++ b/bodo/pandas/plan.py
@@ -162,6 +162,11 @@ class Expression(LazyPlan):
         out.is_series = self.is_series
         return out
 
+    def replace_source(self, new_source: LazyPlan):
+        """Replace the source of the expression with a new source plan."""
+        if self.source == new_source:
+            return self
+
 
 class LogicalProjection(LogicalOperator):
     """Logical operator for projecting columns and expressions."""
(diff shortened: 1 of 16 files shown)


bodo/pandas/plan.py
```diff
@@ -363,6 +365,10 @@ def execute_plan(plan: LazyPlan):
     """
     import bodo
 
+    if bodo.dataframe_library_debug:
```
chrisoh7: The warning that I brought up in standup. @ehsantn @DrTodd13

bodo/tests/test_df_lib/test_frontend.py
```diff
@@ -211,3 +211,26 @@ def test_single_fallback_warning_emitted():
...
+    import warnings
+
+    import bodo
+
+    df = pd.DataFrame({"A": ["2", "3"]})
+    bodo.dataframe_library_debug = 1
+
+    with warnings.catch_warnings(record=True) as w:
+        warnings.simplefilter("always")
+        df = df.A.str.lower()
```
ehsantn: Why not use `with pytest.warns`?

bodo/pandas/plan.py
ehsantn: How would you update our unit tests to use this?

DrTodd13:
FWIW, I sill prefer the approach with the global.

ehsantn:
> FWIW, I sill prefer the approach with the global.

I think counting the executed plans in a global actually simplifies creating a context manager for running tests and matching the execution times. I don't know how to do it with this.

bodo/tests/test_df_lib/test_frontend.py
chrisoh7: Oops, changed the second `with` to use `pytest.warns`. For the first one we don't want it to raise any warnings.

bodo/tests/test_df_lib/test_frontend.py
```diff
@@ -241,3 +241,50 @@ def test_single_fallback_warning_emitted():
...
+    assert not w, f"Unexpected warning {w[0]}"
+    assert df.is_lazy_plan()
+
+    with pytest.warns(BodoPlanExecutionWarning) as record:
+        df.execute_plan()
+
+    assert len(record) == 1, "Failed to raise execution warning."
+
+
+def test_execution_counter():
```
chrisoh7, line 246: @ehsantn @DrTodd13 I added a basic execution counter mechanism and a test that demonstrates how we could apply it to unit tests, would be great if you'd be able to take a look and let me know what you think.

bodo/pandas/utils.py
```diff
@@ -864,3 +870,21 @@ def silenced_method(*args, **kwargs):
         return silenced_method
 
     return attr
+
+
+# Plan execution counter
+plan_exec_count = 0
```
DrTodd13: In general, this looks okay but why did you decide to put it in utils.py instead of plan.py?

bodo/tests/test_df_lib/test_frontend.py
```diff
@@ -241,3 +241,50 @@ def test_single_fallback_warning_emitted():
...
+        plans.append(df.A.str.lower())
+
+    assert get_exec_counter() == 0, "Premature execution of LazyPlan."
+
+    for plan in plans:
+        assert plan.is_lazy_plan()
+        plan.execute_plan()
+
+    assert get_exec_counter() == 5, "Unexpected execution count."
+    reset_exec_counter()  # Might not be necessary
```
DrTodd13:
```suggestion
```

bodo/pandas/utils.py
chrisoh7: I'm not sure. I'll move it to plan.py.

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -682,34 +686,36 @@ def test_head(datapath):
...
+            {
+                "a": pd.array([1, 2, 3] * 10, "Int64"),
+                "b": pd.array([4, 5, 6] * 10, "Int64"),
+                "c": ["a", "b", "c"] * 10,
+            },
+            index=index_val[:30],
+        )
+        bdf = bd.from_pandas(df)
+        out_pd = df.apply(lambda x: x["a"] + 1, axis=1)
+        out_bodo = bdf.apply(lambda x: x["a"] + 1, axis=1)
```
DrTodd13, line 707: The execution occurs in the apply frontend but the output is still lazy?  So , perhaps add a lazy check for out_bodo here.

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -682,34 +686,36 @@ def test_head(datapath):
...
+                "A": pd.array([1, 2, 3, 7], "Int64"),
+                "B": ["A1\t", "B1 ", "C1\n", "Abc\t"],
+                "C": pd.array([4, 5, 6, -1], "Int64"),
+            }
+        )
+        df.index = index_val[: len(df)]
+        bdf = bd.from_pandas(df)
+        out_pd = df.B.str.strip().str.lower()
+        out_bodo = bdf.B.str.strip().str.lower()
+        assert out_bodo.is_lazy_plan()
```
DrTodd13: Maybe in general we can remove these is_lazy_plan as that seems redundant with the count check.

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -753,43 +759,52 @@ def test_set_df_column(datapath, index_val):
     df.index = index_val[: len(df)]
     bdf = bd.from_pandas(df)
```
DrTodd13: I'd say the context manager should wrap all the bodo.pandas operation so including this line.  In the other tests the pandas stuff at the beginning was also included and that is okay to keep everything consistent.

bodo/tests/test_df_lib/test_frontend.py
```diff
@@ -241,3 +241,24 @@ def test_single_fallback_warning_emitted():
...
+    plans = []
+
+    with assert_executed_plan_count(0):
+        for _ in range(5):
+            plans.append(df.A.str.lower())
+
+    with assert_executed_plan_count(5):
+        for plan in plans:
+            assert plan.is_lazy_plan()
+            plan.execute_plan()
```
DrTodd13, line 264: Add a case where it fails and it catches the assertion and verifies it.

bodo/pandas/plan.py
```diff
@@ -823,3 +826,30 @@ def match_binop_expr_source_plans(lhs, rhs):
...
+        cls.count = 0
+
+    @classmethod
+    def get(cls):
+        return cls.count
+
+
+@contextmanager
+def assert_executed_plan_count(n: int):
+    try:
```
DrTodd13: What is the reason that the try and the finally with the counter reset is needed?

DrTodd13 (commented):
Thanks Chris.  Mostly looks good.  A few comments.

bodo/pandas/plan.py
chrisoh7: So I thought in the absolute worst case the counter could overflow, but let me know if you think it's fine to just remove it.

bodo/tests/test_df_lib/test_frontend.py
chrisoh7: Can you elaborate a bit more? Do you mean something like this case? 
```
with assert_executed_plan_count(5):
    assert False
        
```

bodo/pandas/plan.py
DrTodd13: Fine to remove it I think. Less than 2^56 seconds before the sun destroys the earth. :)

bodo/tests/test_df_lib/test_frontend.py
DrTodd13:
```
try:
    with assert_executed_plan_count(1):
        pass
except AssertionError as e:
    assert e == "Expected 1 plan executions, but got 0"  # Created an assertion but not the expected error message.
else:
    assert False  # Shouldn't have created an assertion but didn't.
```

bodo/tests/test_df_lib/test_frontend.py
```diff
@@ -241,3 +241,24 @@ def test_single_fallback_warning_emitted():
         and "not implemented" in warning_msg
         and "Falling back to Pandas" in warning_msg
     ), f"Unexpected warning message: {warning_msg}"
+
+
+def test_execution_counter():
+    """Test that plan execution warning is correctly raised."""
```
ehsantn: Update docstring.

PR #682, opened 2025-07-17
Support left/right/outer/cross joins

ehsantn:

As title.

## Testing strategy

Added unit test.

## User facing changes

Left/right/outer/cross joins now work.

## Checklist

### DIFF
--- a/bodo/pandas/_plan.cpp
+++ b/bodo/pandas/_plan.cpp
@@ -586,15 +586,20 @@ duckdb::unique_ptr<duckdb::LogicalComparisonJoin> make_comparison_join(
     // Create join node.
     auto comp_join =
         duckdb::make_uniq<duckdb::LogicalComparisonJoin>(join_type);
+
+    lhs_duck->ResolveOperatorTypes();
+    rhs_duck->ResolveOperatorTypes();
+
     // Create join condition.
-    duckdb::LogicalType cbtype(duckdb::LogicalTypeId::INTEGER);
     for (std::pair<int, int> cond_pair : cond_vec) {
         duckdb::JoinCondition cond;
         cond.comparison = duckdb::ExpressionType::COMPARE_EQUAL;
         cond.left = duckdb::make_uniq<duckdb::BoundColumnRefExpression>(
-            cbtype, lhs_duck->GetColumnBindings()[cond_pair.first]);
+            lhs_duck->types[cond_pair.first],
+            lhs_duck->GetColumnBindings()[cond_pair.first]);
         cond.right = duckdb::make_uniq<duckdb::BoundColumnRefExpression>(
-            cbtype, rhs_duck->GetColumnBindings()[cond_pair.second]);
+            rhs_duck->types[cond_pair.second],
+            rhs_duck->GetColumnBindings()[cond_pair.second]);
         // Add the join condition to the join node.
         comp_join->conditions.push_back(std::move(cond));
     }
--- a/bodo/pandas/physical/join.h
+++ b/bodo/pandas/physical/join.h
@@ -7,23 +7,24 @@
 #include "duckdb/planner/expression/bound_columnref_expression.hpp"
 #include "duckdb/planner/joinside.hpp"
 #include "duckdb/planner/operator/logical_comparison_join.hpp"
+#include "duckdb/planner/operator/logical_cross_product.hpp"
 #include "expression.h"
 #include "operator.h"
 
 #ifndef CONSUME_PROBE_BATCH
-#define CONSUME_PROBE_BATCH(build_table_outer, probe_table_outer,         \
-                            has_non_equi_cond, use_bloom_filter,          \
-                            build_table_outer_exp, probe_table_outer_exp, \
-                            has_non_equi_cond_exp, use_bloom_filter_exp)  \
-    if (build_table_outer == build_table_outer_exp &&                     \
-        probe_table_outer == probe_table_outer_exp &&                     \
-        has_non_equi_cond == has_non_equi_cond_exp &&                     \
-        use_bloom_filter == use_bloom_filter_exp) {                       \
-        is_last = join_probe_consume_batch<                               \
-            build_table_outer_exp, probe_table_outer_exp,                 \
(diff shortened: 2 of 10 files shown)


bodo/pandas/_plan.cpp
```diff
@@ -586,15 +586,20 @@ duckdb::unique_ptr<duckdb::LogicalComparisonJoin> make_comparison_join(
     // Create join node.
     auto comp_join =
         duckdb::make_uniq<duckdb::LogicalComparisonJoin>(join_type);
+
+    lhs_duck->ResolveOperatorTypes();
```
ehsantn, line 590: Existing bug I just found reviewing the join code (not sure how important it is).

bodo/pandas/physical/sort.h
```diff
@@ -17,7 +17,7 @@ class PhysicalSort : public PhysicalSource, public PhysicalSink {
                           std::shared_ptr<bodo::Schema> input_schema,
                           std::vector<duckdb::ColumnBinding>& source_cols,
                           int64_t limit = -1, int64_t offset = -1)
-        : output_schema(input_schema), limit(limit), offset(offset) {
```
ehsantn, line 20: Unused variable fails compilation on Mac.

bodo/pandas/physical/join.h
```diff
@@ -261,60 +334,98 @@ class PhysicalJoin : public PhysicalSourceSink, public PhysicalSink {
...
-                    build_kept_cols, probe_kept_cols, is_last);
+            HashJoinState* join_state =
+                static_cast<HashJoinState*>(this->join_state_.get());
+
+            bool has_bloom_filter = join_state->global_bloom_filter != nullptr;
+
+            std::shared_ptr<table_info> input_batch_reordered =
+                ProjectTable(input_batch, this->probe_col_inds);
+
+            CONSUME_PROBE_BATCH(
```
DrTodd13, line 362: How about something like the following instead of all this repeated code?

```
#include <array>
#include <cstddef>
#include <utility>

// Your template you want to call:
template <bool B0, bool B1, bool B2, bool B3>
void foo() {
  // … your implementation …
}

// A little wrapper so we can take its address:
template <bool B0, bool B1, bool B2, bool B3>
void foo_wrapper() {
  foo<B0, B1, B2, B3>();
}

// Alias for our function‐pointer type
using fn_t = void(*)();

// build a constexpr array of 16 function pointers.
// index = (B0<<0)|(B1<<1)|(B2<<2)|(B3<<3)
template <std::size_t... Is>
static constexpr std::array<fn_t, sizeof...(Is)> make_dispatch_table(std::index_sequence<Is...>) {
  return {{
    &foo_wrapper< (Is & 1) != 0,
                  (Is >> 1 & 1) != 0,
                  (Is >> 2 & 1) != 0,
                  (Is >> 3 & 1) != 0>...
  }};
}

// our global dispatch table
static constexpr auto dispatch_table =
  make_dispatch_table(std::make_index_sequence<16>{});

// runtime dispatcher
inline void dispatch(bool V0, bool V1, bool V2, bool V3) {
  unsigned idx = unsigned(V0)
               | (unsigned(V1) << 1)
               | (unsigned(V2) << 2)
               | (unsigned(V3) << 3);
  dispatch_table[idx]();  // direct O(1) call to foo<…>()
}
```

bodo/pandas/physical/join.h
```diff
@@ -7,8 +7,27 @@
 #include "duckdb/planner/expression/bound_columnref_expression.hpp"
 #include "duckdb/planner/joinside.hpp"
 #include "duckdb/planner/operator/logical_comparison_join.hpp"
+#include "duckdb/planner/operator/logical_cross_product.hpp"
+#include "expression.h"
 #include "operator.h"
 
+#ifndef CONSUME_PROBE_BATCH
+#define CONSUME_PROBE_BATCH(build_table_outer, probe_table_outer,          \
```
DrTodd13, line 15: If using my suggestion below, this becomes a function.

bodo/pandas/physical/join.h
ehsantn: The existing code is long but I think it's a lot easier to understand. Also, we templated this during BodoSQL optimization process I believe so doesn't seem worth it to change it.

bodo/pandas/physical/join.h
```diff
@@ -233,6 +296,17 @@ class PhysicalJoin : public PhysicalSourceSink, public PhysicalSink {
                                 OperatorResult prev_op_result) override {
         bool local_is_last = prev_op_result == OperatorResult::FINISHED;
 
+        if (join_state_->n_keys == 0) {
```
IsaacWarren: Can we have a more clear check that it's a cross join than checking n keys? Even if we just check n_keys in a function called is cross join or something like that

bodo/pandas/physical/join.h
IsaacWarren: I think we were worried about the overheads of function pointers, not sure how valid that is but that was the reasoning

IsaacWarren (approved):
Thanks Ehsan

bodo/pandas/physical/join.h
ehsantn: Done.

bodo/pandas/physical/join.h
DrTodd13: I don't think there is much if any function pointer overhead but in this case, you are taking 4 bools, forming an index from them, doing a table lookup, and running a function pointer versus with the code as it is, don't you always do 30 boolean equality checks (assuming short-circuit) ?  That is true even if it matches the very first case?  From a performance perspective, I would think that 3 left-shifts, 3 ORs, a memory operation, and a call is going to be faster than 30 boolean comparison and 30 boolean ands.

DrTodd13: The existing code could be faster if you goto a label after all those calls if you process the batch.

IsaacWarren: I think it's the memory operation that we were worried about but maybe the comparison need memory operations too I'm not sure without thinking about it more

PR #683, opened 2025-07-18
[WIP] Use cfunc for map

scott-routledge2:
## Checklist


bodo/pandas/series.py
```diff
@@ -630,13 +629,43 @@ def map(self, arg, na_action=None, engine="bodo"):
...
+                array_from_cpp_table,
+                array_to_info,
+                delete_table,
+                table_type,
+            )
+
+            empty_series = self.head(0)
+
+            @bodo.cfunc(table_type(table_type))
+            def map_wrapper(in_cpp_table):
```
scott-routledge2: As a sanity check, this runs without error:
``` py
@bodo.cfunc(table_type(table_type))
def map_wrapper(in_cpp_table):
  return in_cpp_table
```

bodo/decorators.py
```diff
@@ -414,6 +414,38 @@ def _jit(signature_or_function=None, pipeline_class=None, **options):
...
 
+def cfunc(signature_or_function=None, pipeline_class=None, **options):
+    _init_extensions()
+
+    # set nopython by default
+    if "nopython" not in options:
+        options["nopython"] = True
+
+    # options['parallel'] = True
+    options["parallel"] = {
```
DrTodd13, line 426: Possible oversubscription?  Workers should handle parallelization and not have parallel=True within a worker?

DrTodd13: I think we need to think about this because if we put this here then it becomes like a public API and then somebody could try to create a callback using bodo and then we need to understand whether they are using it in a single-thread context where parallelization would be desired or an already parallelized context where it wouldn't.

bodo/pandas/series.py
```diff
@@ -630,13 +629,43 @@ def map(self, arg, na_action=None, engine="bodo"):
...
-                return S.map(arg, na_action=na_action)
+                # Apply map on Series
+                out_series = series.map(arg, na_action=na_action)
+
+                # Convert output series to cpp table
+                out_arr = array_to_info(get_series_data(out_series))
+                out_cpp_table = arr_info_list_to_table([out_arr])
+                return out_cpp_table
+
+            map_wrapper_ptr = ctypes.c_void_p(map_wrapper.address).value
```
DrTodd13: I guess map_wrapper has to be compiled before you can get the address but what here is triggering the compile?

DrTodd13: Also, do we need to have a try here somewhere to catch a compile error and fallback to the Python route?

scott-routledge2: It looks like cfuncs are compiled by the wrapper:
https://github.com/numba/numba/blob/1358274054f88c19838ba11543b8f87b05cf0632/numba/core/decorators.py#L278

We can just wrap this block in a try: catch I think?

DrTodd13: Okay.  You can wrap whole thing in try/catch or just define the method and apply the decorator afterwards in the try/catch.

bodo/decorators.py
scott-routledge2: Could it just be an internal wrapper for now? Seems unlikely it would be useful for users yet.
Also regarding the first point, I think this is what we do in the Jit case as well when distributed=False etc. There were others errors when I set parallel to False.

PR #685, opened 2025-07-18
Fix Series.map bodo engine JIT flags

ehsantn:
As title. Just noticed this issue with some warnings.


scott-routledge2 (approved):
Thanks for the fix @ehsantn

PR #687, opened 2025-07-21
Order projection map and dropduplicates.

DrTodd13:

1. Add support for projection maps in LogicalOrder nodes.
2. Add support for dropduplicates.
3. Flag tpch tests as now working based on above changes.
4. Frame `__setitem__` for existing column now calls internal Pandas function to skip a length check that triggers plan execution.  This is the same trick we use for when we set a new column.

Disabled some of the tpch tests for the moment to try to detect which one is potentially hanging.

## Testing strategy

run_ci

## User facing changes

Dropduplicates works.

## Checklist


bodo/pandas/series.py
```diff
@@ -911,6 +911,10 @@ def rsub(self, other, level=None, fill_value=None, axis=0):
         """Return Subtraction of series and other, element-wise (binary operator rsub)."""
         return gen_arith(self, other, "rsub")
 
+    @property
```
ehsantn: This already exists looks like.

bodo/pandas/frame.py
```diff
@@ -1252,6 +1254,28 @@ def sort_values(
             ),
         )
 
+    @check_args_fallback("none")
+    def drop_duplicates(
```
ehsantn, line 1259: This needs stand alone unit tests in addition to TPC-H tests to test other cases and catch issues independently.

bodo/pandas/frame.py
```diff
@@ -1252,6 +1254,28 @@ def sort_values(
...
+        inplace: bool = False,
+        ignore_index: bool = False,
+    ) -> BodoDataFrame | None:
+        from bodo.pandas.base import _empty_like
+
+        zero_size_self = _empty_like(self)
+        exprs = make_col_ref_exprs(list(range(len(zero_size_self.columns))), self._plan)
+        return wrap_plan(
+            plan=LazyPlan(
+                "LogicalDistinct",
```
ehsantn: Let's use the new standalone classes approach.

bodo/pandas/frame.py
```diff
@@ -1252,6 +1254,28 @@ def sort_values(
             ),
         )
 
+    @check_args_fallback("none")
```
ehsantn: I assume the arguments are not supported yet.

bodo/pandas/_plan.cpp
```diff
@@ -533,9 +558,7 @@ std::vector<int> get_projection_pushed_down_columns(
  *
  */
 static void RunFunction(duckdb::DataChunk &args, duckdb::ExpressionState &state,
-                        duckdb::Vector &result) {
-    throw std::runtime_error("Cannot run Bodo UDFs during optimization.");
-}
+                        duckdb::Vector &result) {}
```
ehsantn, line 561: Why is this removed? I think the exception was necessary to avoid query planning time data access as the comment suggests but don't remember the exact details.

bodo/pandas/_physical_conv.cpp
```diff
@@ -391,3 +391,17 @@ void PhysicalPlanBuilder::Visit(duckdb::LogicalCopyToFile& op) {
     finished_pipelines.emplace_back(this->active_pipeline->Build(physical_op));
     this->active_pipeline = nullptr;
 }
+
+void PhysicalPlanBuilder::Visit(duckdb::LogicalDistinct& op) {
```
ehsantn, line 395: Make sure `op.order_by` is not set? Example DuckDB code: https://github.com/duckdb/duckdb/blob/d29a92f371179170688b4df394478f389bf7d1a6/src/optimizer/pushdown/pushdown_distinct.cpp#L10

bodo/pandas/physical/aggregate.h
```diff
@@ -136,6 +136,54 @@ class PhysicalAggregate : public PhysicalSource, public PhysicalSink {
             /*use_sql_rules*/ false, /* pandas_drop_na_*/ dropna.value());
     }
 
+    explicit PhysicalAggregate(std::shared_ptr<bodo::Schema> in_table_schema,
+                               duckdb::LogicalDistinct& op) {
+        std::map<std::pair<duckdb::idx_t, duckdb::idx_t>, size_t> col_ref_map =
```
ehsantn, line 120: Refactoring of the duplicated code here would be great.

bodo/pandas/series.py
DrTodd13: Looks like a merge error.  Fixed it.

bodo/pandas/frame.py
DrTodd13: Done.

bodo/pandas/_plan.cpp
DrTodd13: Well, if anything is going to access the data during query planning time it would be this function but if you don't do anything in this function then I am assuming that it won't happen downstream.  I removed it because one of the queries calls this function and so errors out.  I think if you don't do anything in this function you get something like NA that doesn't match anything and so won't trigger any erroneous optimizations.  My investigation of this was somewhat limited so there could be undesired downstream effects.

bodo/pandas/physical/aggregate.h
DrTodd13: Done.

bodo/pandas/frame.py
DrTodd13: Clarified supported="none".

bodo/pandas/_physical_conv.cpp
DrTodd13: Good catch.  I originally wrote this code before we decided to be careful about these unchecked fields.

bodo/pandas/frame.py
DrTodd13: Done.

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -2533,3 +2533,18 @@ def test_fallback_wrapper_deep_fallback():
     assert "TypeError triggering deeper fallback" in warning_msg, (
         f"Unexpected warning message: {warning_msg}"
     )
+
+
+def test_drop_duplicates():
```
scott-routledge2, line 2538: This test is currently failing on CI, maybe we need to sort the output before checking?

bodo/pandas/physical/aggregate.h
```diff
@@ -136,6 +115,34 @@ class PhysicalAggregate : public PhysicalSource, public PhysicalSink {
...
+        // single data column
+        std::vector<int32_t> f_in_offsets(f_in_cols.size() + 1);
+        std::iota(f_in_offsets.begin(), f_in_offsets.end(), 0);
+
+        this->groupby_state = std::make_unique<GroupbyState>(
+            std::make_unique<bodo::Schema>(*in_table_schema_reordered), ftypes,
+            std::vector<int32_t>(), f_in_offsets, f_in_cols, this->keys.size(),
+            std::vector<bool>(), std::vector<bool>(), cols_to_keep_vec, nullptr,
+            get_streaming_batch_size(), true, -1, -1, -1, false, std::nullopt,
+            /*use_sql_rules*/ false, /* pandas_drop_na_*/ true);
```
scott-routledge2: I believe `pandas_drop_na_` should be false in this case (we don't want to drop keys containing missing values) Can add a few missing values to the unit test below to confirm:

``` py
>>> df
     brand style  rating
0     None   cup     4.0
1  Yum Yum   cup     4.0
2  Indomie   cup     3.5
3  Indomie  pack    15.0
4  Indomie  pack     5.0
>>> df.drop_duplicates()
     brand style  rating
0     None   cup     4.0
1  Yum Yum   cup     4.0
2  Indomie   cup     3.5
3  Indomie  pack    15.0
4  Indomie  pack     5.0
```

scott-routledge2 (approved):
Thanks @DrTodd13 , see my comments about drop_duplicates testing.

bodo/tests/test_df_lib/test_end_to_end.py
DrTodd13: @scott-routledge2 Where are you seeing that?  There was a problem in an earlier version where static cardinality was wrongly present for aggregate making the output seem way too big even though the backing data was the right size.  I fixed that but right now I don't see drop_duplicates in the output at all.

bodo/pandas/physical/aggregate.h
DrTodd13: Nice catch.  Thanks.

bodo/tests/test_df_lib/test_end_to_end.py
scott-routledge2: I had to look at the raw logs since they were getting cut off (the gear button on the side)

PR #711, opened 2025-07-29
Add names to query profiler

scott-routledge2:
## Checklist


IsaacWarren (approved):
Thanks Scott. bodosql_streaming_batch_size also needs updated in bodo/__init__.py

scott-routledge2:
> Thanks Scott. bodosql_streaming_batch_size also needs updated in bodo/**init**.py

I think that is only used in bodosql? Since we don't have any Python code in the DataFrame Library that uses `bodosql_streaming_batch_size`

IsaacWarren:
> > Thanks Scott. bodosql_streaming_batch_size also needs updated in bodo/**init**.py
> 
> I think that is only used in bodosql? Since we don't have any Python code in the DataFrame Library that uses `bodosql_streaming_batch_size`
> 
> Ohh I see so it is consistent with _shuffle.h?

Yeah, just want to make sure they match

PR #712, opened 2025-07-29
[BSE-4970] Support Series.quantile

chrisoh7:
## Checklist

### DIFF
--- a/bodo/pandas/_physical_conv.cpp
+++ b/bodo/pandas/_physical_conv.cpp
@@ -11,6 +11,7 @@
 #include "physical/join.h"
 #include "physical/limit.h"
 #include "physical/project.h"
+#include "physical/quantile.h"
 #include "physical/reduce.h"
 #include "physical/sample.h"
 #include "physical/sort.h"
@@ -123,6 +124,36 @@ void PhysicalPlanBuilder::Visit(duckdb::LogicalAggregate& op) {
             // The same operator will exist in both pipelines.  The sink of the
             // previous pipeline and the source of the next one.
             // We record the pipeline dependency between these two pipelines.
+            this->active_pipeline =
+                std::make_shared<PipelineBuilder>(physical_op);
+            return;
+        }
+        if (function_names[0].starts_with("quantile")) {
+            auto& agg_expr =
+                op.expressions[0]->Cast<duckdb::BoundAggregateExpression>();
+            BodoAggFunctionData& bind_info =
+                agg_expr.bind_info->Cast<BodoAggFunctionData>();
+            auto bodo_schema = std::make_shared<bodo::Schema>();
+            auto col_schema = bind_info.out_schema;
+            auto bodo_col_schema = bodo::Schema::FromArrowSchema(col_schema);
+            for (size_t i = 0; i < bodo_col_schema->column_types.size(); i++) {
+                bodo_schema->append_column(
+                    bodo_col_schema->column_types[i]->copy());
+                bodo_schema->column_names.push_back(std::to_string(i));
+            }
+            bodo_schema->metadata = std::make_shared<TableMetadata>(
+                std::vector<std::string>({}), std::vector<std::string>({}));
+
+            // TODO: do preprocessing and validations as necessary.
+            // TODO: decide if out_schema is necessary in this case.
+            std::vector<float> quantiles{};
+            for (auto it : function_names) {
+                quantiles.push_back(std::stod(it.substr(9)));
+            }
+            auto physical_op =
+                std::make_shared<PhysicalQuantile>(bodo_schema, quantiles);
+            finished_pipelines.emplace_back(
+                this->active_pipeline->Build(physical_op));
             this->active_pipeline =
                 std::make_shared<PipelineBuilder>(physical_op);
             return;
--- a/bodo/pandas/physical/quantile.h
+++ b/bodo/pandas/physical/quantile.h
@@ -1,0 +1,148 @@
+#pragma once
+
+#include <arrow/compute/expression.h>
+#include <arrow/scalar.h>
+#include <arrow/type_fwd.h>
+#include <memory>
+#include <stdexcept>
+#include <utility>
+#include "../../libs/_distributed.h"
+#include "../libs/_bodo_to_arrow.h"
+#include "/tmp/datasketches-prefix/include/DataSketches/kll_sketch.hpp"
+#include "operator.h"
+#include "physical/expression.h"
+
+using KLLDoubleSketch = datasketches::kll_sketch<double>;
+
+#undef CHECK_ARROW
+#define CHECK_ARROW(expr, msg)                                             \
+    if (!(expr.ok())) {                                                    \
+        std::string err_msg = std::string("PhysicalKLL::ConsumeBatch: ") + \
+                              msg + " " + expr.ToString();                 \
+        throw std::runtime_error(err_msg);                                 \
+    }
+
+class PhysicalQuantile : public PhysicalSource, public PhysicalSink {
+   public:
+    explicit PhysicalQuantile(
+        std::shared_ptr<bodo::Schema> out_schema, std::vector<float> quantiles,
+        uint16_t k = datasketches::kll_constants::DEFAULT_K)
+        : out_schema(out_schema),
+          quantiles(quantiles),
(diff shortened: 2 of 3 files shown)


bodo/pandas/_physical_conv.cpp
```diff
@@ -127,6 +128,36 @@ void PhysicalPlanBuilder::Visit(duckdb::LogicalAggregate& op) {
...
+            for (size_t i = 0; i < bodo_col_schema->column_types.size(); i++) {
+                bodo_schema->append_column(
+                    bodo_col_schema->column_types[i]->copy());
+                bodo_schema->column_names.push_back(std::to_string(i));
+            }
+            bodo_schema->metadata = std::make_shared<TableMetadata>(
+                std::vector<std::string>({}), std::vector<std::string>({}));
+
+            // TODO: do preprocessing and validations as necessary.
+            // TODO: decide if out_schema is necessary in this case.
```
scott-routledge2: What does this todo mean? I think out_schema is essentially `bodo_schema` (it should always be some number of float64 in this case depending on the length of the quantiles array)

Also, I think we should do argument validation (i.e. check q is in [0,1] in Python side of things)

bodo/pandas/_physical_conv.cpp
```diff
@@ -127,6 +128,36 @@ void PhysicalPlanBuilder::Visit(duckdb::LogicalAggregate& op) {
                 std::make_shared<PipelineBuilder>(physical_op);
             return;
         }
+        if (function_names[0].starts_with("quantile")) {
+            auto& agg_expr =
```
scott-routledge2: Could potentially reuse some of the code for creating the output schema below

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -2895,3 +2895,133 @@ def test_dataframe_reset_index_pipeline():
...
         reset_index=False,
     )
+
+
+@pytest.mark.parametrize("quantiles", [[0, 0.25, 0.5, 0.75, 0.9, 1]])
+def test_series_quantile(quantiles):
+    """Tests that approximate quantiles using KLL fall within expected error bounds."""
+
+    def kll_error_bounds(q, k=200, pmf=False):
+        eps = 1.0 / np.sqrt(k) * (1.7 if pmf else 1.33)
```
scott-routledge2, line 2942: What is pmf?

bodo/pandas/series.py
```diff
@@ -1078,6 +1079,65 @@ def reset_index(
         """
         return reset_index(self, drop, level, name=name)
 
+    @check_args_fallback(unsupported=["interpolation"])
+    def quantile(self, q=0.5, interpolation="linear"):
```
scott-routledge2: I don't think the interpolation argument makes sense in this case? Maybe we should fallback if it was supplied at all?

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -2895,3 +2895,133 @@ def test_dataframe_reset_index_pipeline():
...
+
+        assert true_low <= approx <= true_high, (
+            f"Quantile {q} estimate {approx} not within [{true_low}, {true_high}]"
+        )
+
+
+def test_series_quantile_empty():
+    """Tests that quantile on an empty BodoSeries returns either a scalar pd.NA or a BodoSeries of pd.NA."""
+
+    pds = pd.Series([])
```
scott-routledge2, line 3028: I think this will have pyarrow type null so it won't go through the C++ code path, could also do an example where the type is int but there is no data, for example:
`s = pd.Series([1,2,3]); s = s[s>4]`
Also, try a singleton Series with multiple quantile values?

bodo/pandas/physical/quantile.h
```diff
@@ -0,0 +1,159 @@
...
+            }
+            all_bytes.resize(total_size);
+        }
+
+        CHECK_MPI(MPI_Gatherv(bytes.data(), local_size, MPI_BYTE,
+                              all_bytes.data(), recv_counts.data(),
+                              displs.data(), MPI_BYTE, 0, MPI_COMM_WORLD),
+                  "quantile.h::Finalize(): MPI error on MPI_Gatherv:");
+
+        // TODO: consider possible optimization of merging and querying sketches
```
scott-routledge2: I think you could implement a custom reduction operator although in my experience that can get pretty complicated.

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -2895,3 +2895,133 @@ def test_dataframe_reset_index_pipeline():
...
+            "B": [0.5, 1.5, 2.5] * 30 + [5.5] * 10,
+            "C": list(range(100)),
+            "D": [100, 200, 300] * 30 + [None] * 10,
+        }
+    )
+
+    for col in df.columns:
+        s = df[col]
+        with assert_executed_plan_count(1):
+            approx = s.quantile(q)
```
scott-routledge2, line 3004: Test passing a list of quantiles as well?

scott-routledge2 (approved):
A couple testing/validating suggestion but overall LGTM, thanks @chrisoh7 !

bodo/tests/test_df_lib/test_end_to_end.py
chrisoh7: They use different values for calculating single-sided error (our case) vs. double-sided error (for getting pmf estimations). Technically we're not going to estimate prob. mass functions ever so let me know if I should just get rid of that parameter.

bodo/pandas/series.py
chrisoh7: We fall back when quantile is called with interpolation, but are you suggesting we should just delete it from the method signature?

scott-routledge2: I was thinking we could the default value in the signature to be "no default", so it also fallsback if someone does `quantile(0.5, interpolation="linear")` as well

bodo/pandas/_physical_conv.cpp
chrisoh7: Done.

bodo/pandas/_physical_conv.cpp
chrisoh7: Done.

bodo/tests/test_df_lib/test_end_to_end.py
chrisoh7: Done.

bodo/tests/test_df_lib/test_end_to_end.py
chrisoh7: This one covers lists. The next one covers scalar q's.

bodo/pandas/physical/quantile.h
```diff
@@ -0,0 +1,169 @@
...
+        override {
+        int rank, size;
+        MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+        MPI_Comm_size(MPI_COMM_WORLD, &size);
+
+        auto bytes = sketch->serialize();
+        int local_size = (int)bytes.size();
+
+        std::vector<int> recv_counts(size);
+        CHECK_MPI(MPI_Gather(&local_size, 1, MPI_INT, recv_counts.data(), 1,
```
ehsantn, line 99: You can't have blocking collectives in streaming loops. This should be in finalize stage of build to ensure single run.

chrisoh7: Done. It was moved to ProduceBatch from Finalize because the collection was happening twice (since Finalize called twice). I moved it back to Finalize and added a flag to prevent it.

bodo/pandas/_physical_conv.cpp
```diff
@@ -144,6 +145,29 @@ void PhysicalPlanBuilder::Visit(duckdb::LogicalAggregate& op) {
         }
         bodo_schema->metadata = std::make_shared<TableMetadata>(
             std::vector<std::string>({}), std::vector<std::string>({}));
+
+        // If function_names includes quantiles, create a PhysicalQuantile
+        // operator
+        if (function_names[0].starts_with("quantile")) {
```
DrTodd13, line 152: description of function name structure so that line 159 below makes sense.

DrTodd13 (approved):

PR #714, opened 2025-07-30
Support writing to S3 Vectors

ehsantn:

As title. Uses boto3 in backend for now until AWS C++ SDK with s3vectors enabled becomes available in our dependencies.

## Testing strategy

Add unit tests.

## User facing changes

Writing to S3 Vectors is added.

## Checklist


bodo/pandas/physical/write_s3_vectors.h
```diff
@@ -0,0 +1,101 @@
...
+        // Sync is_last flag
+        bool is_last = prev_op_result == OperatorResult::FINISHED;
+        is_last = static_cast<bool>(sync_is_last_non_blocking(
+            is_last_state.get(), static_cast<int32_t>(is_last)));
+
+        if (is_last) {
+            finished = true;
+        }
+
+        iter++;
```
DrTodd13, line 38: iter is unused?

bodo/tests/test_df_lib/test_ai_funcs.py
```diff
@@ -0,0 +1,91 @@
...
+        }
+    )
+    df["metadata"] = [
+        {"source_text": texts[0], "genre": "scifi"},
+        {"source_text": texts[1], "genre": "scifi"},
+        {"source_text": texts[2], "genre": "family"},
+    ]
+    bdf = bd.from_pandas(df)
+    region = "us-east-2"
+    bucket_name = "ehsan-test-vector"
```
DrTodd13, line 41:
```suggestion
    bucket_name = "test-write-s3-vectors-test-vector"
```

bodo/pandas/physical/write_s3_vectors.h
ehsantn: Correct, but useful for debugging and logging (to be added soon).

bodo/tests/test_df_lib/test_ai_funcs.py
ehsantn: The bucket type is S3 Vectors so `ehsan-test-vector` won't conflict with anything.

.github/workflows/_test_python_source.yml
```diff
@@ -62,9 +62,10 @@ jobs:
             role-session-name: BodoEnginePrCiSession
             role-skip-session-tagging: true
             role-duration-seconds: 10800
-        - name: Install Extra Test dependencies
-          # This isn't published on pypi so no good way to list it as a dependency
-          run: pip install 'git+https://github.com/apache/polaris.git@release/1.0.x#subdirectory=client/python'
+        # TODO[BSE-5031]: Enable polaris tests after fixing pip installation issues
+        # - name: Install Extra Test dependencies
+        #   # This isn't published on pypi so no good way to list it as a dependency
+        #   run: pip install 'git+https://github.com/apache/polaris.git@release/1.0.x#subdirectory=client/python'
```
IsaacWarren: I don't like this but I guess they weren't running anyways

bodo/pandas/_plan.h
```diff
@@ -350,6 +350,19 @@ duckdb::unique_ptr<duckdb::LogicalCopyToFile> make_iceberg_write_node(
...
+ * @param source input data to write
+ * @param pyarrow_schema schema of the data to write
+ * @param vector_bucket_name name of the S3 bucket to write vectors to
+ * @param index_name name of the vector index to write to
+ * @param region AWS region for the S3 bucket
+ * @return duckdb::unique_ptr<duckdb::LogicalCopyToFile> created node
+ */
+duckdb::unique_ptr<duckdb::LogicalCopyToFile> make_s3_vectors_write_node(
+    std::unique_ptr<duckdb::LogicalOperator> &source, PyObject *pyarrow_schema,
+    std::string vector_bucket_name, std::string index_name, PyObject *region);
```
IsaacWarren, line 366: Why is region a pyobject instead of a string?

bodo/pandas/frame.py
```diff
@@ -576,6 +577,56 @@ def to_iceberg(
         if not success:
             raise BodoError("Iceberg write failed.")
 
+    @check_args_fallback(unsupported="none")
+    def to_s3_vectors(
+        self,
+        vector_bucket_name: str,
+        index_name: str,
+        region: str = None,
```
IsaacWarren, line 585: The design had this taking 3 series instead of this being a dataframe method so we don't rely on a dataframe with specific column names

bodo/pandas/physical/write_s3_vectors.h
IsaacWarren: Can we add query profiler metrics? That way we can just use iter from the calling exector loop

bodo/pandas/physical/write_s3_vectors.h
```diff
@@ -0,0 +1,101 @@
...
+                "Failed to import bodo.pandas.utils module");
+        }
+
+        // Call the write_s3_vectors_helper() with the table_info pointer and
+        // vector/index names
+        PyObject* vector_bucket_name_py =
+            PyUnicode_FromString(vector_bucket_name.c_str());
+        PyObject* index_name_py = PyUnicode_FromString(index_name.c_str());
+        if (!vector_bucket_name_py || !index_name_py) {
+            Py_DECREF(bodo_module);
```
IsaacWarren, line 73: We might need to decref vector_bucket_name_py or index_name_py depending on which is nullptr

bodo/pandas/physical/write_s3_vectors.h
```diff
@@ -0,0 +1,101 @@
...
+            throw std::runtime_error(
+                "Failed to create Python strings for bucket and index names");
+        }
+        PyObject* result = PyObject_CallMethod(
+            bodo_module, "write_s3_vectors_helper", "LOOO",
+            reinterpret_cast<int64_t>(new table_info(*input_batch)),
+            vector_bucket_name_py, index_name_py, region);
+        if (!result) {
+            PyErr_Print();
```
IsaacWarren, line 82: Same here

bodo/pandas/physical/write_s3_vectors.h
```diff
@@ -0,0 +1,101 @@
...
+        // Clean up Python objects
+        Py_DECREF(vector_bucket_name_py);
+        Py_DECREF(index_name_py);
+        Py_DECREF(bodo_module);
+        Py_DECREF(result);
+    }
+
+    std::string vector_bucket_name;
+    std::string index_name;
+    PyObject* region;
```
IsaacWarren: These can all be const?

bodo/pandas/utils.py
```diff
@@ -595,6 +595,26 @@ def run_func_on_table(cpp_table, result_type, in_args):
...
+
+    df = cpp_table_to_df(cpp_table)
+
+    if not len(df):
+        return
+
+    df = df.loc[:, ["key", "data", "metadata"]]
+    df["data"] = df.data.map(lambda x: {"float32": x.tolist()})
+
+    s3vectors = boto3.client("s3vectors", region_name=region)
```
IsaacWarren, line 636: Is creating this client every batch expensive?

.github/workflows/_test_python_source.yml
ehsantn: I agree. opened a follow up to fix it assuming upstream doesn't change by the time we enable Iceberg tests.

bodo/pandas/_plan.h
ehsantn: Because it could be None and it's easier to just pass it through.

bodo/pandas/frame.py
ehsantn: The problem with 3 series objects is that we would have to make sure they are in the same source dataframe to match the relational model which has reliability implications. This implicitly communicates to the user about matching values too. I think I've seen APIs requiring a dataframe with specific column names (maybe Scikit-learn APIs are like that?).

bodo/pandas/physical/write_s3_vectors.h
ehsantn: This is AI generated and I usually don't worry about making these error paths perfect since the program is usually exiting in our use cases anyways (and these objects are very small).

bodo/pandas/physical/write_s3_vectors.h
ehsantn: Same.

bodo/pandas/physical/write_s3_vectors.h
ehsantn: Done for the string ones. CPython PyObject APIs usually don't take consts.

bodo/pandas/utils.py
ehsantn: Didn't seem like it. I create a follow up issue to optimize in general after we gain more experience with it: https://bodo.atlassian.net/browse/BSE-5032

ehsantn:
> Thanks Ehsan, left a couple comments. My main concern is the API

See comment on the API. I'd be happy to discuss more.

bodo/pandas/physical/write_s3_vectors.h
ehsantn: left for follow ups to limit the scope.

IsaacWarren (approved):
Thanks Ehsan, the metrics are really simple if you take a look at another operator

ehsantn:
> Thanks Ehsan, the metrics are really simple if you take a look at another operator

Will do in the immediate follow up PR.

PR #715, opened 2025-07-30
[BSE-4977] Complete support for numeric/nonnumeric Series.describe

chrisoh7:

Adds support for nonnumeric describe, and replaces Pandas fallback quantile with Bodo's newly implemented quantile. 

## Testing strategy

PR CI, edits existing numeric describe tests to take approx. quantile into account. Adds end-to-end tests for non-numeric describe. 

## User facing changes

Series.describe() for both numeric and nonnumeric BodoSeries are fully supported without fallbacks.

## Checklist

### DIFF
--- a/bodo/tests/test_df_lib/test_end_to_end.py
+++ b/bodo/tests/test_df_lib/test_end_to_end.py
@@ -2289,8 +2289,7 @@ def test_series_describe_numeric(percentiles):
             "B": np.flip(np.arange(n, dtype=np.int32)),
             "C": np.append(np.arange(n // 2), np.flip(np.arange(n // 2))),
             "D": np.append(np.flip(np.arange(n // 2)), np.arange(n // 2)),
-            "E": [None] * n,
-            "F": list(range(n - 1)) + [None],
+            "E": list(range(n - 1)) + [None],
         }
     )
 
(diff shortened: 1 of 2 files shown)


bodo/pandas/plan.py
```diff
@@ -1014,3 +1014,30 @@ def assert_executed_plan_count(n: int):
     yield
     end = PlanExecutionCounter.get()
     assert end - start == n, f"Expected {n} plan executions, but got {end - start}"
+
+
+def nonnumeric_describe(series):
```
scott-routledge2: Move to series.py?

bodo/pandas/plan.py
```diff
@@ -1014,3 +1014,30 @@ def assert_executed_plan_count(n: int):
...
+    """Computes non-numeric series.describe() using DataFrameGroupBy."""
+
+    plan = series._plan
+
+    plan.is_series = False
```
scott-routledge2: Maybe a comment about why this workaround is necessary (i.e. we don't support size() on zero columns input, we don't support series groupby )

docs/docs/api_docs/dataframe_lib/series/index.md
```diff
@@ -17,14 +17,14 @@ The Bodo DataFrame Library supports Pandas Series methods and accessors that are
 ## Computations / descriptive stats
 - [`bodo.pandas.BodoSeries.abs`][bodoseriesabs]
 - [`bodo.pandas.BodoSeries.clip`][bodoseriesclip]
+- [`bodo.pandas.BodoSeries.describe`][bodoseriesdescribe]
```
scott-routledge2, line 20: We should probably add a separate docs page for describe/quantile explaining that we compute approximate percentiles, and other differences between the two, can be a follow up

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -2291,6 +2297,49 @@ def test_series_describe():
         ):
             describe_pd = df[c].describe()
             describe_bodo = bdf[c].describe()
+
+        # For quantile columns, check approximate bounds instead of strict equality
```
scott-routledge2, line 2304: Add test with percentile argument specified?

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -2291,6 +2297,49 @@ def test_series_describe():
...
+            "B": ["apple"] * 3 + ["APPLE"] * 2 + [None],
+        }
+    )
+
+    bdf = bd.from_pandas(df)
+    for c in df.columns:
+        with assert_executed_plan_count(3):
+            # Since BodoSeries cannot have mixed dtypes, BodoSeries.describe casts all elements to string.
+            # Applying map(str) to pandas output is a workaround to enable_test_equal to compare values of differing dtypes.
+            describe_pd = df[c].describe().map(str)
```
scott-routledge2, line 2347: Maybe instead of casting everything to string and returning bodoseries, we can just return a Pandas series of mixed types? Looks like we already return Pandas Series so for numeric describe (they should return the same type in both cases). Since it's a small Series I don't think it matters a lot, and we can add a note in the API docs that this executes some plans.

scott-routledge2 (commented):
Thanks @chrisoh7, see my comments.

scott-routledge2 (approved):

PR #718, opened 2025-07-31
BSE-5033: Add spawn process on workers

IsaacWarren:

As title, adds method to start and stop processes on spawned workers

## Testing strategy
New test

## User facing changes
N/A, will add docs/example when the embedding/generation are merged

## Checklist

### DIFF
--- a/bodo/spawn/worker.py
+++ b/bodo/spawn/worker.py
@@ -483,10 +483,12 @@ def handle_spawn_process(
     logger: logging.Logger,
 ):
     """Handle spawning a new process and return the process handle"""
-    debug_worker_msg(logger, f"Spawning process with command {command}")
     pid = None
     popen = None
-    if bodo.get_rank() not in bodo.libs.distributed_api.get_nodes_first_rank():
+    if bodo.get_rank() not in bodo.libs.distributed_api.get_nodes_first_ranks(
+        comm_world
+    ):
+        debug_worker_msg(logger, f"Spawning process with command {command}")
         popen = subprocess.Popen(
             command,
             env=env,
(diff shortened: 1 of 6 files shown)


bodo/tests/test_spawn/test_spawn_mode.py
```diff
@@ -464,3 +466,42 @@ def test_spawn_jupyter_worker_output_redirect():
...
+    proc = bodo.spawn_process_on_workers(["python", "-c", "print(0)"])
+    bodo.stop_process_on_workers(proc)
+    assert proc._rank_to_pid is not None, "Process should have a rank to PID mapping"
+    assert isinstance(proc._rank_to_pid, dict), (
+        "Rank to PID mapping should be a dictionary"
+    )
+    assert len(proc._rank_to_pid) == 1, (
+        "There should be one process spawned on the worker since we test on one node"
+    )
+    time.sleep(1)  # Give some time for the process to close
```
ehsantn, line 488: I wish we could wait on something instead of sleep calls to avoid flakiness but this is probably good enough.

bodo/spawn/worker.py
```diff
@@ -472,6 +475,81 @@ def exec_func_handler(
...
+    command: str | list[str],
+    env: dict[str, str],
+    cwd: str | None,
+    comm_world: MPI.Intracomm,
+    logger: logging.Logger,
+):
+    """Handle spawning a new process and return the process handle"""
+    pid = None
+    popen = None
+    if bodo.get_rank() not in bodo.libs.distributed_api.get_nodes_first_ranks(
```
ehsantn: Only non-first rank workers start a worker process? Shouldn't it be only one rank per node?

bodo/tests/test_spawn/test_spawn_mode.py
IsaacWarren: Yeah, I agree. I think it's actually dead by the time it returns but psutil has an internal cache so it doesn't see it instantly

bodo/spawn/worker.py
IsaacWarren: Thank you for catching! I'm not sure how that not got in there. Yes it should only be one rank per node

scott-routledge2 (approved):
LGTM, thanks @IsaacWarren !

PR #721, opened 2025-08-01
Set batch size on Nightly

scott-routledge2:

We set batch size to 32768 in the DataFrame library to get better performance, however some tests in Bodo/BodoSQL testing specific behavior are tuned to a different batch size (4096), so we can keep the tests the same by setting batch size.

## Checklist


IsaacWarren (approved):
Would it be hard to only mark the affected tests? If so, no worries

scott-routledge2:
> Would it be hard to only mark the affected tests? If so, no worries

I think we set BODO_STREAMING_BATCH_SIZE at import time, but we could use the default batch size for the DataFrame Library tests?

IsaacWarren:
> > Would it be hard to only mark the affected tests? If so, no worries
> 
> I think we set BODO_STREAMING_BATCH_SIZE at import time, but we could use the default batch size for the DataFrame Library tests?

Good point, either way

.github/workflows/_test_python_source.yml
```diff
@@ -129,6 +129,7 @@ jobs:
             BODO_SPAWN_MODE: "0"
             BODO_BUFFER_POOL_REMOTE_MODE: "1"
             BODO_BUFFER_POOL_DEBUG_MODE: "1"
+            BODO_STREAMING_BATCH_SIZE: "4096"
```
ehsantn: Let's just add a comment here to explain why.

ehsantn (approved):

PR #725, opened 2025-08-04
Add cfunc support for apply

scott-routledge2:

Adds option for apply to use cfunc. 
Also fixes bug in map/apply where UDFs on certain timestamp types were failing with a runtime error. 

I ran a small benchmark using the geo_hash function/nyc taxi, `engine=bodo` was roughly 10x faster on 2 workers. However, initialization is much slower than python (we saw this with map too)

## Testing strategy

PR CI, local benchmark (12 million rows):

engine=bodo: 11.366211s
engine=python: 99.091960s

``` py
""" Run UDF on NYC Taxi Data. Uses dataset from:
https://www.kaggle.com/datasets/elemento/nyc-yellow-taxi-trip-data
usage:
    python geohash_bodo.py --path PATH --use-bodo
"""

import time
import argparse
import bodo.pandas as pd


# geohash encode impl adapted from https://github.com/vinsci/geohash/blob/master/Geohash/geohash.py
#  Note: the alphabet in geohash differs from the common base32
#  alphabet described in IETF's RFC 4648
#  (http://tools.ietf.org/html/rfc4648)
__base32 = '0123456789bcdefghjkmnpqrstuvwxyz'
__decodemap = {}
for i in range(len(__base32)):
    __decodemap[__base32[i]] = i
del i


def encode(pos, precision=12):
    """
    Encode a position given in float arguments latitude, longitude to
    a geohash which will have the character count precision.
    """
    latitude, longitude = pos.pickup_latitude, pos.pickup_longitude

    lat_interval, lon_interval = (-90.0, 90.0), (-180.0, 180.0)
    geohash = ["__bodo_dummy__"]
    bits = [ 16, 8, 4, 2, 1 ]
    bit = 0
    ch = 0
    even = True
    while len(geohash) < precision:
        if even:
            mid = (lon_interval[0] + lon_interval[1]) / 2
            if longitude > mid:
                ch |= bits[bit]
                lon_interval = (mid, lon_interval[1])
            else:
                lon_interval = (lon_interval[0], mid)
        else:
            mid = (lat_interval[0] + lat_interval[1]) / 2
            if latitude > mid:
                ch |= bits[bit]
                lat_interval = (mid, lat_interval[1])
            else:
                lat_interval = (lat_interval[0], mid)
        even = not even
        if bit < 4:
            bit += 1
        else:
            geohash.append(__base32[ch])
            bit = 0
            ch = 0
    return ''.join(geohash[1:])


def apply_test(df, func, use_bodo=True):
    """ Apply UDF to DataFrame and return result. """
    engine = "bodo" if use_bodo else "python"

    start_apply = time.perf_counter()
    result = df.apply(func, axis=1, engine=engine).execute_plan()
    end_apply = time.perf_counter()
    print("apply with {} engine took: {:2f}s".format("bodo" if use_bodo else "python", end_apply - start_apply))

    return result


def main():
    """ Read NYC taxi data from CSV file and apply UDF on columns. """
    parser = argparse.ArgumentParser(description="Run UDF on NYC Taxi Data.")
    parser.add_argument('--path', type=str, required=True, help='Path to NYC taxi data (CSV).')
    parser.add_argument('--use-bodo', action='store_true', help='Whether to use Bodo as UDF engine.')

    args = parser.parse_args()

    cols = ["pickup_latitude", "pickup_longitude"]

    start_read = time.time()
    coords_df = pd.read_csv(args.path, usecols=cols)
    print(type(coords_df))
    end_read = time.time()
    print("read took: {:2f}".format(end_read - start_read))

    geohash_pickup = apply_test(coords_df, encode, use_bodo=args.use_bodo)

    print(geohash_pickup.head(10))


if __name__ == "__main__":
    main()
```

## User facing changes

Apply might be faster by default, users have the option of specifying `engine="bodo"` or `engine="python"`, default is "bodo"

## Checklist


bodo/pandas/utils.py
```diff
@@ -975,8 +1001,8 @@ def get_scalar_udf_result_type(obj, method_name, func, *args, **kwargs) -> pd.Se
 
     except_msg = ""
     for sample_size in sample_sizes:
-        df_sample = obj.head(sample_size).execute_plan()
```
scott-routledge2, line 978: I am pretty sure this was a bug since head(sample_size) might already be a pandas object (e.g. if it is the output of JIT)

docs/docs/api_docs/dataframe_lib/series/map.md
```diff
@@ -17,13 +17,16 @@ from within the execution pipeline are avoided.
...
-    
+
 <p class="api-header">Parameters</p>
 
 : __arg : *function, collections.abc.Mapping subclass or Series*:__ Mapping correspondence.  *function* may be a Python function or a dispatcher generated through numba.jit or bodo.jit.
 
 : __na_actions : *{None, ‘ignore’}, default None*:__ If 'ignore' then NaN values will be propagated without passing them to the mapping correspondence.
 
+: __engine : *{'bodo', 'python'}, default 'bodo'*:__  The engine to use to compute the UDF. By default, engine='bodo' will apply bodo.jit
+to `arg` with fallback to python as described above. Use engine='python' to avoid any jit compilation.
```
DrTodd13:
```suggestion
to `arg` with fallback to Python as described above. Use engine='python' to avoid any jit compilation.
```

docs/docs/api_docs/dataframe_lib/dataframe/apply.md
```diff
@@ -33,6 +42,9 @@ See [`pandas.DataFrame.apply`](https://pandas.pydata.org/docs/reference/api/pand
 
 : __args : *tuple*:__ Additional positional arguments to pass to *func*.
 
+: __engine : *{'bodo', 'python', 'numba'}, default 'bodo'*:__ The engine to use to compute the UDF. By default, `engine="bodo"` will apply bodo.jit
+to `func` with fallback to python described above. Use engine='python' to avoid any jit compilation. `engine='numba'` will trigger a fall back to [`pandas.DataFrame.apply`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.apply.html#pandas.DataFrame.apply).
```
DrTodd13:
```suggestion
to `func` with fallback to Python described above. Use engine='python' to avoid any jit compilation. `engine='numba'` will trigger a fall back to [`pandas.DataFrame.apply`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.apply.html#pandas.DataFrame.apply).
```

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -699,6 +702,52 @@ def test_apply(datapath, index_val):
...
+                "c": ["a", "b", "c"] * 10,
+            },
+            index=index_val[:30],
+        )
+        bdf = bd.from_pandas(df)
+
+        def unknown_func(x):
+            return x + 10
+
+        def apply_func(row):
```
DrTodd13, line 742: Do we document anywhere what you have to do if you have a Python function you pass to apply and it calls some other function and you want it to be JIT compiled?  In other words, register_jitable on unknown_func?  Perhaps nice if all such functions encountered were recursively compiled but I don't think anyone does that due to the difficulty.

bodo/tests/test_dataframe_part2.py
```diff
@@ -587,7 +587,8 @@ def test_impl(df):
     check_func(test_impl, (df,))
 
 
-@pytest.mark.df_lib
+# TODO (df_lib): properly support RangeIndex in apply
```
DrTodd13: Small preference for putting this TODO after the pytest.mark.df_lib on the same line so they are clearly related.

bodo/utils/conversion.py
```diff
@@ -108,11 +108,14 @@ def generic(self, args, kws):
...
                 isinstance(data.dtype, (types.Boolean, types.Integer, types.Float))
                 or data.dtype == bodo.timedelta64ns
+                or data.dtype == bodo.datetime64ns
             ):
                 if data.dtype == types.bool_:
                     output = bodo.boolean_array_type
                 elif data.dtype == bodo.timedelta64ns:
                     output = bodo.timedelta_array_type
+                elif data.dtype == bodo.datetime64ns:
+                    output = bodo.DatetimeArrayType(None)
```
IsaacWarren, line 118: What is None in this context?

IsaacWarren (approved):
Thanks Scott

DrTodd13 (approved):
Looks pretty good.  Thanks.

bodo/tests/test_df_lib/test_end_to_end.py
scott-routledge2: I think it should be in the docs but let me look. Maybe we should link to a how to use JIT page in the apply docs?

bodo/utils/conversion.py
scott-routledge2: It's for the timezone, I don't think datetime64 stores timezone information?

bodo/tests/test_df_lib/test_end_to_end.py
scott-routledge2: We don't have anything in the docs. Will add in a followup PR

PR #729, opened 2025-08-04
BSE-5036: Add llm_generate

IsaacWarren:
## Checklist

### DIFF
--- a/bodo/pandas/series.py
+++ b/bodo/pandas/series.py
@@ -789,6 +789,56 @@ class BodoSeries(pd.Series, BodoLazyWrapper):
             {},
         )
 
+    def map_partitions(self, func, *args, **kwargs):
+        """
+        Apply a function to each partition of the series.
+
+        If self is a lazy plan, then the result will also be a lazy plan
+        (assuming result is Series and the dtype can be infered). Otherwise, the lazy
+        plan will be evaluated.
+        NOTE: this pickles the function and sends it to the workers, so globals are
+        pickled. The use of lazy data structures as globals causes issues.
+
+        Args:
+            func (Callable): A callable which takes in a DataFrame as its first
+                argument and returns a DataFrame or Series that has the same length
+                its input.
+            *args: Additional positional arguments to pass to func.
+            **kwargs: Additional key-word arguments to pass to func.
+
+        Returns:
+            DataFrame or Series: The result of applying the func.
+        """
+        import bodo.spawn.spawner
+
+        if self._exec_state == ExecState.PLAN:
+            required_fallback = False
+            try:
+                empty_series = get_scalar_udf_result_type(
+                    self, "map_partitions", func, *args, **kwargs
+                )
+            except BodoLibNotImplementedException as e:
+                required_fallback = True
+                msg = (
+                    f"map_paritions(): encountered exception: {e}, while trying to "
+                    "build lazy plan. Executing plan and running map_paritions on "
+                    "workers (may be slow or run out of memory)."
+                )
+                warnings.warn(BodoLibFallbackWarning(msg))
+
+                self_arg = self.execute_plan()
+
+            if not required_fallback:
+                return _get_series_python_func_plan(
+                    self._plan, empty_series, func, args, kwargs
+                )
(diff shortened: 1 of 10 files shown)


IsaacWarren:
@ehsantn and @DrTodd13 I'm looking for early reviews for this approach to accelerate merging while I work on embeddings which should be largely the same and documentation

bodo/pandas/series.py
```diff
@@ -789,6 +789,56 @@ def map_with_state(self, init_state_fn, row_fn, na_action=None, output_type=None
             {},
         )
 
+    def map_partitions(self, func, *args, **kwargs):
+        """
+        Apply a function to each partition of the series.
+
+        If self is a lazy plan, then the result will also be a lazy plan
+        (assuming result is Series and the dtype can be infered). Otherwise, the lazy
+        plan will be evaluated.
```
DrTodd13: If self is not a lazy plan then the result will not be a lazy plan either?

bodo/pandas/series.py
```diff
@@ -789,6 +789,56 @@ def map_with_state(self, init_state_fn, row_fn, na_action=None, output_type=None
...
+        NOTE: this pickles the function and sends it to the workers, so globals are
+        pickled. The use of lazy data structures as globals causes issues.
+
+        Args:
+            func (Callable): A callable which takes in a DataFrame as its first
```
DrTodd13:
```suggestion
            func (Callable): A callable which takes in a Series as its first
```

bodo/pandas/series.py
```diff
@@ -789,6 +789,56 @@ def map_with_state(self, init_state_fn, row_fn, na_action=None, output_type=None
...
+                argument and returns a DataFrame or Series that has the same length
```
DrTodd13, line 805:
```suggestion
                argument and returns a DataFrame or Series that has the same length
```
```suggestion
                argument and returns a Series that has the same length
```

bodo/pandas/series.py
```diff
@@ -789,6 +789,56 @@ def map_with_state(self, init_state_fn, row_fn, na_action=None, output_type=None
...
+                its input.
+            *args: Additional positional arguments to pass to func.
+            **kwargs: Additional key-word arguments to pass to func.
+
+        Returns:
+            DataFrame or Series: The result of applying the func.
```
DrTodd13, line 811:
```suggestion
            Series: The result of applying the func.
```

bodo/pandas/series.py
IsaacWarren: It can produce a dataframe or series

bodo/pandas/series.py
IsaacWarren: I just copied this from  DataFrame.map_partitions, @ehsantn do you remember the semantics?

bodo/pandas/series.py
```diff
@@ -1430,6 +1480,57 @@ def per_row(tokenizer, row):
...
+            )
+        if model is not None:
+            generation_kwargs["model"] = model
+
+        def map_func(series, api_token, endpoint, generation_kwargs):
+            import asyncio
+
+            import openai
+
+            client = openai.AsyncOpenAI(
```
ehsantn, line 1510: `map_with_state` may be useful if this is expensive. Not necessary if the actual calls are a lot more expensive.

IsaacWarren: This should only be done once per partition, so once per worker/rank

ehsantn: I assume the map function is called once per batch, not once per rank.

IsaacWarren: Oh yeah you're right

DrTodd13: Yeah, that's why I added map_partition_with_state because you want to process as a batch and do init once per worker.  I'll probably get to that function tomorrow.

bodo/pandas/series.py
ehsantn: If self is not lazy, we just run the function on partitions with `submit_func_to_workers` and return the parallel output Series based on the code.

PR #735, opened 2025-08-06
Try JIT before falling back to Python.

DrTodd13:

We do 1) native dataframe library and if that fails now 2) JIT fallback and if that fails 3) Python.

## Testing strategy

run_ci

## User facing changes

None

## Checklist

### DIFF
--- a/bodo/pandas/utils.py
+++ b/bodo/pandas/utils.py
@@ -1232,3 +1232,114 @@ def fallback_wrapper(self, attr):
         return silenced_method
 
     return attr
+
+
+class JITFallback:
+    # Holds a mapping of a tuple of class name and function name to either
+    # False to say that compilation previously failed for that function or
+    # a callable dispatcher for the JIT compiled version of that function.
+    fallback_cache = {}
+    compile_success = 0
+    compile_fail = 0
+    after_success = 0
+    python_fallback = 0
+
+    class JITFallbackFail(Exception):
+        pass
+
+    def __init__(self, base_obj, name):
+        self.base_obj = base_obj if base_obj is not pd else None
+        self.name = name
+
+    def __call__(self, *args, **kwargs):
+        key = (
+            (
+                self.base_obj.__class__.__name__,
+                self.name,
+                *[type(x) for x in args],
+                *list(kwargs.keys()),
+                *[type(x) for x in kwargs.values()],
+            ),
+            None,
+        )
+        # See if we previously tried to compile this function.
+        cache_entry = JITFallback.fallback_cache.get(key, None)
+        if (
+            self.name in ("duplicated")
+            # not in (
+            #    "items",
+            #    "set_index",
+            #    "rename_axis",
+            #    "copy",
+            #    "keys",
+            #    "apply",
+            #    "groupby",
+            # )
+            and cache_entry != False
+        ):
+            # None means it wasn't in the cache either way so we can try to
+            # JIT compile it.
+            if cache_entry is None:
+                if self.base_obj is None:
+                    from bodo.numba_compat import is_func_overloaded
+
+                    # Do a better check here if this function appears overloaded.
+                    jit_supported = is_func_overloaded("pandas", self.name)
+                else:
+                    overloads = get_overloads(self.base_obj.__class__.__name__)
+                    jit_supported = self.name in overloads
+
+                if jit_supported:
+                    fname = f"bodo_jitfallback_{self.name}"
+                    self_arg = "self, " if self.base_obj is not None else ""
+                    sig_args = ",".join(
+                        [f"arg{i}" for i in range(len(args))] + list(kwargs.keys())
+                    )
+                    caller_args = ",".join(
+                        [f"arg{i}" for i in range(len(args))]
+                        + [f"{x}={x}" for x in kwargs.keys()]
+                    )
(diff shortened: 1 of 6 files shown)


bodo/pandas/utils.py
```diff
@@ -1219,3 +1218,129 @@ def silenced_method(*args, **kwargs):
...
+                *[type(x) for x in kwargs.values()],
+            ),
+            None,
+        )
+        # See if we previously tried to compile this function.
+        cache_entry = JITFallback.fallback_cache.get(key, None)
+        if (
+            self.name
+            not in (
+                "items",
```
scott-routledge2: Why does JIT fallback not work for these functions?

bodo/pandas/utils.py
```diff
@@ -1219,3 +1218,129 @@ def silenced_method(*args, **kwargs):
...
+                else:
+                    cm_args = (self.base_obj, *args, *tuple(kwargs.values()))
+                return cache_entry(*cm_args)
+
+        JITFallback.python_fallback += 1
+        msg = (
+            f"{self.name} is not implemented in Bodo Dataframe Library yet. "
+            "Falling back to Pandas (may be slow or run out of memory)."
+        )
+        warnings.warn(BodoLibFallbackWarning(msg))
```
scott-routledge2: This might cause unnecessary warnings like in TPCH `pd.Timestamp("1994-03-01")`. Wondering if there is a way to detect whether a function is performance critical and only warn in that case?

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -325,6 +329,7 @@ def test_filter_pushdown(datapath, file_path, op):
         check_pandas_types=False,
         sort_output=True,
         reset_index=False,
+        check_names=False,
```
ehsantn: Why is this changed? How would this PR impact output names?

ehsantn (approved):
Thanks @DrTodd13! Looks good to me overall. I think some corner cases may pop up from time to time that we just need to handle later.

bodo/tests/test_df_lib/test_end_to_end.py
DrTodd13: Was necessary with previous version of code but not the re-worked code so I've reverted this.

bodo/pandas/utils.py
```diff
@@ -1232,3 +1239,102 @@ def silenced_method(*args, **kwargs):
...
+                if self.base_obj is None:
+                    cm_args = args + tuple(kwargs.values())
+                else:
+                    cm_args = (self.base_obj, *args, *tuple(kwargs.values()))
+                return cache_entry(*cm_args)
+
+        JITFallback.python_fallback += 1
+        raise JITFallback.JITFallbackFail()
+
+        # msg = (
```
ehsantn: Remove commented out code?

bodo/pandas/utils.py
```diff
@@ -1232,3 +1239,102 @@ def silenced_method(*args, **kwargs):
...
+                        [f"arg{i}" for i in range(len(args))]
+                        + [f"{x}={x}" for x in kwargs.keys()]
+                    )
+                    func_text = f"def {fname}({self_arg}{sig_args}):\n"
+                    if self.base_obj is None:
+                        func_text += f"    return pd.{self.name}({caller_args})\n"
+                    else:
+                        func_text += f"    return self.{self.name}({caller_args})\n"
+
+                    # print("new_func_text:", func_text)
```
ehsantn: Remove

bodo/pandas/utils.py
```diff
@@ -1232,3 +1239,102 @@ def silenced_method(*args, **kwargs):
...
+                self.name,
+                *[type(x) for x in args],
+                *list(kwargs.keys()),
+                *[type(x) for x in kwargs.values()],
+            ),
+            None,
+        )
+        # See if we previously tried to compile this function.
+        cache_entry = JITFallback.fallback_cache.get(key, None)
+        if self.name in ("duplicated", "pivot") and cache_entry != False:
```
ehsantn, line 1280: Is `self.name in ("duplicated", "pivot")` debug code that needs removed? Or is it related to unit testing?

ehsantn (approved):
Looks good to me. Thanks @DrTodd13

bodo/pandas/utils.py
DrTodd13: done

bodo/pandas/utils.py
DrTodd13: done

bodo/pandas/utils.py
DrTodd13: This is the whitelist I mentioned.  Scott had a good point about too many warnings and performance critical functions so I turned off the warning for non-callable attributes of the class.  There are some routines for which the JIT fallback function runs but then the output of that is of some form that breaks things downstream.  We don't know what we don't know so even if I excluded things I don't think work then there are surely untested things that also don't work which will break our claims of seamless auto-fallback.  We can revisit later.

ehsantn: Makes sense. Let's just add some comments in the code and make the list explicit.

scott-routledge2 (approved):
LGTM, thanks @DrTodd13

PR #749, opened 2025-08-11
[WIP] Groupby.agg UDF support

scott-routledge2:
## Checklist

### DIFF
--- a/bodo/libs/groupby/_groupby_col_set.cpp
+++ b/bodo/libs/groupby/_groupby_col_set.cpp
@@ -1487,6 +1487,9 @@ void UdfColSet::alloc_update_columns(
     if (this->combine_step) {
         offset = 1;
     }
+
+    std::cout << "allocating update column: " << udf_table_idx << std::endl;
+
     // for update table we only need redvars (skip first column which is
     // output column)
     for (int i = udf_table_idx + offset; i < udf_table_idx + 1 + n_redvars;
@@ -1556,6 +1559,13 @@ GeneralUdfColSet::GeneralUdfColSet(std::shared_ptr<array_info> in_col,
     : UdfColSet(in_col, false, udf_table, udf_table_idx, 0, use_sql_rules) {}
 
 GeneralUdfColSet::~GeneralUdfColSet() = default;
+
+std::unique_ptr<bodo::Schema> GeneralUdfColSet::getRunningValueColumnTypes(
+    const std::shared_ptr<bodo::Schema>& in_schema) const {
+    std::vector<int> col_idxs;
+    col_idxs.push_back(udf_table_idx);
+    return udf_table->schema()->Project(col_idxs);
+}
 
 void GeneralUdfColSet::fill_in_columns(
     const std::shared_ptr<table_info>& general_in_table,
--- a/bodo/libs/groupby/_groupby_col_set.h
+++ b/bodo/libs/groupby/_groupby_col_set.h
@@ -1088,27 +1088,27 @@ class UdfColSet : public BasicColSet {
               std::shared_ptr<::arrow::MemoryManager> mm =
                   bodo::default_buffer_memory_manager()) override;
 
-    virtual void setUpdateCols(
-        std::vector<std::shared_ptr<array_info>> update_cols_) override {
-        throw std::runtime_error(
-            "UDFColSet not implemented for streaming groupby");
-    }
-    virtual void setCombineCols(
+    // void setUpdateCols(
+    //     std::vector<std::shared_ptr<array_info>> update_cols_) override {
+    //     throw std::runtime_error(
+    //         "UDFColSet not implemented for streaming groupby");
+    // }
+    void setCombineCols(
         std::vector<std::shared_ptr<array_info>> combine_cols_) override {
         throw std::runtime_error(
             "UDFColSet not implemented for streaming groupby");
     }
-    virtual void setInCol(
-        std::vector<std::shared_ptr<array_info>> new_in_cols) override {
-        throw std::runtime_error(
-            "UDFColSet not implemented for streaming groupby");
-    }
-    virtual void clear() override {
+    // void setInCol(
+    //     std::vector<std::shared_ptr<array_info>> new_in_cols) override {
+    //     throw std::runtime_error(
+    //         "UDFColSet not implemented for streaming groupby");
+    // }
+    void clear() override {
         throw std::runtime_error(
             "UDFColSet not implemented for streaming groupby");
     }
 
-   private:
+   protected:
     const std::shared_ptr<table_info>
         udf_table;            // the table containing type info for UDF columns
     const int udf_table_idx;  // index to my information in the udf table
@@ -1133,6 +1133,15 @@ class GeneralUdfColSet : public UdfColSet {
      */
     void fill_in_columns(const std::shared_ptr<table_info>& general_in_table,
                          const grouping_info& grp_info) const;
+
+    std::unique_ptr<bodo::Schema> getRunningValueColumnTypes(
+        const std::shared_ptr<bodo::Schema>& in_schema) const override;
+
+    // clear state of all general UDFs at the same time
+    void clear() override {}
+
+    // clear state once all col sets have been processed
+    void clear_after_gen_udf() { BasicColSet::clear(); }
 };
 
 /**
--- a/bodo/libs/streaming/_groupby.cpp
+++ b/bodo/libs/streaming/_groupby.cpp
@@ -3,9 +3,12 @@
 #include <fmt/format.h>
 #include <mpi.h>
 #include <algorithm>
+#include <cassert>
 #include <cstring>
 #include <list>
 #include <memory>
+#include <optional>
+#include <sstream>
 #include <tuple>
 #include "../_array_hash.h"
 #include "../_array_operations.h"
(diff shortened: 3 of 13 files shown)


bodo/libs/groupby/_groupby_col_set.cpp
```diff
@@ -1576,6 +1579,78 @@ void GeneralUdfColSet::fill_in_columns(
     }
 }
 
+// ############################## StreaminglUDF ##############################
```
IsaacWarren:
```suggestion
// ############################## StreamingUDF ##############################
```

bodo/libs/groupby/_groupby_col_set.cpp
```diff
@@ -1576,6 +1579,78 @@ void GeneralUdfColSet::fill_in_columns(
...
+      udf_table(std::move(udf_table)),
+      udf_table_idx(udf_table_idx),
+      func(func) {}
+
+StreamingUDFColSet::~StreamingUDFColSet() = default;
+
+std::unique_ptr<bodo::Schema> StreamingUDFColSet::getRunningValueColumnTypes(
+    const std::shared_ptr<bodo::Schema>& in_schema) const {
+    std::vector<int> col_idxs;
+    col_idxs.push_back(udf_table_idx);
```
IsaacWarren: Small nit
```suggestion
    std::vector<int> col_idxs = {udf_table_idx}
```

bodo/libs/groupby/_groupby_col_set.h
```diff
@@ -1135,6 +1136,50 @@ class GeneralUdfColSet : public UdfColSet {
...
+    void update(const std::vector<grouping_info>& grp_infos,
+                bodo::IBufferPool* const pool = bodo::BufferPool::DefaultPtr(),
+                std::shared_ptr<::arrow::MemoryManager> mm =
+                    bodo::default_buffer_memory_manager()) override;
+
+   private:
+    const std::shared_ptr<table_info>
+        udf_table;            // the table containing type info for UDF columns
+    const int udf_table_idx;  // index to my information in the udf table
+    stream_udf_t* func;  // A ccallback for computing the UDF on a single group.
```
IsaacWarren:
```suggestion
    stream_udf_t* func;  // A callback for computing the UDF on a single group.
```
Can this be const?

bodo/libs/groupby/_groupby_ftypes.h
```diff
@@ -101,7 +101,9 @@ struct Bodo_FTypes {
         // more explicit ftypes. This is used only in the min_row_number_filter
         // window function path.
         idx_n_columns = 70,
-        n_ftypes = 71,
+        // Streaming UDFs for Bodo DataFrames groupby.agg
+        stream_udf = 71,
```
IsaacWarren, line 105: I think this needs added to aggregate.py

bodo/pandas/groupby.py
```diff
@@ -491,7 +536,115 @@ def _get_aggfunc_str(func):
...
+                )
+                return out_arr
+        else:
+
+            @numba.njit
+            def coerce_to_array_impl(scalar):
+                out_arr = coerce_scalar_to_array(
+                    scalar, 1, out_col_type, dict_encode=False
+                )
+                # coerce_scalar_to_array doesn't respect nullable types, so using coerce_to_array
```
IsaacWarren, line 577: How hard would this be to fix?

bodo/pandas/physical/aggregate.h
```diff
@@ -24,6 +27,33 @@ struct PhysicalAggregateMetrics {
...
+ * @param cfunc_wrapper Python function that takes no arguments, called on
+ * workers to compile a cfunc for computing a UDF and expose its address.
+ * @return stream_udf_t* cfunc for applying UDFs on a grouped table.
+ */
+stream_udf_t* get_cfunc_from_wrapper(PyObject* cfunc_wrapper) {
+    if (cfunc_wrapper == Py_None) {
+        throw std::runtime_error("agg: No cfunc_wrapper found for aggfunc.");
+    }
+
+    PyObject* result = PyObject_CallNoArgs(cfunc_wrapper);
```
IsaacWarren, line 43: Does result need to be DECREFED?

IsaacWarren (approved):
Thanks Scott!

bodo/pandas/groupby.py
scott-routledge2: Probably not that hard, we could add a `use_nullable_types` similar to `coerce_to_array`

bodo/pandas/physical/aggregate.h
scott-routledge2: Yes, thanks

bodo/libs/groupby/_groupby_ftypes.h
scott-routledge2: I don't think it needs to be added to `aggregate.py` since it doesn't have an equivalent in that codepath? I added it in the C++ only section of the definition

bodo/pandas/groupby.py
```diff
@@ -22,14 +26,54 @@
...
+    "nunique",
+}
+
+
+class GroupbyAggFunc:
+    """Stores data for computing aggfuncs."""
+
+    in_col: str
+    func: pt.Callable | str  # The actual function
+    func_name: str  # The function name as it will appear int the output.
```
ehsantn:
```suggestion
    func_name: str  # The function name as it will appear in the output.
```

bodo/pandas/groupby.py
```diff
@@ -416,13 +455,18 @@ def _groupby_agg_plan(
     grouped_selection = grouped.selection_for_plan
 
     zero_size_df = _empty_like(grouped._obj)
-    empty_data_pandas = zero_size_df.groupby(grouped._keys, as_index=grouped._as_index)[
+
+    # Convert to Pandas types to avoid gaps in Arrow conversion
+    zero_size_df_pandas = convert_to_pandas_types(zero_size_df)
+    empty_data_pandas = zero_size_df_pandas.groupby(
```
ehsantn, line 461: Do we need to cast the output back to Arrow dtypes or `_cast_groupby_agg_columns` covers it?

bodo/libs/groupby/_groupby_col_set.h
scott-routledge2: My IDE complained when I added the const qualifier that it didn't have an effect since it is a function type.

bodo/libs/streaming/_groupby.cpp
```diff
@@ -2654,7 +2656,7 @@ GroupbyState::GroupbyState(
             ftype == Bodo_FTypes::shift || ftype == Bodo_FTypes::transform ||
             ftype == Bodo_FTypes::ngroup || ftype == Bodo_FTypes::window ||
             ftype == Bodo_FTypes::listagg || ftype == Bodo_FTypes::nunique ||
-            ftype == Bodo_FTypes::head || ftype == Bodo_FTypes::gen_udf ||
+            ftype == Bodo_FTypes::head || ftype == Bodo_FTypes::stream_udf ||
```
ehsantn, line 2659: Why remove `gen_udf`?

bodo/libs/streaming/_groupby.cpp
```diff
@@ -2812,12 +2819,17 @@ GroupbyState::GroupbyState(
             }
 
             std::unique_ptr<bodo::Schema> running_values_schema =
-                this->getRunningValueColumnTypes(local_input_cols,
-                                                 std::move(in_arr_types_copy),
-                                                 ftypes[i], 0, window_args);
+                this->getRunningValueColumnTypes(
+                    local_input_cols, std::move(in_arr_types_copy), ftypes[i],
+                    0, window_args, udf_out_types, udf_idx);
```
ehsantn: Maybe we can pass the corresponding array from `udf_out_types` for cleaner API?

bodo/libs/groupby/_groupby_col_set.cpp
```diff
@@ -1576,6 +1579,77 @@ void GeneralUdfColSet::fill_in_columns(
...
+    }
+
+    // get the rows in each group
+    for (size_t i = 0; i < in_col->length; i++) {
+        int64_t i_grp = grp_info.row_to_group[i];
+        group_rows[i_grp].push_back(i);
+    }
+
+    // TODO: make out_arrs a bodo::vector for large number of groups case
+    std::vector<std::shared_ptr<array_info>> out_arrs(grp_info.num_groups);
```
ehsantn, line 1625: Maybe we can have something like a table builder and append to it instead of accumulating like this.

ehsantn (approved):
Thanks for the effort @scott-routledge2! Looks good to me overall. The input column matching may be possible to improve but not a big deal right now (may be done as part of apply actually).

bodo/libs/streaming/_groupby.cpp
scott-routledge2: `gen_udf` is not used anywhere else in streaming, it was a little misleading to me.

bodo/libs/streaming/_groupby.cpp
scott-routledge2: I agree, but the `udf_out_types`, `udf_idx` was already in the makeColSet API which is why I reused here.

bodo/pandas/groupby.py
scott-routledge2: Yes `_cast_groupby_agg_columns` converts types back to arrow. Pandas agg is used here for generating the column names and validating the shape of `func`

PR #763, opened 2025-08-18
BSE-5047: Optimize Quantile

IsaacWarren:

This PR templates PhysicalQuantile on the type being aggregated so we can avoid casting every value to a common type.
This is roughly 2x faster on a simple local benchmark with 4 workers, although it is still nowhere near pandas' performance on small data.
```py
import time

import numpy as np

import bodo
import bodo.pandas as bpd


def test_series_describe():
    bodo.spawn.spawner.get_spawner()
    series = bpd.Series(np.arange(1000000))
    start_time = time.time()
    b_result = series.describe()
    print(b_result)
    end_time = time.time()
    print(f"Bodo pandas describe took {end_time - start_time:.4f} seconds")
test_series_describe()

```

## Testing strategy
Existing unit test

## User facing changes
Additional type checking on Series.Describe

## Checklist

### DIFF
--- a/bodo/pandas/_physical_conv.cpp
+++ b/bodo/pandas/_physical_conv.cpp
@@ -171,22 +171,23 @@ void PhysicalPlanBuilder::Visit(duckdb::LogicalAggregate& op) {
             std::make_shared<PipelineBuilder>(physical_op);                 \
         return;                                                             \
     }
-            CREATE_QUANTILE(Bodo_CTypes::CTypeEnum::INT8);
-            CREATE_QUANTILE(Bodo_CTypes::CTypeEnum::INT16);
-            CREATE_QUANTILE(Bodo_CTypes::CTypeEnum::INT32);
-            CREATE_QUANTILE(Bodo_CTypes::CTypeEnum::INT64);
-            CREATE_QUANTILE(Bodo_CTypes::CTypeEnum::UINT8);
-            CREATE_QUANTILE(Bodo_CTypes::CTypeEnum::UINT16);
-            CREATE_QUANTILE(Bodo_CTypes::CTypeEnum::UINT32);
-            CREATE_QUANTILE(Bodo_CTypes::CTypeEnum::UINT64);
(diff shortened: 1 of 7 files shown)


bodo/pandas/_physical_conv.cpp
```diff
@@ -159,13 +159,39 @@ void PhysicalPlanBuilder::Visit(duckdb::LogicalAggregate& op) {
...
+        auto physical_op =                                                  \
+            std::make_shared<PhysicalQuantile<dtype_to_type<dtype>::type>>( \
+                bodo_schema, quantiles);                                    \
+        finished_pipelines.emplace_back(                                    \
+            this->active_pipeline->Build(physical_op));                     \
+        this->active_pipeline =                                             \
+            std::make_shared<PipelineBuilder>(physical_op);                 \
+        return;                                                             \
+    }
+            CREATE_QUANTILE(Bodo_CTypes::CTypeEnum::INT8);
```
DrTodd13: Does it make sense to try to order these on likelihood of occurrence?  e.g., int64, float64 first

IsaacWarren: Good idea. Since it's only done once during plan creation it probably doesn't matter too much but might as well

bodo/pandas/series.py
```diff
@@ -1208,6 +1208,15 @@ def quantile(self, q=0.5, interpolation=lib.no_default):
 
         # Drop Index columns since not necessary for reduction output.
         pa_type = self.dtype.pyarrow_dtype
+        if not (
+            pa_type.is_integer()
+            or pa_type.is_floating()
+            or pa_type.is_decimal()
+            or pa_type.is_temporal()
```
scott-routledge2: Do we have tests for decimal/temporal case?

scott-routledge2 (approved):
LGTM, thanks @IsaacWarren

bodo/pandas/series.py
IsaacWarren: Pandas doesn't work for decimal and the returned type is wrong for temporal, just removing for now

PR #765, opened 2025-08-19
Skip dataframes demo on e2e ci

scott-routledge2:
As titled
## Checklist


DrTodd13 (approved):

scott-routledge2:
> Thanks Scott, Why do we need to skip this?

We need to set up a an open ai key and a test vector bucket/index. Also there are some iceberg components (not sure if that's an issue). I think this would be a really good smoke test for the dataframe library, but would require some more set up.

PR #770, opened 2025-08-20
Split window function code from main module

ehsantn:
Breaks the link between window function code and the main module to reduce the main module size. Also includes a few small changes.


scott-routledge2 (approved):
LGTM, thanks @ehsantn

PR #771, opened 2025-08-21
Revert shared library splitting to avoid caching and package size issues

ehsantn:
As title. Splitting didn't really help import time much and introduced a bunch of problems. Keeps the refactoring code in place and only reverses splitting parts.


scott-routledge2 (approved):
Thanks @ehsantn !

PR #774, opened 2025-08-22
Disable native type inference code

ehsantn:
As title. Saves some package size and minor import time.


scott-routledge2 (approved):
LGTM, thanks @ehsantn

PR #775, opened 2025-08-22
Import JIT Lazily

ehsantn:
## Checklist

### DIFF
--- a/bodo/__init__.py
+++ b/bodo/__init__.py
@@ -242,13 +242,7 @@ import bodo.pandas_compat
 import bodo.numba_compat  # isort:skip
 import numba
 from numba import (  # re-export from Numba
-    gdb,
-    gdb_breakpoint,
-    gdb_init,
-    pndindex,
     prange,
-    stencil,
-    threading_layer,
     typed,
     typeof,
 )


ehsantn:
The top-level files (__init__.py, decorators.py, compiler.py), bodo/spawn, bodo/pandas, parquet_pio.py and fs_io.py need the most attention. A lot of the rest are just replacing bodo.type with bodo.types.type.

This Ruff rule seems useful. Can it be applied to specified files only?

BodoSQL/bodosql/kernels/parse_url_array_kernel.py
```diff
@@ -234,7 +237,7 @@ def parse_url_wrapper_overload(data):
     from urllib.parse import urlparse
 
     def impl(data):  # pragma: no cover
-        with bodo.objmode(
+        with numba.objmode(
```
DrTodd13, line 240: Would the numba import moved here help import times?

bodo/memory.pyx
```diff
@@ -39,4 +37,3 @@ default_buffer_pool_smallest_size_class = bodo.memory_cpp.default_buffer_pool_sm
 default_buffer_pool_bytes_allocated = bodo.memory_cpp.default_buffer_pool_bytes_allocated
 default_buffer_pool_bytes_pinned = bodo.memory_cpp.default_buffer_pool_bytes_pinned
 
-atexit.register(bodo.memory_cpp.default_buffer_pool_cleanup)
```
IsaacWarren, line 42: Where does this happen now?

pyproject.toml
```diff
@@ -159,6 +159,7 @@ exclude = [
 
     # TODO: Skip specific rules for these files
     "bodo/__init__.py",
+    "bodo/types.py",
```
IsaacWarren, line 162: Why do we need this? Can we exclude specific rules since it's new

IsaacWarren (approved):
Thanks Ehsan

bodo/memory.pyx
ehsantn: Here: https://github.com/bodo-ai/Bodo/blob/8ba37bd055b4a365dece371f75fdde1bc397d4c6/bodo/__init__.py#L398

pyproject.toml
ehsantn: To avoid deleting unused imports, reordering etc. The file is just imports so doesn't seem worth it to go through specific rules.

bodo/io/parquet_pio.py
```diff
@@ -684,13 +562,37 @@ def unify_fragment_schema(dataset: ParquetDataset, piece: ParquetPiece, frag):
     # missing from the file will be filled with nulls at read time.
     added_columns = fileset_schema_names - dataset_schema_names
     if added_columns:
+        # Import compiler lazily to access BodoError
+        import bodo.decorators  # isort:skip
```
DrTodd13, line 566: Just kind of weird to import something and not use it.  Is it not possible to overload getattr on bodo.utils.typing and import decorators there if necessary?

bodo/__init__.py
```diff
@@ -237,193 +244,119 @@ def get_sql_config_str() -> str:
...
-    no_warning_objmode,
-)
-from bodo.ir.unsupported_method_template import (
-    overload_unsupported_attribute,
-    overload_unsupported_method,
-)
+def jit(*args, **kwargs):
+    # Import compiler lazily
+    from bodo.decorators import jit as _jit
+    return _jit(*args, **kwargs)
```
scott-routledge2, line 251: Is there a difference between just calling `bodo.jit` versus `import bodo.decorators; bodo.jit...`? Will users have to do `import bodo.decorators` or import compiler on workers for anything?

scott-routledge2 (approved):
LGTM thanks @ehsantn  the import feels a lot snappier now! My one comment was whether we wanted to add top level paths for types to not break existing workloads e.g. `wrap_python(bodo.none)`? (And import the compiler stuff lazily like we do for jit)

DrTodd13 (approved):
Tried a few tests and seems to work nicely.

bodo/__init__.py
ehsantn: Using `bodo.jit` will import the compiler automatically. There is no need for importing `bodo.decorators` manually by users.

ehsantn:
> wrap_python(bodo.none)

I thought about tricks for exporting the data types at the top level without importing the compiler, but the solutions didn't seem great. The main one would be to use a top level getattr function but it may slow down every `bodo.attr` access in CPython. We can discuss it.

bodo/io/parquet_pio.py
ehsantn: We can use a top level getattr function but it may slow down every bodo.attr access in CPython. We can discuss it.

BodoSQL/bodosql/kernels/parse_url_array_kernel.py
ehsantn: What do you mean? Just avoided exporting `objmode` since exporting it without importing the compiler seemed difficult and unnecessary (we have `wrap_python` now which is preferred).

DrTodd13: Just talking about numba import time.  How bad is that?  This file unilaterally imports numba but if import numba moved here then it only imports as needed.  Minor possible improvement.

ehsantn: Numba import is ~0.4s so not small. But this file is in BodoSQL which imports the compiler anyways.

PR #777, opened 2025-08-25
Support setting df.columns attribute

scott-routledge2:

Let's users set columns of BodoDataFrames e.g.
`df.columns = ["C", "D"]` 
(used in TPCH Q8)

## Checklist

### DIFF
--- a/bodo/pandas/plan.py
+++ b/bodo/pandas/plan.py
@@ -44,7 +44,7 @@ class LazyPlan:
         self.pa_schema = pa.Schema.from_pandas(self.empty_data)
 
     def _update_column_names(self, new_cols):
-        """Update the column names by modifying empty_data and pa_schema"""
+        """Update column names in empty_data and pa_schema."""
         self.empty_data.columns = new_cols
         self.pa_schema = pa.Schema.from_pandas(self.empty_data)
 
@@ -288,7 +288,7 @@ class LogicalGetIcebergRead(LogicalOperator):
         self.pa_schema = arrow_schema
 
     def _update_column_names(self, new_cols):
-        """Update the column names by modifying empty_data and pa_schema"""
+        """Update column names in empty_data and pa_schema."""
         raise NotImplementedError(
             "_update_column_names is not implemented for LogicalGetIcebergRead."
         )


bodo/pandas/plan.py
```diff
@@ -282,6 +287,12 @@ def __init__(
         # TODO: avoid this to support operations like renaming columns
         self.pa_schema = arrow_schema
 
+    def _update_column_names(self, new_cols):
+        """Update the column names by modifying empty_data and pa_schema"""
+        raise NotImplementedError(
```
scott-routledge2, line 292: For safety, currently we are adding a trivial projection on top of read_iceberg to enable renaming, so in theory this code should never be executed

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -3376,3 +3374,52 @@ def test_top_level_jit_fallback(datapath):
...
+    def get_bodo_df(df):
+        return df
+
+    bdf = get_bodo_df(df)
+    pdf = df.copy()
+
+    bdf.columns = new_cols
+    pdf.columns = new_cols
+
+    _test_equal(bdf, pdf, check_pandas_types=False)
```
scott-routledge2, line 3429: TODO: Fix JIT section of the test

bodo/pandas/plan.py
```diff
@@ -43,6 +43,11 @@ def __init__(self, plan_class, empty_data, *args):
 
         self.pa_schema = pa.Schema.from_pandas(self.empty_data)
 
+    def _update_column_names(self, new_cols):
+        """Update the column names by modifying empty_data and pa_schema"""
+        self.empty_data.columns = new_cols
```
DrTodd13, line 48: Is there a presumption here that len(new_cols) matches the current number of columns in the plan?  Error check this?

scott-routledge2: Good point, since `empty_data` is a Pandas DataFrame, I was relying on their internal column setting mechanism to catch errors. So if someone does:
```
df = bodo.pandas.DataFrame({"A": [1,2,3], "B": [1,2,3]})
df.columns = ["C"]
```
They will get an error like: 
```
ValueError: Length mismatch: Expected axis has 2 elements, new values have 3 elements
```

bodo/pandas/frame.py
```diff
@@ -358,6 +358,27 @@ def index(self, value):
         self.execute_plan()
         super()._set_axis(1, value)
 
+    @property
+    def columns(self):
+        return super().columns
+
+    @columns.setter
+    def columns(self, value):
```
ehsantn, line 355: This code looks kind of scary in general. Does this need to update `self._head_df` and `self._mgr._md_head` too? See `_update_setitem_internal_state`. Let's add a test for the `head()` case after setting column names.

ehsantn (commented):
See comment.

bodo/pandas/frame.py
scott-routledge2: Done.

ehsantn (approved):
Thanks for the fixes.

PR #785, opened 2025-08-26
Fix LogisticRegression classifier example

scott-routledge2:

The distributed implementation is wrong as it was using a decision_function (which gives distances between -inf to +inf) to supply input to log_loss (which expects probabilities between 0 and 1)

What changed in example CI was that we are pip installing some additional package (imblearn), which was previously downgrading sklearn to 1.6.1 up until about 2 weeks ago, when the CI started failing. In sklearn 1.7, they added validation to the inputs of log loss, which is why the tests were silently passing before.

## Testing strategy
locally with `mpiexec -n 2 pytest bodo/tests -m ml`

## Checklist


IsaacWarren (approved):
Thanks Scott!

PR #789, opened 2025-08-28
Fix cfunc caching on the platform

scott-routledge2:

On the platform, when all ranks try to write to NFS at the same time, it can cause Stale file handle/Text busy errors. In JIT, we only do the write on Rank 0 for NFS and on one rank per node in other cases. This PR just makes the corresponding change in `CFunc.compile`

## Checklist


DrTodd13 (approved):
thanks.  lgtm

PR #791, opened 2025-08-29
Fix compiler import on workers

ehsantn:
precommit removed a critical import that causes NYC Taxi example to fail. This fixes it.

### DIFF
--- a/bodo/spawn/utils.py
+++ b/bodo/spawn/utils.py
@@ -188,6 +188,6 @@ def import_compiler_on_workers():
     """
 
     def import_compiler():
-        pass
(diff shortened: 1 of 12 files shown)


bodo/spawn/utils.py
```diff
@@ -188,6 +188,6 @@ def import_compiler_on_workers():
     """
 
     def import_compiler():
-        pass
+        import bodo.decorators  # isort:skip # noqa
```
ehsantn, line 191: This is the actual change.

scott-routledge2 (approved):
Thanks @ehsantn

PR #792, opened 2025-08-29
Update benchmark scripts to include DataFrames

scott-routledge2:
## Checklist

### DIFF
--- a/benchmarks/nyc_taxi/bodo/nyc_taxi_precipitation.py
+++ b/benchmarks/nyc_taxi/bodo/nyc_taxi_precipitation.py
@@ -5,16 +5,17 @@ Similar to:
 https://github.com/toddwschneider/nyc-taxi-data/blob/c65ad8332a44f49770644b11576c0529b40bbc76/citibike_comparison/analysis/analysis_queries.sql#L1
 """
 
+import argparse
 import time
 
 import pandas as pd
 
 import bodo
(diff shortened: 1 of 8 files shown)


benchmarks/nyc_taxi/bodo/nyc_taxi_precipitation.py
```diff
@@ -88,18 +86,31 @@ def get_time_bucket(t):
...
     hvfhv_dataset = "s3://bodo-example-data/nyc-taxi/fhvhv_tripdata/"
-    get_monthly_travels_weather(weather_dataset, hvfhv_dataset)
+
+    if args.use_jit:
+        start = time.time()
+        bodo.jit(cache=True)(get_monthly_travels_weather)(
+            weather_dataset, hvfhv_dataset
+        )
+        end = time.time()
+        print("Total E2E time:", (end - start))
```
DrTodd13: Curious why the else doesn't have an end-to-end time measurement?

scott-routledge2: I originally had that when I thought we also wanted to measure compilation time/time spawning workers. On the else case it doesn't really matter since all the the heavy loading/computation happens inside the function anyways. Although since we are mainly emphasizing time spent inside the function for JIT, I'm wondering if I should remove for clarity?

PR #794, opened 2025-08-30
Remove pyarrow upperbound version constraint

kevinjqliu:

Allows upstream library, such as pyiceberg, to freely use newer versions of pyarrow

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@ requires = [
     "cython>=3.0",
     # Build Libraries
     "numpy>=1.24",
-    "pyarrow>=19.0,<19.1",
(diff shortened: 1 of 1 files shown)


ehsantn:
Thanks for opening this @kevinjqliu. The challenge is that Bodo uses Arrow C++, which makes building pip wheels difficult for different Arrow versions. We currently have to pin a version, unless if we statically bundle Arrow as well. I'll explore this with the team.

The upcoming PyIceberg 0.10 release enables upgrading to Arrow 21, which we will do to unblock ASAP.

Other thoughts @scott-routledge2 @IsaacWarren ?

PR #800, opened 2025-09-02
Todd/cfunc init

DrTodd13:

Improve cfunc init time.

## Testing strategy

run_ci

## User facing changes

None.

## Checklist


bodo/pandas/physical/expression.cpp
```diff
@@ -519,6 +519,11 @@ std::shared_ptr<ExprResult> PhysicalUDFExpression::ProcessBatch(
     // Actually run the UDF.
     std::shared_ptr<table_info> udf_output;
     if (cfunc_ptr) {
+        if (cfunc_ptr == (table_udf_t)1) {
+            PyThreadState* save = PyEval_SaveThread();
```
ehsantn, line 523: Just curious, are these `PyEval_SaveThread` calls fundamentally necessary? Seems good for safety but looking at the docs they don't seem necessary.

bodo/pandas/physical/expression.h
```diff
@@ -810,36 +814,57 @@ class PhysicalUDFExpression : public PhysicalExpression {
...
-                throw std::runtime_error(
-                    "Expected an integer from compile_cfunc");
-            }
-
-            this->cfunc_ptr =
-                reinterpret_cast<table_udf_t>(PyLong_AsLongLong(result));
-
-            Py_DECREF(bodo_module);
-            Py_DECREF(result);
+            PyThreadState *save = PyEval_SaveThread();
```
ehsantn, line 827: Let's add a comment to Python docs as a reference here.
https://docs.python.org/3/c-api/init.html#thread-state-and-the-global-interpreter-lock

ehsantn (approved):
Thanks @DrTodd13. Looks good to me. Just curious, how much does this help with performance right now?

bodo/pandas/physical/expression.cpp
DrTodd13: Maybe there is some incantation whereby you don't need them but debugging this PR getting the GIL state correct was the biggest problem area.

bodo/pandas/physical/expression.h
DrTodd13: done

bodo/pandas/physical/expression.cpp
DrTodd13: In terms of performance, in my local testing I thought I saw about a second.  With some later code, Scott was having trouble seeing a signal.  In general, it will depend on how far the cfunc is down the pipeline and as a result how much time it takes for the first batch to get to the cfunc operator.  If there is expensive I/O at the start of the pipeline then in many cases I think this will be good.

ehsantn: Ok, makes sense.

PR #801, opened 2025-09-02
Todd/tpch 22 perf

DrTodd13:

Materializes duplicated subqueries at the Bodo plan level.

## Testing strategy

run_ci
Local tpch testing.

## User facing changes

None.

## Checklist


DrTodd13:
The TPCH Q21 test was checking for 2 plan executions but materialization causes there to be 4.  This is functioning as the test of this materialization feature.

bodo/pandas/plan.py
```diff
@@ -695,6 +758,30 @@ def _exec_plan(plan):
...
         return ret
 
+    # Find duplicated subqueries in the given plan and materialize them.
+    while True:
+        # Find the next duplicated subquery in the given plan.
+        top_most_subquery = plan.get_subqueries()
+        if top_most_subquery is not None:
+            # If there is one then go back from the plan to the
+            # dataframe or series object.
+            materialized_frame_or_series = wrap_plan(top_most_subquery)
```
DrTodd13, line 768: I tried another approach where the plan points back to the dataframe and this line then follows that back link.  That approach would have the advantage that if there was another query that needed that dataframe that it wouldn't be computed again.  The way this PR is now, it will only temporarily materialize within the context of one plan submitted to execute_plan.  The approach with the back pointer fails though in parallel mode with that error about spawn mode already running seemingly on the line where the back pointer is set.  I don't understand that at all.  If there are thoughts of a solution then we could consider turning this on.

ehsantn:
See https://github.com/bodo-ai/Bodo/pull/803

PR #803, opened 2025-09-02
Avoid extra agg/merge in TPCH Q22

ehsantn:
Removes the extra agg and merge in the SQL to Pandas translation of TPC-H Q22. We can avoid the subquery caching issue intended for Q22 for now.

https://github.com/bodo-ai/Bodo/blob/dcbb26af8eac979d85d410315be05de6ebe5ed95/BodoSQL/bodosql/tests/test_tpch_second_half.py#L615

The "not exists" translation also looks suboptimal but leaving for future PRs if necessary.

### DIFF
--- a/benchmarks/tpch/dataframe_lib.py
+++ b/benchmarks/tpch/dataframe_lib.py
@@ -921,12 +921,10 @@ def tpch_q22(customer, orders, pd=pd):
         customer_filtered, on="C_CUSTKEY", how="inner"
     )
     customer_selected = customer_selected.loc[:, ["CNTRYCODE", "C_ACCTBAL"]]
-    agg1 = customer_selected.groupby(["CNTRYCODE"], as_index=False, sort=False).size()
-    agg1.columns = ["CNTRYCODE", "NUMCUST"]
-    agg2 = customer_selected.groupby(["CNTRYCODE"], as_index=False, sort=False).agg(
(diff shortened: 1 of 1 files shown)


DrTodd13 (approved):
Doesn't get rid of the filter duplication though right? So PR 801 still useful?  If 801 still needed and it solves this problem then is this a good way to exercise that capability rather than trying to avoid it as here?  Users not always going to write great code right?

ehsantn:
> Doesn't get rid of the filter duplication though right? So PR 801 still useful? If 801 still needed and it solves this problem then is this a good way to exercise that capability rather than trying to avoid it as here? Users not always going to write great code right? I guess in this case there's an optimization that we aren't doing, namely joining groupby's together, but if we did then this would be equivalent I think.

My concern is that subquery caching is a very complicated subject in databases and we can't just implement something simple quickly. We need to review the literature and spend quite some time testing, and also be ready to maintain a complex feature. Here is the design doc for the BodoSQL subquery matching for example: https://bodo.atlassian.net/wiki/spaces/B/pages/1694400514/SQL+Sub-Plan+Caching

This particular case is the result of someone translating SQL to Pandas "wrong", and doesn't sound to me like regular Pandas users would write code like this. Caching parquet read (plus filter) could be a narrower scoped feature to consider if necessary in this case.

Let's discuss and decide tomorrow.

PR #804, opened 2025-09-03
Lazy Scalar and Series Reductions

IsaacWarren:
## Checklist

### DIFF
--- a/bodo/pandas/scalar.py
+++ b/bodo/pandas/scalar.py
@@ -1,0 +1,122 @@
+import typing as pt
+from collections.abc import Callable
+
+from bodo.pandas.lazy_metadata import LazyMetadata
+from bodo.pandas.lazy_wrapper import BodoLazyWrapper
+from bodo.pandas.series import BodoSeries
+
+
+class BodoScalar(BodoLazyWrapper):
+    wrapped_series: BodoSeries
+
+    def __init__(self, wrapped_series: BodoSeries):
+        self.wrapped_series = wrapped_series
+
+    def _get_result_id(self) -> str | None:
+        return self.wrapped_series._get_result_id()
+
+    @classmethod
+    def from_lazy_metadata(
+        cls,
+        lazy_metadata: LazyMetadata,
+        collect_func: Callable[[str], pt.Any] | None = None,
+        del_func: Callable[[str], None] | None = None,
+    ) -> BodoLazyWrapper:
+        return cls(BodoSeries.from_lazy_metadata(lazy_metadata, collect_func, del_func))
+
+    def update_from_lazy_metadata(self, lazy_metadata: LazyMetadata):
+        self.wrapped_series.update_from_lazy_metadata(lazy_metadata)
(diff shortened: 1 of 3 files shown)


IsaacWarren:
> Should single-element getitem on series also become a BodoScalar? Can you easily make that fit this infrastructure which is pretty reduction focused?

I think it should be and should be straightforward once we have the physical implementation of it, I don't want to do it as part of this PR since it's a new feature because we don't have any way to filter on row number/access row number as far as I know, the semantics are a little fuzzy too without an explicit sort

bodo/pandas/scalar.py
```diff
@@ -0,0 +1,146 @@
...
+        from bodo.pandas.utils import BodoLibFallbackWarning
+
+        prev_lazy_plan = self.is_lazy_plan()
+        self.wrapped_series.execute_plan()
+        if prev_lazy_plan:
+            # If we were lazy before we need to confirm
+            # that we have exactly one unique value
+            with warnings.catch_warnings():
+                warnings.simplefilter("ignore", category=BodoLibFallbackWarning)
+            assert self.wrapped_series.nunique() in {0, 1}
```
scott-routledge2, line 52:
```suggestion
            assert self.wrapped_series.nunique() == 1
```

IsaacWarren: I included zero because a series of all pd.NA returns 0 from nunique

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -2463,7 +2458,7 @@ def test_series_describe_empty():
     bds = bd.Series([1, 2, 3])
     bds = bds[bds > 4]
 
-    with assert_executed_plan_count(1):
+    with assert_executed_plan_count(4):
```
scott-routledge2, line 2461: Why is the executed_plan_count 4?

bodo/pandas/series.py
```diff
@@ -2188,19 +2214,7 @@ def _compute_series_reduce(bodo_series: BodoSeries, func_names: list[str]):
...
-    # TODO: use parallel reduction for slight improvement in very large scales
-    for i in range(len(df.columns)):
-        func_name = func_names[i]
-        reduced_val = getattr(
-            df[str(i)], "sum" if func_name == "count" else func_name
-        )()
-        res.append(reduced_val)
-    assert len(res) == len(func_names)
-    return res
+    return wrap_plan(plan=plan)
```
scott-routledge2, line 2217: Update docstring for _compute_series_reduce since it returns a BodoDataFrame now

bodo/tests/test_df_lib/test_end_to_end.py
IsaacWarren: Because we always return a bododataframe from _compute_series_reduce even if it's an empty plan, getting each column is a projection in a separate plan but it only runs the aggregate once since I manually executed the plan in describe

PR #807, opened 2025-09-04
Update TPCH benchmark PD implementation

scott-routledge2:

Updates first 8 queries with Polars benchmark implementation with some minor modification.
The code is generally a bit simpler, less explicitly .loc'ing and dropping columns, also gives correct columns names.

Q4 fallback because we don't support drop_duplicates on multiple columns.
The remaining queries can be pull from Dask's queries with a few modification since they are in a similar style to the polars one.

## Checklist


IsaacWarren (approved):

benchmarks/tpch/dataframe_lib.py
```diff
@@ -257,243 +166,202 @@ def tpch_q02(part, partsupp, supplier, nation, region, pd=pd):
...
+    sel = agg.loc[:, ["O_ORDERKEY", "REVENUE", "O_ORDERDATE", "O_SHIPPRIORITY"]]
+    sel = sel.rename(columns={"O_ORDERKEY": "L_ORDERKEY"})
 
+    sorted = sel.sort_values(by=["REVENUE", "O_ORDERDATE"], ascending=[False, True])
+    result_df = sorted.head(10)
+
+    return result_df
+
+
+# TODO: Fix fallback in drop_duplicates
```
DrTodd13: Did you observe drop_duplicates falling back?  I don't think it is supposed to fallback now.

DrTodd13 (approved):
thanks

benchmarks/tpch/dataframe_lib.py
scott-routledge2: It falls back because the `subset` argument is not supported. I opened an issue.

PR #810, opened 2025-09-04
Avoid confusing warning in rename(copy=False)

ehsantn:
This warning was confusing to me when running NYC Taxi demo and I think does more harm than good. `copy` keyword is changing semantics and actually going away in Pandas, and it's semantics are not very clear to begin with. I think users just try to avoid copy if possible for performance reasons, not correctness of single copy of mutable data.


scott-routledge2 (approved):
Thanks. Maybe we can remove the "copy" argument from supported args in-case older Pandas code is expecting the column names to be updated in place?

DrTodd13:
> Thanks. Maybe we can remove the "copy" argument from supported args in-case older Pandas code is expecting the column names to be updated in place?

My recollection is that there are instances of copy=True and copy=False used in our tests and benchmarks.  Wouldn't removing copy as a supported arg break those tests?

PR #813, opened 2025-09-05
Update remaining TPCH queries

scott-routledge2:

* Revert Q04 since I think the previous one is a bit simpler
* For the most part tried to update queries to Dask with a few changes:
1. groupby.func.reset_index.rename -> groupby(as_index=False).agg
2. Broke up operators into steps and change variable names to match polars PD queries
3. upper cased column names

Also fixes a few bugs:
* Testing bug where we were not actually comparing correctness to Pandas, this actually exposed a few bugs (Q17 and Q20), but when I rewrote those queries failures went away. Opened issues.
* Old version of query 20 was incorrect
* Q11 depends on Scale factor
* In some places column names were renamed/output was cast to DataFrame or only head was collected to match what polars/dask was doing. Results are also rounded to two decimal points.

## Testing strategy
Compared results to Pandas (of both versions)

## Checklist

### DIFF
--- a/benchmarks/tpch/dataframe_lib.py
+++ b/benchmarks/tpch/dataframe_lib.py
@@ -501,7 +501,6 @@ def tpch_q13(customer, orders, pd=bodo.pandas):
     return result_df
 
 
-# TODO: support Series.where
 def tpch_q14(lineitem, part, pd=bodo.pandas):
     """Adapted from:
     https://github.com/coiled/benchmarks/blob/13ebb9c72b1941c90b602e3aaea82ac18fafcddc/tests/tpch/dask_queries.py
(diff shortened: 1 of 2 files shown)


bodo/pytest.ini
```diff
@@ -130,3 +130,8 @@ markers =
     # DataFrame Library Azure NP=2
     bodo_df_1of2: part 1/2 of the test suite
     bodo_df_2of2: part 2/2 of the test suite
+
+    # DataFrame Library Azure NP=2
```
scott-routledge2, line 134: Was getting warnings that this marker was not defined

DrTodd13 (approved):

IsaacWarren (approved):
Thanks Scott

PR #814, opened 2025-09-07
CTE implementation.

DrTodd13:

Generate C++ plan tree with CTE nodes.  Implement backend support for CTE nodes.  Add ability to run UDF at query optimization time.

## Testing strategy

run_ci

## User facing changes

None

## Checklist


bodo/libs/_bodo_common.cpp
```diff
@@ -1588,6 +1588,44 @@ std::unique_ptr<array_info> alloc_array_like(
     }
 }
 
+std::unique_ptr<array_info> alloc_array_like(
```
IsaacWarren: We have alloc_empty_array in _table_builder_utils.cpp that I think does the same thing as this with support for maps and array item arrays

bodo/libs/_table_builder_utils.cpp
```diff
@@ -92,6 +92,18 @@ std::shared_ptr<table_info> alloc_table_like(
                                         table->column_names, table->metadata);
 }
 
+std::shared_ptr<table_info> alloc_table_like(
```
IsaacWarren: How is this different than alloc_table above? Don't we need the dictionary builders?

bodo/pandas/_executor.h
```diff
@@ -7,6 +7,10 @@
 #include "_physical_conv.h"
 #include "_pipeline.h"
 #include "duckdb/planner/logical_operator.hpp"
+#include "physical/cte.h"
+
+// Holds table_index to PhysicalCTE mapping during physical plan construction.
+extern std::map<duckdb::idx_t, std::shared_ptr<PhysicalCTE>> g_ctes;
```
IsaacWarren: Can we pass this to the executor from the planner instead of making it a global?

bodo/pandas/_physical_conv.cpp
```diff
@@ -253,6 +256,45 @@ void PhysicalPlanBuilder::Visit(duckdb::LogicalComparisonJoin& op) {
...
+    // The active pipeline finishes after the duplicate side.
+    this->active_pipeline = nullptr;
+    /*
+     * The build side of joins want to insert at the front of the finished
+     * pipelines but that causes a problem with CTEs because the use of a
+     * CTE can then preceed its calculation.  When we are done with a CTE,
+     * we now lock the current set of finished pipelines so the ordering
+     * cannot be changed.
+     */
+    this->lock_finished();
```
IsaacWarren, line 277: I don't think I understand what's going on here

bodo/pandas/physical/aggregate.h
```diff
@@ -352,6 +359,7 @@ class PhysicalAggregate : public PhysicalSource, public PhysicalSink {
     PhysicalAggregateMetrics metrics;
     // Mapping of input table column indices to move keys to the front.
     std::vector<int64_t> input_col_inds;
+    int num_consume = 0;
```
IsaacWarren: Where is this used?

bodo/pandas/plan.py
```diff
@@ -155,6 +222,13 @@ def __init__(self, empty_data, *args):
         super().__init__(self.__class__.__name__, empty_data, *args)
 
 
+class LogicalOperatorLeaf(LogicalOperator):
+    """Base class for all logical operators in the Bodo query plan."""
```
ehsantn:
```suggestion
    """Base class for all logical operators in the Bodo query plan that are leaf of the plan tree (e.g. Parquet read)."""
```

bodo/pandas/vendor/duckdb/src/common/enum_util.cpp
IsaacWarren, line 1: Assuming all the changes to vendored files are from upstream

bodo/pandas/_plan.cpp
```diff
@@ -580,6 +584,440 @@ std::vector<int> get_projection_pushed_down_columns(
...
+    switch (arrow_type) {
+        case arrow::Type::BOOL: {
+            // Cast to the concrete Arrow boolean array type
+            auto bool_arr = std::static_pointer_cast<arrow::BooleanArray>(arr);
+            // Get a pointer to the DuckDB vector's boolean storage
+            auto data = duckdb::FlatVector::GetData<bool>(vec);
+            auto &validity = duckdb::FlatVector::Validity(vec);
+
+            for (duckdb::idx_t i = 0; i < count; i++) {
+                if (bool_arr->IsNull(i)) {
```
IsaacWarren: Instead of only setting each value if it's valid can we just do a memcpy and then set the validity values in a loop, that should be faster and it shouldn't matter what the value is at invalid locations

bodo/pandas/_plan.cpp
```diff
@@ -612,7 +1088,8 @@ duckdb::unique_ptr<duckdb::Expression> make_scalar_func_expr(
     duckdb::ScalarFunction scalar_function = duckdb::ScalarFunction(
         scalar_name, source->types, out_type, RunFunction, nullptr, nullptr,
         nullptr, nullptr, duckdb::LogicalTypeId::INVALID,
-        duckdb::FunctionStability::VOLATILE,
+        duckdb::FunctionStability::CONSISTENT,
+        // duckdb::FunctionStability::VOLATILE,
```
IsaacWarren:
```suggestion
```

bodo/libs/_bodo_common.cpp
```diff
@@ -1588,6 +1588,44 @@ std::unique_ptr<array_info> alloc_array_like(
...
+                "StructType");
+        }
+        std::vector<std::shared_ptr<array_info>> child_arrays;
+        child_arrays.reserve(dtype_as_struct->child_types.size());
+        for (auto& child_type : dtype_as_struct->child_types) {
+            child_arrays.push_back(alloc_array_like(child_type));
+        }
+        return alloc_struct(0, std::move(child_arrays));
+        // throw std::runtime_error("alloc_array_like from bodo::DataType for
+        // STRUCT unimplemented");
```
IsaacWarren:
```suggestion
```

IsaacWarren (approved):
Thanks Todd this looks great!
Is it expected to not change the number of run plans in some of our tests like tpch?

bodo/libs/_table_builder_utils.cpp
DrTodd13: The new one takes a Schema, the old one takes an existing table_info.

bodo/pandas/plan.py
```diff
@@ -80,63 +81,129 @@ def __str__(self):
 
     __repr__ = __str__
 
-    def generate_duckdb(self, cache=None):
+    def bfs_duplicate(self):
```
ehsantn, line 104: Let's add docstring with some explanation.

bodo/pandas/plan.py
```diff
@@ -80,63 +81,129 @@ def __str__(self):
...
+        visited = set()
+        queue = deque([self])
+
+        while queue:
+            node = queue.popleft()
+            if id(node) in visited and not isinstance(node, LogicalOperatorLeaf):
```
ehsantn, line 121: No CTE for things like reading Parquet twice? Any reason?

bodo/libs/_bodo_common.cpp
DrTodd13: Done.

bodo/pandas/plan.py
```diff
@@ -80,63 +81,129 @@ def __str__(self):
...
+                node,
+                (
+                    LogicalComparisonJoin,
+                    LogicalSetOperation,
+                    LogicalInsertScalarSubquery,
+                    LogicalCrossProduct,
+                ),
+            )
+
+        if cte_ref is not None and self is cte_ref[0]:
```
ehsantn, line 200: These different code paths need comments with explanation and examples.

bodo/pandas/plan.py
```diff
@@ -80,63 +81,129 @@ def __str__(self):
...
+            if id(self) in cache:
+                return cache[id(self)]
+            cte_ref_plan = LogicalCTERef(self.empty_data, cte_ref[1])
+            ret = getattr(plan_optimizer, cte_ref_plan.plan_class)(
+                cte_ref_plan.pa_schema, cte_ref[1]
+            )
+            cache[id(self)] = ret
+            return ret
+        elif cte_node is not None:
+            if id(self) in cache:
```
ehsantn, line 221: Should `id(self)` be `id(cte_node[0])` here?

bodo/pandas/plan.py
```diff
@@ -80,63 +81,129 @@ def __str__(self):
...
+                raise Exception("Should never find cache re-use for cte_node.")
+            duplicate = cte_node[0].generate_duckdb(cache=cache)
+            uses_duplicate = self.generate_duckdb(cache=cache, cte_ref=cte_node)
+            cte_plan = LogicalMaterializedCTE(
+                self.empty_data, cte_node[0], self, cte_node[1]
+            )
+            use_cache = should_use_cache(self)
```
ehsantn: Unused.

bodo/pandas/_physical_conv.cpp
DrTodd13: The build side of join runs before the probe side.  So, our previous code just adds the build pipeline to be the very first pipeline.  If the build side uses a CTE then that pipeline would try to run before the pipeline that generates the CTE result runs.  So, for build side, you could keep our single list of pipelines and try to figure out where in the list the pipeline for the build side should go.  In this PR, I just tried to make that determination really easy by saying that once you've created a CTE pipeline that you can't insert before it and so they are "locked."  Conceptually, in my PR, there is a single pipeline list but you get that by combining the locked pipelines with the non-locked ones.  The CTE pipelines become locked and thus the build side will add first thing after the locked CTE pipelines.

bodo/pandas/physical/aggregate.h
DrTodd13: Left over debug stuff.  Removed.  Thanks.

bodo/pandas/plan.py
```diff
@@ -80,63 +81,129 @@ def __str__(self):
...
+
+            ret = getattr(plan_optimizer, cte_plan.plan_class)(
```
ehsantn, line 241: `LogicalMaterializedCTE` becomes top level if there is any CTE used in the query plan tree?

bodo/pandas/vendor/duckdb/src/common/enum_util.cpp
DrTodd13: All except the spot where I turn off CTE inlining since we can't deep copy now.

bodo/pandas/plan.py
```diff
@@ -80,63 +81,129 @@ def __str__(self):
...
-        args = [recursive_check(x, use_cache) for x in reversed(self.args)]
-        args.reverse()
-
-        # Create real duckdb class.
-        ret = getattr(plan_optimizer, self.plan_class)(self.pa_schema, *args)
-        # Add to cache so we don't convert it again.
-        cache[id(self)] = ret
-        return ret
+        if cte_ref is None:
+            cte_node = self.bfs_duplicate()
```
ehsantn, line 167: If there are no duplicates, this replicates the BFS on all subtrees redundantly? Also, what if there are multiple duplicates in unrelated subtrees?

bodo/pandas/plan.py
```diff
@@ -80,63 +81,129 @@ def __str__(self):
 
     __repr__ = __str__
 
-    def generate_duckdb(self, cache=None):
+    def bfs_duplicate(self):
+        visited = set()
+        queue = deque([self])
+
+        while queue:
```
ehsantn, line 117: I remember you mentioned something about the order of processing being important. Could you add it as comment?

bodo/pandas/_plan.cpp
```diff
@@ -588,7 +1026,45 @@ std::vector<int> get_projection_pushed_down_columns(
  */
 static void RunFunction(duckdb::DataChunk &args, duckdb::ExpressionState &state,
```
ehsantn, line 1004: Update docstring.

bodo/pandas/_plan.cpp
```diff
@@ -580,6 +584,440 @@ std::vector<int> get_projection_pushed_down_columns(
...
 }
 
+std::shared_ptr<arrow::Array> duckdbVectorToArrowArray(duckdb::Vector &vec,
+                                                       duckdb::idx_t count) {
+    using namespace duckdb;
+
+    // Flatten to ensure contiguous data
+    vec.Flatten(count);
+
+    auto &type = vec.GetType();
```
ehsantn: Too many non-trivial `auto` uses in this function to me.

bodo/pandas/_plan.cpp
```diff
@@ -580,6 +584,440 @@ std::vector<int> get_projection_pushed_down_columns(
...
+                if (!status.ok()) {
+                    throw std::runtime_error("builder.Append failed.");
+                }
+            }
+            std::shared_ptr<arrow::Array> arr;
+            status = builder.Finish(&arr);
+            if (!status.ok()) {
+                throw std::runtime_error("builder.Finish failed.");
+            }
+            return arr;
```
ehsantn: Looks to me like this repeated code could be in a templated function that takes `builder` and `data` as input and returns the output array.

bodo/pandas/_plan.cpp
```diff
@@ -580,6 +584,440 @@ std::vector<int> get_projection_pushed_down_columns(
...
+        }
+        default:
+            throw std::runtime_error(
+                "Unsupported DuckDB type for Arrow conversion " +
```
ehsantn, line 828: I assume unsupported types will just potentially affect some optimization in run function code path and won't lead to user visible errors?

bodo/pandas/_executor.h
DrTodd13: Did something better.

bodo/pandas/_plan.cpp
```diff
@@ -580,6 +584,440 @@ std::vector<int> get_projection_pushed_down_columns(
...
+
+    switch (arrow_type) {
+        case arrow::Type::BOOL: {
+            // Cast to the concrete Arrow boolean array type
+            auto bool_arr = std::static_pointer_cast<arrow::BooleanArray>(arr);
+            // Get a pointer to the DuckDB vector's boolean storage
+            auto data = duckdb::FlatVector::GetData<bool>(vec);
+            auto &validity = duckdb::FlatVector::Validity(vec);
+
+            for (duckdb::idx_t i = 0; i < count; i++) {
```
ehsantn: Same as above about refactoring to templated function.

bodo/pandas/_pipeline.cpp
```diff
@@ -106,9 +130,36 @@ uint64_t Pipeline::Execute() {
...
+        std::cout << "Rank " << rank << " Pipeline::Execute after ProduceBatch "
+                  << source->ToString() << " "
+                  << static_cast<int>(produce_result) << std::endl;
+#endif
         // Run the between_ops and sink of the pipeline allowing repetition
         // in the HAVE_MORE_OUTPUT case.
-        finished = midPipelineExecute(0, batch, produce_result);
+        finished = midPipelineExecute(0, batch, produce_result, rank);
+
+        // If the next operator in the pipeline isn't finished even though we
```
ehsantn, line 224: I think the normal approach is that the source should handle this and return an empty table. It would simplify this code quite a bit I think. See Parquet example:

https://github.com/bodo-ai/Bodo/blob/88ffc3768217f54b335cbc53455de1dcf815ca45/bodo/io/arrow_reader.h#L256

bodo/pandas/_pipeline.cpp
```diff
@@ -11,34 +11,57 @@
...
     if (idx >= between_ops.size()) {
+        // Iterating here as in the normal section below so that if the sink
+        // says HAVE_MORE_OUTPUT that we can iterate with an empty batch.
+        while (true) {
 #ifdef DEBUG_PIPELINE
-        for (unsigned i = 0; i < idx; ++i)
-            std::cout << " ";
-        std::cout << "midPipelineExecute before ConsumeBatch "
-                  << sink->ToString() << std::endl;
+            for (unsigned i = 0; i < idx; ++i)
```
ehsantn: These debug lines are making the code hard to read. Are they in day to day use and necessary to have checked in?

bodo/pandas/_executor.h
ehsantn: Same comment

bodo/pandas/_executor.h
```diff
@@ -41,15 +56,27 @@ class Executor {
         // Pipelines generation ensures that pipelines are in the right
         // order and that the dependencies are satisfied (e.g. join build
         // pipeline is before probe).
-        QueryProfileCollector::Default().Init();
+        int rank;
+        MPI_Comm_rank(MPI_COMM_WORLD, &rank);
 #ifdef DEBUG_PIPELINE
-        std::cout << "ExecutePipelines with " << pipelines.size()
```
ehsantn, line 46: Same about debug code.

bodo/libs/_bodo_common.cpp
ehsantn: Done?

bodo/pandas/physical/aggregate.h
```diff
@@ -267,10 +267,17 @@ class PhysicalAggregate : public PhysicalSource, public PhysicalSink {
         override {
         time_pt start_produce = start_timer();
         bool out_is_last = false;
-        std::shared_ptr<table_info> next_batch =
-            groupby_produce_output_batch_wrapper(this->groupby_state.get(),
-                                                 &out_is_last, true);
+        std::shared_ptr<table_info> next_batch;
+        if (!this->groupby_state->output_state ||
```
ehsantn: Why is this needed? This code path should have this functionality:
https://github.com/bodo-ai/Bodo/blob/98f01e3384fa312cf82adde308af8a41afa6c185/bodo/libs/_chunked_table_builder.cpp#L1764

bodo/pandas/physical/cte.h
```diff
@@ -0,0 +1,115 @@
...
+
+    int64_t getOpId() const { return PhysicalSink::getOpId(); }
+
+   private:
+    std::unique_ptr<ChunkedTableBuilderState> collected_rows;
+    const std::shared_ptr<bodo::Schema> output_schema;
+    friend class PhysicalCTERef;
+
+   protected:
+    duckdb::LogicalMaterializedCTE& node;
```
ehsantn: Not used?

bodo/pandas/_plan.cpp
DrTodd13: Would like to save this for a follow-up if needed.  This code isn't performance critical and only happens for small sample values used to run UDFs during optimization.

bodo/libs/_table_builder_utils.cpp
IsaacWarren: On line 66 it takes a schema?

DrTodd13: Fixed.

bodo/pandas/plan.py
DrTodd13: Done.

bodo/pandas/plan.py
DrTodd13: I guess if you are filtering it might be useful but if a leaf node and no filters then seems to just increase memory pressure?

ehsantn: It could potentially avoid slow I/O. Chunked buffer spills to disk if necessary. But the way it's used in this PR may not work with spilling. Not a big deal for this PR right now.

bodo/pandas/plan.py
DrTodd13: Done.

bodo/pandas/plan.py
DrTodd13: Don't think that is possible.  For this check to fire, it would have to be the case that we failed to find the current node as a duplicated node for the higher subtree that it is in.

bodo/pandas/plan.py
DrTodd13: Done.

bodo/pandas/plan.py
DrTodd13: Yes.

bodo/pandas/_executor.h
```diff
@@ -8,6 +8,7 @@
 #include "_pipeline.h"
 #include "duckdb/planner/logical_operator.hpp"
 
+// Holds table_index to PhysicalCTE mapping during physical plan construction.
```
scott-routledge2: add to docstring below?

bodo/pandas/plan.py
DrTodd13: That relates I think to Isaac's question above.  You need to do this top-down ergo bfs rather than dfs.  Can clarify main bfs_duplicated comment.

bodo/pandas/_executor.h
scott-routledge2: tracing seems useful to me. Maybe we can wrap it in a macro for better readability? Basically just define as a print with some args if DEBUG_PIPELINE is defined and a noop otherwise.

bodo/pandas/_plan.cpp
DrTodd13: Done.

bodo/pandas/_plan.cpp
DrTodd13: Fixed.

bodo/pandas/physical/cte.h
DrTodd13: Done.

bodo/pandas/_pipeline.cpp
DrTodd13: I'm finding these pretty useful.  Perhaps can remove at some point.  In the mean time, can you collapse in the IDE to make it easier to read it?

bodo/pandas/physical/cte.h
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    /**
+     * @brief Get the output schema of join probe
+     *
+     * @return std::shared_ptr<bodo::Schema>
+     */
+    const std::shared_ptr<bodo::Schema> getOutputSchema() override {
+        // return bodo::Schema::FromArrowSchema(
+        //     cte->output_schema->ToArrowSchema());
```
scott-routledge2:
```suggestion
```

bodo/pandas/physical/join.h
```diff
@@ -635,36 +639,22 @@ class PhysicalJoin : public PhysicalProcessBatch, public PhysicalSink {
...
      * output according to bindings
      */
-    static void initOutputColumnMapping(std::vector<uint64_t>& col_inds,
-                                        const std::vector<uint64_t>& keys,
-                                        const std::set<uint64_t>& non_equi_keys,
-                                        uint64_t ncols,
-                                        std::set<int64_t>& bound_inds) {
+    static void initOutputColumnMapping(
+        std::vector<uint64_t>& col_inds, const std::vector<uint64_t>& keys,
+        const std::set<uint64_t>& non_equi_keys, uint64_t ncols,
```
scott-routledge2:
```suggestion
        uint64_t ncols,
```
non_equi_keys isn't used anymore?

bodo/pandas/plan.py
```diff
@@ -80,63 +81,129 @@ def __str__(self):
...
+                cte_plan.pa_schema, duplicate, uses_duplicate, cte_node[1]
+            )
+            cache[id(self)] = ret
+            return ret
+        else:
+            # Sometimes the same LazyPlan object is encountered twice during the same
+            # query so  we use the cache dict to only convert it once.
+            # If previously converted then use the last result.
+            # Don't cache expression nodes.
+            # TODO - Try to eliminate caching altogether since it seems to cause
```
scott-routledge2, line 251: This TODO seems to contradict the note below?

bodo/pandas/plan.py
DrTodd13: Fixed the redundant BFS searches.  If there are additional duplicates on the non-duplicate side of the top-level CTE then right now it won't find them.  Will do this in a follow-up if we find it is happening.

bodo/pandas/_plan.cpp
DrTodd13: I think this will cause fallback to Pandas right now.  If you treated the UDF as volatile instead then you could probably run the plan.  Again, the CTE optimization path is useless at this point since we can't inline so if that is the only part that calls this then we could turn off that pass.  Let me know.

bodo/pandas/_pipeline.cpp
DrTodd13: I actually think that is less consistent with the other way we do things.  For mid-pipeline, if our child says have_more_output then we handle passing the empty table outside of previous or next pipeline node.

bodo/pandas/_executor.h
DrTodd13: Imrpoved.

bodo/pandas/physical/cte.h
DrTodd13: Done.

bodo/pandas/physical/join.h
DrTodd13: Done.

bodo/pandas/plan.py
DrTodd13: Good catch but I think this was the state of things previously and it just wasn't documented why we were still using caching.  At least now it is clearer what you have to accomplish if you want to eliminate caching.

bodo/pandas/_executor.h
DrTodd13: Improved a bit I think.

bodo/pandas/_pipeline.cpp
DrTodd13: Improved.

bodo/pandas/physical/aggregate.h
DrTodd13: Will that line work if you've never added any rows to the table?

bodo/pandas/_plan.cpp
DrTodd13: Done.

bodo/pandas/_plan.cpp
DrTodd13: Done.

bodo/pandas/_plan.cpp
ehsantn: Looks to me like we can add a few more data types to make this code path a lot more complete with small effort. `TIMESTAMP_NS`, `DATE`, `DECIMAL` seem to be the main ones. Maybe `TIMESTAMP_TZ`, `SQLNULL` , `TIME`, `STRING_LITERAL`, `INTEGER_LITERAL`, `STRUCT`, `LIST`, `ARRAY`, `MAP`, `BLOB` for later.

bodo/pandas/_plan.cpp
```diff
@@ -580,15 +584,249 @@ std::vector<int> get_projection_pushed_down_columns(
...
+                } else {
+                    auto view = str_arr->GetView(i);
+                    data[i] =
+                        StringVector::AddString(vec, view.data(), view.size());
+                }
+            }
+            break;
+        }
+        default:
+            std::cout << "arrow type " << static_cast<int>(arrow_type)
```
ehsantn, line 992: Same as above here.

bodo/pandas/_pipeline.cpp
```diff
@@ -11,50 +129,46 @@
...
-                   OperatorResult::FINISHED;
-#ifdef DEBUG_PIPELINE
-        for (unsigned i = 0; i < idx; ++i)
-            std::cout << " ";
-        std::cout << "midPipelineExecute after ConsumeBatch "
-                  << sink->ToString() << std::endl;
-#endif
-        return ret;
+        // Iterating here as in the normal section below so that if the sink
+        // says HAVE_MORE_OUTPUT that we can iterate with an empty batch.
```
ehsantn, line 137: What kind of sink returns `HAVE_MORE_OUTPUT`?

bodo/pandas/_pipeline.cpp
ehsantn: This just looks scary a bit but seems ok.

bodo/pandas/physical/aggregate.h
ehsantn: I think so since the output buffer is created in the groupby state constructor using schema:

https://github.com/bodo-ai/Bodo/blob/48066ec372ea70df08fb5da2fb2f949ea89d33c0/bodo/libs/streaming/_groupby.cpp#L1887
https://github.com/bodo-ai/Bodo/blob/48066ec372ea70df08fb5da2fb2f949ea89d33c0/bodo/libs/_chunked_table_builder.cpp#L621

bodo/pandas/_pipeline.cpp
DrTodd13: I forget but I'm pretty sure I saw a sink doing this, else I don't think I would have added it.

PR #815, opened 2025-09-08
BSE-5104: Support BodoScalars in filter expressions

IsaacWarren:

BodoScalar support in BodoSeries binops
ScalarSubqueryExpression support in BodoDataFrame filter
Expose LogicalCrossJoin to LazyPlans
Adds support for NestedLoopJoin with conditition functions, the optimizer turned a cross product into a join with only a non-equi key for tpch q22 so this was necessary.

## Testing strategy
Unit tests/PR CI

## User facing changes
Filtering on the result of an aggregate can happen without multiple plan executions

## Checklist

### DIFF
--- a/bodo/pandas/scalar.py
+++ b/bodo/pandas/scalar.py
@@ -59,7 +59,7 @@ class BodoScalar(BodoLazyWrapper):
     @property
     def __pandas_priority__(self):
         """
-        Overrride this so we don't call get_value during planning.'
+        Override this so we don't call get_value during planning.'
         This is used by pandas during comparisons and arithmetic operations.'
         """
         return None


bodo/pandas/base.py
```diff
@@ -146,9 +146,12 @@ def _empty_like(val):
     """
     import pyarrow as pa
 
-    if not isinstance(val, (BodoDataFrame, BodoSeries)):
+    if type(val) not in {BodoDataFrame, BodoSeries, BodoScalar}:
```
DrTodd13, line 149: Why doesn't the isinstance check work for BodoScalar?

IsaacWarren: Because we overload \_\_class\_\_ to point to the wrapped value

bodo/pandas/scalar.py
```diff
@@ -52,11 +56,20 @@ def get_value(self):
             assert self.wrapped_series.nunique() in {0, 1}
         return self.wrapped_series[0]
 
+    @property
+    def __pandas_priority__(self):
+        """
+        Overrride this so we don't call get_value during planning.'
```
scott-routledge2:
```suggestion
        Override this so we don't call get_value during planning.'
```

PR #819, opened 2025-09-09
Enable import tidy in bodo/tests and fix errors

scott-routledge2:

Makes Removes JIT import from main testing files, utils, test_iceberg and test_df_lib
(caching_tests and test_streaming still potentially have compiler imports, I didn't think it was worth it to also fix those tests since they are irrelevant to DataFrames/Spawn tests. There were a few changes that I needed to make to get this to work:
1. Move imports inside of functions
2. Convert some parameters to be lazy by making them lambdas and calling them in the fixture/test that needs them
3. Refactor some utilities to `utils_jit.py` file so they can be imported inside of tests
4. Removed some decorators/moved them inside of tests for example pytest.skipif can be moved inside the test with a if + pytest.skip, some parameterizations like lists of unsupported functions were moved inside of tests using a for loop to avoid importing the module they are defined in.

Followups:
Remove additional dependencies inside of
* _test_equal depends on the compiler
* check_func depends on the compiler
* memory leak check depends on the compiler

## Testing strategy

Manually checked at the first test that runs and check the list of modules loaded to confirm none of the bad modules were included.

CI/Nightly

## Checklist


bodo/tests/test_array_agg.py
```diff
@@ -424,14 +432,6 @@ def impl(df):
     )
 
 
-@bodo.jit(distributed=False)
-def bodo_get_distinct_group(raw_data, raw_order):
```
scott-routledge2, line 428: unused as far as I can tell

bodo/tests/test_dataframe_errorchecking.py
```diff
@@ -2,34 +2,24 @@
...
 import pytest
 
 import bodo
 from bodo.tests.utils import pytest_pandas
-from bodo.utils.typing import BodoError
 
 pytestmark = pytest_pandas
 
 
-@pytest.fixture(params=bodo.hiframes.pd_dataframe_ext.dataframe_unsupported)
```
scott-routledge2, line 18: Unfortunately we can't use this pattern of parametrizing on unsupported methods/attrs; either use a for loop inside of test or hardcode list instead (unless maybe we refactor them to a different file)

bodo/tests/test_date.py
```diff
@@ -1510,7 +1511,34 @@ def series_value(request):
     return request.param
 
 
-@pytest.mark.parametrize("date_fields", bodo.hiframes.pd_timestamp_ext.date_fields)
+@pytest.mark.parametrize(
+    "date_fields",
```
scott-routledge2: Date fields/methods unlikely to change so I opted to inline them here

bodo/tests/test_iceberg/test_iceberg_theta.py
```diff
@@ -169,69 +167,6 @@ def test_iceberg_write_disabled_theta(
         bodo.enable_theta_sketches = orig_enable_theta
 
 
-@run_rank0
```
scott-routledge2, line 172: refactored to test_iceberg/utils_jit.py to avoid decorators in test files

buildscripts/bodo/azure/azure-test-conda-linux-macos.yml
```diff
@@ -135,6 +135,12 @@ jobs:
       # For caching tests we need to update the location of the decryption file.
       # We set the absolute path as an environment variable.
       export BODO_TRACING_DECRYPTION_FILE_PATH=`echo "$(System.DefaultWorkingDirectory)/buildscripts/decompress_traces.py"`
+
+      if [[ "${{ parameters.test_type }}" == "SPAWN" || "${{ parameters.test_type }}" == "DF_LIB" ]]; then
```
scott-routledge2: Adding an ignore for sub-directories that are only needed for compiler tests. These can also be converted to lazily import the compiler although not necessary in my opinion

bodo/tests/test_array_agg.py
```diff
@@ -106,9 +105,12 @@
             id="date",
         ),
         pytest.param(
-            (
+            lambda: (
```
scott-routledge2, line 108: Converted certain params to lambdas that are evaluated inside the body of fixtures/functions to avoid importing at collection time. The body of fixtures do not evaluate until tests actually use them, but anything in parametrize gets evaluated immediately.

bodo/tests/test_metadata.py
```diff
@@ -384,73 +388,78 @@ def impl2(S):
 # cause the output to not be dist
 
 
-@bodo.jit(distributed=["out"])
-def int_gen_dist_df():
-    out = pd.DataFrame({"A": pd.Series(np.arange(1))})
-    return out
+@pytest.fixture
```
scott-routledge2, line 394: The indirect argument let's you pass params to a fixture of the same name first before they get used in the test.

bodo/tests/test_df_lib/test_plan.py
```diff
@@ -6,11 +6,11 @@
 
 import pyarrow as pa
 
-from bodo.ext import plan_optimizer
-
```
IsaacWarren: Does the plan optimizer use the compiler?

scott-routledge2: plan_optimizer does not import the compiler, so should be safe to include this at the top level actually.

bodo/utils/utils.py
```diff
@@ -38,6 +38,8 @@
 from numba.np.numpy_support import as_dtype
 
 import bodo
+
+import bodo.decorators  # isort:skip # noqa
```
scott-routledge2: Adding decorators here because the import will not work on it's own otherwise, and we use it in a lot of places for tests e.g. for `run_rank0`

scott-routledge2: nevermind, think this might be a circular import

.github/workflows/_test_python_source.yml
```diff
@@ -120,6 +120,12 @@ jobs:
           env:
             BODO_TESTING_ONLY_RUN_1D_VAR: true
             PYTEST_MARKER: ${{ inputs.pytest-marker }}
+            # Ignore streaming, caching and other tests that would
+            # normally import the compiler in spawn/df_lib modes.
+            PYTEST_IGNORE: >-
+              ${{ inputs.test-type == 'NORMAL' && ' ' || 
+                  '--ignore bodo/tests/test_streaming
```
ehsantn, line 127: Looks like we are basically ignoring `test_streaming` in CI now? Ok to me just asking.

scott-routledge2: We are ignoring them in DataFrames/Spawn CI now. So the tests will still run on CI in JIT mode

PR #827, opened 2025-09-12
Enable iceberg tests

IsaacWarren:
Re-enable iceberg tests and upgrade spark pyiceberg, going to wait until pyiceberg conda-forge is done

## Testing strategy
PR CI

## User facing changes
N/A

## Checklist


bodo/io/iceberg/file_io.py
```diff
@@ -23,38 +23,25 @@ def _map_wasb_to_abfs(scheme: str, netloc: str) -> tuple[str, str]:
...
 
     @staticmethod
-    def parse_location(location: str) -> tuple[str, str, str]:
-        """Return the path without the scheme."""
+    def parse_location(
+        location: str, *args, properties: Properties = EMPTY_DICT, **kwargs
+    ) -> Tuple[str, str, str]:
+        """Return (scheme, netloc, path) for the given location.
 
-        if _is_windows_path(location):
```
ehsantn, line 52: Windows is now working differently? Have you tested it?

bodo/io/parquet_pio.py
```diff
@@ -1329,16 +1337,14 @@ def parquet_file_schema(
 
     col_names = []
     col_types = []
-
     # during compilation we only need the schema and it has to be the same for
     # all processes, so we can set parallel=True to just have rank 0 read
     # the dataset information and broadcast to others
-    pq_dataset = get_parquet_dataset(
+    pq_dataset = get_dataset_unify_nulls(
```
ehsantn: Why this change? `get_parquet_dataset` runs only on rank 0 but `get_dataset_unify_nulls` doesn't for example.

I don't like the code duplication of these two functions in general also.

bodo/io/parquet_pio.py
```diff
@@ -1446,6 +1452,13 @@ def parquet_file_schema(
             # should still be included.
             selected_columns.append(index_col)
 
+    # Convert dictionary columns to use int32 indices
+    # since our c++ dict string array uses int32 indices
+    for i in range(len(arrow_types)):
+        if isinstance(arrow_types[i], pa.DictionaryType):
+            arrow_types[i] = pa.dictionary(pa.int32(), arrow_types[i].value_type)
```
ehsantn, line 1461: `ordered` flag isn't relevant here?

bodo/io/parquet_reader.cpp
```diff
@@ -137,7 +137,6 @@ PyObject* ParquetReader::get_dataset() {
     }
 
     this->ds_partitioning = PyObject_GetAttrString(ds, "partitioning");
-    this->set_arrow_schema(PyObject_GetAttrString(ds, "schema"));
```
ehsantn: This is removed since `ArrowReader` constructor sets schema anyways?

bodo/tests/test_quickstart_docs.py
```diff
@@ -10,12 +10,15 @@
...
-from bodo.tests.utils import check_func, pytest_spawn_mode, temp_env_override
+from bodo.tests.utils import check_func, temp_env_override
 from bodo.utils.testing import ensure_clean2
 
-pytestmark = pytest_spawn_mode + [pytest.mark.test_docs] + [pytest.mark.df_lib]
+pytestmark = [pytest.mark.test_docs]
 
 
+@pytest.mark.df_lib
+# ensure_clean2 uses a barrier which requires jit_dependency
```
ehsantn, line 20: Avoiding JIT in the barrier should be easy if necessary, FYI.

bodo/io/parquet_reader.cpp
IsaacWarren: Exactly, we use the schema passed in instead of the arrow inferred schema since arrow changed how they infer schemas with this release

bodo/io/iceberg/file_io.py
IsaacWarren: Pyiceberg changed their defaults and it looked like it would match this behavior, I'll run pr ci on windows

bodo/io/parquet_pio.py
IsaacWarren: It's just copying whatever ordered was set to in the original type

bodo/io/parquet_pio.py
IsaacWarren: Yeah I agree the duplication isn't good, I changed it because it seems that arrow changed how they infer parquet dataset schemas and they weren't unifying across files so if the column in the first file is all null the schema is inferred to be null for that column, for this change if I wrap it to only run on rank 0 does that work?

bodo/io/parquet_pio.py
ehsantn: `pa.dictionary` sets `ordered=False` by default looks like. Am I missing something?

bodo/io/parquet_pio.py
ehsantn: There are other differences too. Also, `get_parquet_dataset` is used in other places as well. What about adding the schema unification code to this function to match previous Arrow behavior?

bodo/io/parquet_pio.py
IsaacWarren: If it infers the dictionary type is ordered we want to propagate that right? Just change the index type

ehsantn: Not a big deal.

bodo/io/parquet_pio.py
IsaacWarren: Ok, I refactored to use get_parquet_dataset and a new function for unifying null columns, I tried to add the functionality to get_parquet_dataset and it does it but only if getting row counts, the logic is pretty entangled at this point so I just added this utility function to use on the resulting dataset if we want this behavior

bodo/io/parquet_pio.py
```diff
@@ -1542,9 +1537,13 @@ def get_dataset_unify_nulls(
...
     # If there are nulls in the schema, inspect the fragments
     # until the null columns can be resolved to a non-null type.
+    row_count = 0
     for piece, frag in zip(pieces, dataset_.get_fragments()):
         unify_fragment_schema(dataset, piece, frag)
+        row_count += piece._bodo_num_rows
         if not any(pa.types.is_null(typ) for typ in dataset.schema.types):
             break
 
+    unify_schemas_across_ranks(dataset, row_count)
```
ehsantn: Maybe avoid running this if we are on the spawner (check MPI size is 1 or something)? This should work but just in case it has issues or overheads.

ehsantn (approved):
Thanks for the fixes.

bodo/io/parquet_pio.py
ehsantn: Also, let's test reading some Parquet files to make sure things work and there is no regression.

.github/workflows/pr_ci.yml
```diff
@@ -118,7 +118,7 @@ jobs:
     with:
       batch: ${{ matrix.batch }}
       total-batches: 12
-      pytest-marker: "not slow and not weekly and not iceberg or smoke"
```
scott-routledge2, line 121: I think this change needs to be added to `azure-release-pipelines.yml` as well?

bodo/pandas/base.py
```diff
@@ -112,19 +112,37 @@ def read_parquet(
     filters=None,
     **kwargs,
 ):
-    from bodo.io.parquet_pio import get_dataset_unify_nulls
+    from bodo.io.parquet_pio import get_parquet_dataset, parquet_dataset_unify_nulls
```
scott-routledge2: Maybe we can make these imports top level since not importing JIT?

bodo/ir/iceberg_ext.py
```diff
@@ -1082,54 +1082,73 @@ def get_rtjf_col_min_max_unique_map(
...
-                            pie.LessThan(col, literal(max))
-                            if op == "<"
-                            else pie.LessThanOrEqual(col, literal(max)),
+                            rtjf_filters, pie.In(col, literal(unique_vals))
                         )
-
-            combined_filters = pie.And(file_filters, rtjf_filters)
-
-    return combined_filters
+                        print(rtjf_filters, unique_vals)
```
scott-routledge2:
```suggestion
```

bodo/io/iceberg/file_io.py
IsaacWarren: Looks like we still need to check for this unfortunately

scott-routledge2 (approved):
Thanks @IsaacWarren ! Looks good to me overall

bodo/ir/iceberg_ext.py
IsaacWarren: Thank you!

bodo/io/parquet_pio.py
IsaacWarren: Confirmed performance on nyc taxi locally is the same

PR #828, opened 2025-09-15
Separate JIT dependency from testing utils

scott-routledge2:

Separate JIT dependency from testing utilities (_test_equal and check_func) by ensuring codepaths with DataFrames do not import JIT unless where necessary.

## Testing strategy
Existing tests, manual checks JIT is imported later in tests that use _test_equal/check_func

## Checklist

### DIFF
--- a/bodo/tests/utils.py
+++ b/bodo/tests/utils.py
@@ -1334,7 +1334,7 @@ def _test_equal(
                             if isinstance(a, float) and np.isnan(a)
                             else a
                         ).values,
-                        bodo_out.dtype.pyarrow_dtype,
+                        pa_type,
                     ),
                     py_out.index,
                     bodo_out.dtype,


bodo/tests/utils.py
```diff
@@ -1484,7 +1533,7 @@ def _test_equal(
     elif isinstance(py_out, tuple) or (
         isinstance(py_out, list)
         and len(py_out) > 0
-        and is_distributable_typ(bodo.typeof(py_out))
```
scott-routledge2, line 1487: Not reached by any of the current DF library tests. If needed we could adapt this condition to not use `bodo.typeof` or use it conditionally.

bodo/tests/utils.py
```diff
@@ -1270,7 +1334,7 @@ def _test_equal(
                             if isinstance(a, float) and np.isnan(a)
                             else a
                         ).values,
-                        bodo.typeof(bodo_out.values),
+                        bodo_out.dtype.pyarrow_dtype,
```
ehsantn: Is Bodo output always in Arrow dtype now?

ehsantn (approved):
Looks good to me. Thanks @scott-routledge2. Some issues may pop up with these changes usually but they should be easy to fix.

bodo/tests/utils.py
scott-routledge2: Not sure if this is true for all compiler tests, in this codepath it's checking bodo_out is arrow type and python type is not and the arrow type is nested.

PR #829, opened 2025-09-15
Support Series.where()

ehsantn:
## Checklist

### DIFF
--- a/bodo/pandas/series.py
+++ b/bodo/pandas/series.py
@@ -16,6 +16,7 @@ import pyarrow as pa
 from pandas._libs import lib
 from pandas._typing import (
     Axis,
+    Level,
     SortKind,
     ValueKeyFunc,
 )
@@ -35,6 +36,7 @@ from bodo.pandas.plan import (
     AggregateExpression,
     ArithOpExpression,
     ArrowScalarFuncExpression,
+    CaseExpression,
     ColRefExpression,
     ComparisonOpExpression,
     ConjunctionOpExpression,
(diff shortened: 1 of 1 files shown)


DrTodd13 (approved):
Looks good. thanks

bodo/pandas/series.py
```diff
@@ -1349,6 +1352,102 @@ def quantile(self, q=0.5, interpolation=lib.no_default):
...
+        if lhs is None and other is None:  # pragma: no cover
+            raise BodoLibNotImplementedException(
+                "Series.where operation arguments should have the same dataframe source."
+            )
+        lhs, cond = match_binop_expr_source_plans(lhs, cond)
+        if lhs is None and cond is None:  # pragma: no cover
+            raise BodoLibNotImplementedException(
+                "Series.where operation arguments should have the same dataframe source."
+            )
+        lhs, other = match_binop_expr_source_plans(lhs, other)
```
scott-routledge2, line 1430: Why do we need to match lhs, other twice here?

bodo/pandas/physical/project.h
```diff
@@ -116,6 +116,21 @@ class PhysicalProjection : public PhysicalProcessBatch {
                 } else {
                     col_names.emplace_back("conjunction");
                 }
+            } else if (expr->type == duckdb::ExpressionType::CASE_EXPR) {
+                auto& case_expr = expr->Cast<duckdb::BoundCaseExpression>();
+
+                // TODO(ehsan): cast to common type of inputs if needed to match
```
scott-routledge2, line 122: What happens if the inputs aren't all the same type?

scott-routledge2 (approved):
LGTM, thanks @ehsantn

bodo/pandas/series.py
ehsantn: To make sure all 3 values have a matching source. lhs matches cond in the second call.

bodo/pandas/physical/project.h
ehsantn: Schema types may not match the actual tables being passed around the pipeline. It can potentially cause issues if some buffer allocation type is wrong, but I think there is casting functionality in some code paths that avoids issues automatically.

PR #830, opened 2025-09-16
Remove JIT dep in memory leak check and reorganize tests

scott-routledge2:

* Runs DataFrames tests that do not have jit dependencies in separate sessions on CI to avoid importing JIT before a non-jit test
* Removes dependency on numba/compiler in `memory_leak_check`
* Once scatter/gather don't use JIT, we can remove a majority of the tests from the "JIT dependency"

## Checklist


bodo/libs/_array.cpp
```diff
@@ -4,6 +4,7 @@
 
 #include <arrow/api.h>
 #include <arrow/python/pyarrow.h>
+#include <longobject.h>
```
IsaacWarren: What is this for?

bodo/tests/conftest.py
```diff
@@ -259,6 +269,28 @@ def pytest_collection_modifyitems(items):
...
+        ]
+
+        for item in no_jit_dep:
+            if (
+                hasattr(item, "fixturenames")
+                and "jit_import_check" not in item.fixturenames
+            ):
+                item.fixturenames = ["jit_import_check"] + item.fixturenames
+
+        items[:] = no_jit_dep + jit_dep
```
IsaacWarren: Generally, depending on test ordering is bad practice, maybe we should run them in separate workers in CI?

IsaacWarren (approved):
Thanks Scott

bodo/libs/_array.cpp
scott-routledge2: Added by my IDE, I'll remove

bodo/tests/conftest.py
scott-routledge2: Good point, although there are very few tests that don't depend on JIT right now because of scatter/gather. Maybe we can separate them after that work and have some idea of how many tests are in each category. Another option could be to just run two separate commands in the same runners.

ehsantn (approved):
Thanks @scott-routledge2 .

PR #840, opened 2025-09-22
Add check for distributed function before calling `get_nodes_first_ranks`

scott-routledge2:

In some of our tests we have this pattern of calling a function that is not distributed or a numba njit from only one rank, which causes hangs because `bodo.libs.distributed_api.get_nodes_first_ranks` is expected to be called from all ranks.

## Checklist

### DIFF
--- a/bodo/numba_compat.py
+++ b/bodo/numba_compat.py
@@ -1485,8 +1485,11 @@ def Dispatcher_compile(self, sig):
             else:
                 # Even when not on platform, it's best to minimize I/O contention, so we
                 # write cache files from one rank on each node.
-                first_ranks = bodo.libs.distributed_api.get_nodes_first_ranks()
-                if bodo.get_rank() in first_ranks:
+                # However, when result is not a distributed function (i.e. distributed=False),
+                # avoid calling get_nodes_first_ranks for the case where only one rank is
+                # calling the function, which will hang.
+                is_distributed_impl = cres.metadata.get("distributed_diagnostics", None) is not None
+                if not is_distributed_impl or (bodo.get_rank() in bodo.libs.distributed_api.get_nodes_first_ranks()):
                     self._cache.save_overload(sig, cres)
             return cres.entry_point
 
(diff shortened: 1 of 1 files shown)


bodo/numba_compat.py
```diff
@@ -1485,8 +1485,14 @@ def folded(args, kws):
             else:
                 # Even when not on platform, it's best to minimize I/O contention, so we
                 # write cache files from one rank on each node.
-                first_ranks = bodo.libs.distributed_api.get_nodes_first_ranks()
-                if bodo.get_rank() in first_ranks:
+                # However, when result is not a distributed function (i.e. distributed=False),
+                # avoid calling get_nodes_first_ranks for the case where only one rank is
+                # calling the function, which will hang.
+                is_distributed_impl = cres.metadata.get("distributed_diagnostics", None) is not None
```
scott-routledge2: Wondering if there is a better way to check if a result should not be distributed?

IsaacWarren: Maybe use "is_return_distributed" instead? That's what we use in the workers to figure out how to return the value to the spawner, see worker.py:498

IsaacWarren (approved):
Thanks Scott

bodo/numba_compat.py
scott-routledge2: The issue with this check being here in general I think it might be too conservative and would essentially never write results on rank 0 only for most internal functions and overloads that we use. I'm not sure how big of a deal that would be for compilation time?

ehsantn: One option is to always write for rank 0 since it's always a first rank for its node. This should cover the common cases. What do you think?

scott-routledge2: Yeah I think I like that approach better, will update the PR

IsaacWarren: That assumes the cache location is a shared location though, I don't think that's true if someone sets up their own cluster

scott-routledge2: So maybe just update the tests then?

PR #846, opened 2025-09-24
BSE-5119: torch trainer

IsaacWarren:
As title, includes the new torch trainer, model preparation and docs.

## Testing strategy
Unit test

## User facing changes
New functionality

## Checklist

### DIFF
--- a/bodo/ai/train.py
+++ b/bodo/ai/train.py
@@ -48,26 +48,26 @@ def _init_process_group():
     else:
         device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
     pytorch_rank = MPI.COMM_WORLD.Get_rank()
+
+    gpu_ranks = get_gpu_ranks()
     if device is None or device == torch.device("cpu"):
         device = torch.device("cpu")
     else:
         # Assign each rank to an accelerator
         mpi_rank = MPI.COMM_WORLD.Get_rank()
-        gpu_ranks = get_gpu_ranks()
         num_local_gpus = len(gpu_ranks) // get_num_nodes()
         if mpi_rank in gpu_ranks:
             pytorch_rank = gpu_ranks.index(mpi_rank)
             if hasattr(torch, "accelerator"):
-                device = torch.accelerator.set_device_idx(pytorch_rank % num_local_gpus)
+                torch.accelerator.set_device_index(pytorch_rank % num_local_gpus)
+                device = torch.accelerator.current_accelerator()
             else:
+                torch.cuda.set_device(pytorch_rank % num_local_gpus)
                 device = torch.device(f"cuda:{pytorch_rank % num_local_gpus}")
-
         else:
             pytorch_rank = None
     npes = (
-        len(get_gpu_ranks())
-        if device != torch.device("cpu")
-        else MPI.COMM_WORLD.Get_size()
+        len(gpu_ranks) if device != torch.device("cpu") else MPI.COMM_WORLD.Get_size()
     )
 
     backend = torch.distributed.get_default_backend_for_device(device)
@@ -129,12 +129,6 @@ def prepare_model(
     if pytorch_rank is None:
         return None
 
-    model = model.to(device)
-
-    # No need to wrap the model if only one process is used
-    if pytorch_world_size == 1:
-        return model
-
     if parallel_strategy is not None:
         assert parallel_strategy in ["ddp", "fsdp"], (
             "parallel_strategy should be either 'ddp' or 'fsdp'"
@@ -149,5 +143,5 @@ def prepare_model(
             from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
 
             model = FSDP(model, **parallel_strategy_kwargs)
-        model.to(pytorch_rank)
+    model.to(device)
     return model


bodo/libs/distributed_api.py
```diff
@@ -2792,7 +2792,7 @@ def get_num_gpus(framework="torch"):  # pragma: no cover
         try:
             import torch
 
-            return torch.cuda.device_count()
+            return torch.accelerator.device_count()
```
scott-routledge2: accelerator requires torch>2.5

bodo/tests/test_df_lib/test_ai_funcs.py
```diff
@@ -390,3 +391,78 @@ def test_embed_bedrock_default_formatter():
...
+                self.linear1 = nn.Linear(2, 32)
+                self.relu = nn.ReLU()
+                self.linear2 = nn.Linear(32, 1)
+
+            def forward(self, x):
+                return self.linear2(self.relu(self.linear1(x)))
+
+        model = SimpleModel()
+        model = bodo.ai.train.prepare_model(model, parallel_strategy="ddp")
+        gpu_ranks = bodo.get_gpu_ranks()
```
scott-routledge2, line 426: Would the user need to add this rebalancing code themselves? Could we potentially do this for them or provide a function similar to `ray.train.torch.prepare_data_loader()`

docs/docs/api_docs/dataframe_lib/ai.md
```diff
@@ -0,0 +1,131 @@
...
+        # If we're using an accelerator, rebalance data to match GPU ranks
+        data = bodo.rebalance(data, dests=gpu_ranks)
+
+    # Train on data
+    criterion = nn.MSELoss()
+    optimizer = torch.optim.SGD(model.parameters(), lr=0.01)
+    for epoch in range(config.get("epochs", 5)):
+        batch_size = config.get("batch_size", 2)
+        for i in range(0, len(data), batch_size):
+            batch = data[i : i + batch_size]
```
scott-routledge2, line 105: I was wondering if we should create some Dataset sub-classes to simplify the training code that needs to be written a bit. If the subclass doesn't fit the users needs they can always create their own subclass or write the training loop using raw dataframes

For example for a supervised learning task:
``` py
class DataFrameDataset(Dataset):
    def __init__(self, dataframe, label_col, features_col):
        self.dataframe = dataframe
        self.label_col = label_col
        self.feature_col = features_col

    def __len__(self):
        return len(self.dataframe)

    def __getitem__(self, idx):
        label = self.dataframe.iloc[idx][self.label_col]
        features = self.dataframe.iloc[idx][self.feature_col]
        return torch.tensor(features), torch.tensor(label)

dataset = DataFrameDataset(df, label_col="label", features_col="features")
dataloader = DataLoader(dataset, batch_size=2, shuffle=True)

for X,y in dataloader:
    # training code here
```

bodo/__init__.py
```diff
@@ -380,6 +380,12 @@ def random_shuffle(*args, **kwargs):
     from bodo.libs.distributed_api import random_shuffle
     return random_shuffle(*args, **kwargs)
 
+def get_gpu_ranks(*args, **kwargs):
```
scott-routledge2, line 389: If this is only called on workers then it might cause hangs if the compiler wasn't already imported on the spawner

bodo/tests/test_df_lib/test_ai_funcs.py
```diff
@@ -390,3 +391,78 @@ def test_embed_bedrock_default_formatter():
...
+        model_device = next(model.parameters()).device
+        if model_device.type != "cpu":
+            # If we're using an accelerator, rebalance data to match GPU ranks
+            data = bodo.rebalance(data, dests=gpu_ranks)
+
+        # train on data
+        criterion = nn.MSELoss()
+        optimizer = torch.optim.SGD(model.parameters(), lr=0.01)
+        for epoch in range(config.get("epochs", 5)):
+            batch_size = config.get("batch_size", 2)
```
scott-routledge2, line 440: Wouldn't this cause hangs if the batch size is not evenly divisible by the number of ranks (so some ranks execute more batches than others)?

docs/docs/api_docs/dataframe_lib/ai.md
```diff
@@ -0,0 +1,131 @@
...
+
+    model_device = next(model.parameters()).device
+    if model_device.type != "cpu":
+        # If we're using an accelerator, rebalance data to match GPU ranks
+        data = bodo.rebalance(data, dests=gpu_ranks)
+
+    # Train on data
+    criterion = nn.MSELoss()
+    optimizer = torch.optim.SGD(model.parameters(), lr=0.01)
+    for epoch in range(config.get("epochs", 5)):
```
scott-routledge2, line 102: I think the semantics of this training loop are different than vanilla FSDP since we are not doing any shuffling, and data samples always map to the same rank. Maybe we need our own version of DistributedSampler? https://github.com/pytorch/pytorch/blob/ba56102387ef21a3b04b357e5b183d48f0afefc7/torch/utils/data/distributed.py#L107

scott-routledge2 (approved):
Thanks Isaac! Overall looks good to me.

bodo/ai/train.py
```diff
@@ -0,0 +1,126 @@
...
+
+def _get_open_port():
+    """
+    Finds and returns an available open TCP port.
+    """
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    try:
+        s.bind(("", 0))  # Bind to an ephemeral port (port 0)
+        port = s.getsockname()[1]  # Get the assigned port number
+        return port
```
DrTodd13, line 33: Is this safe?  There's a time window after closing the socket that some other process could open that port number.

bodo/tests/test_df_lib/test_ai_funcs.py
```diff
@@ -390,3 +391,78 @@ def test_embed_bedrock_default_formatter():
...
+                model.module
+                if isinstance(model, torch.nn.parallel.DistributedDataParallel)
+                else model
+            )
+            torch.distributed.checkpoint.state_dict_saver.save(
+                {"model_state_dict": base_model.state_dict()},
+                checkpoint_id=config["checkpoint_dir"],
+            )
+
+            print(f"Epoch {epoch}, Loss: {loss.item()}")
```
DrTodd13: Is this needed?

DrTodd13 (approved):
Thanks.

bodo/ai/train.py
IsaacWarren: That's true but I think it's better than what pytorch recommends which is just a hardcoded port assignment. I'll wrap in a retry

bodo/__init__.py
IsaacWarren: Why? Do we have a solution?

scott-routledge2: When the compiler is imported on the workers, `scatterv` gets replaced with `bodo.libs.distributed_api.scatterv`. The next time someone goes to scatter a dataframe, the spawner will call `scatterv_nojit` and the workers call `bodo.libs.distributed_api.scatterv`, which hangs. One solution could be to import the compiler before `submit_func_to_workers` since training is not going to be very responsive anyways.

PR #848, opened 2025-09-26
Separate JIT dependencies in testing

scott-routledge2:

Separate tests into JIT and Non-JIT and minor bug fixes.

## User facing changes

Compiler is not imported in the generic fallback path.
Minor bug fixes

## Checklist


bodo/spawn/spawner.py
```diff
@@ -533,6 +546,8 @@ def _get_arg_metadata(self, arg, arg_name, is_replicated, dist_flags):
         # Import compiler lazily
         import bodo.decorators  # isort:skip # noqa
 
+        self.import_compiler_on_workers()
```
scott-routledge2, line 555: Compiler should have already been imported above ^

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -337,7 +336,7 @@ def test_filter_pushdown(datapath, file_path, op):
     )
 
 
-@pytest_mark_spawn_mode
```
scott-routledge2, line 340: If we mark this test as spawn mode it will be skipped when running DataFrame Library tests and spawn mode tests don't enable DataFrame library, so the test won't be run properly

bodo/pandas/frame.py
```diff
@@ -340,7 +341,14 @@ def __repr__(self):
     @property
     def index(self):
         self.execute_plan()
-        return super().index
+        index = super().index
+
+        # We use BODO_NONE_DUMMY as a placeholder for missing multi-index names
+        # so that they round trip correctly from Arrow.
+        if isinstance(index, pd.MultiIndex):
```
scott-routledge2, line 348: This is probably not right because if we fallback then we still might have the wrong name (see also pandas/utils.py) in the index, but I couldn't get it to work inside of `execute_plan`

bodo/spawn/spawner.py
```diff
@@ -101,6 +101,15 @@ def get_num_workers():
     return n_pes
 
 
+def _not_all_lazy_plan_args(args: tuple[pt.Any], kwargs: dict[str, pt.Any]) -> bool:
+    """Check if any arg or kwarg is not a LazyPlan"""
```
ehsantn: Let's add some context in the comment on why this is necessary.

bodo/spawn/spawner.py
ehsantn: Let's add comment that this is just in case.

bodo/spawn/spawner.py
```diff
@@ -876,6 +891,11 @@ def spawn_process_on_nodes(
         """Spawn a process on all workers and return a WorkerProcess object"""
         assert not self._is_running, "spawn_process_on_nodes: already running"
 
+        # Import compiler for handle_spawn_process on workers
```
ehsantn: Why? Let's add context.

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -644,6 +644,7 @@ def test_head_pushdown(datapath):
     assert len(bodo_df2) == 3
 
 
+@pytest.mark.jit_dependency
```
ehsantn: Why does this test need JIT?

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -2290,6 +2295,7 @@ def test_Series_constructor(index_val):
     _test_equal(pd_S, bodo_S, check_pandas_types=False)
 
 
+@pytest.mark.skip("TODO fix decimal comparison")
```
ehsantn: This is not fixed by my last PRs?

bodo/tests/test_join.py
```diff
@@ -1959,7 +1957,6 @@ def test_impl():
     )
 
 
-@pytest.mark.df_lib
```
ehsantn, line 1962: I assume this is just redundant?

bodo/tests/test_parquet_read.py
```diff
@@ -64,12 +64,11 @@ def test_impl():
         pytest.param("bool_nulls.pq", id="bool_with_nulls"),
         pytest.param("list_str_arr.pq", id="list_str_arr"),
         pytest.param("list_str_parts.pq", id="list_str_parts"),
-        pytest.param("decimal1.pq", id="decimal"),
+        pytest.param("decimal1.pq", id="decimal", marks=pytest.mark.skip("TODO:FIXME")),
```
ehsantn: Same about decimal PR.

bodo/tests/test_parquet_read.py
```diff
@@ -2954,10 +2924,11 @@ def impl(path):
             check_logger_msg(stream, "(ds.field('A').isin(f0))")
 
 
-@pytest.mark.skipif(
-    bodo.test_dataframe_library_enabled,
-    reason="[BSE-4768] Enable filter pushdown tests.",
-)
+# @pytest.mark.skipif(
```
ehsantn: Remove commented code?

bodo/spawn/spawner.py
scott-routledge2: Because it calls `get_nodes_first_ranks` which requires compiler imported. As far as I could tell we only use it in testing? So didn't think it was a big deal for now.

bodo/tests/test_df_lib/test_end_to_end.py
scott-routledge2: It doesn't, forgot to remove

bodo/tests/test_join.py
scott-routledge2: This test doesn't use check_func, so marking it df_lib won't test anything

bodo/tests/test_df_lib/test_end_to_end.py
scott-routledge2: Oh I think it was fixed, let me check

bodo/spawn/spawner.py
ehsantn: Let's just add a comment that it uses `get_nodes_first_ranks`.

PR #852, opened 2025-09-26
Avoid JIT in Iceberg for DataFrames

ehsantn:

Avoids JIT import in Iceberg read/write of Bodo DataFrames.

## Testing strategy

Existing tests.

## User facing changes

Faster response times.

## Checklist

### DIFF
--- a/bodo/__init__.py
+++ b/bodo/__init__.py
@@ -342,6 +342,24 @@ def scatterv(data, send_counts=None, warn_if_dist=True, root=0, comm=None):
     return scatterv_nojit(data, root, comm)
 
 
+def get_start(total_size, pes, rank):  # pragma: no cover
+    """Same as bodo.libs.distributed_api.get_start() but avoiding JIT compiler import
+    here.
+    """
+    res = total_size % pes
+    blk_size = (total_size - res) // pes
+    return rank * blk_size + min(rank, res)
+
+
+def get_end(total_size, pes, rank):  # pragma: no cover
+    """Same as bodo.libs.distributed_api.get_end() but avoiding JIT compiler import
+    here.
+    """
+    res = total_size % pes
+    blk_size = (total_size - res) // pes
+    return (rank + 1) * blk_size + min(rank + 1, res)
+
+
 def get_nodes_first_ranks(*args, **kwargs):
     # Import compiler lazily
     import bodo.decorators
--- a/bodo/io/iceberg/common.py
+++ b/bodo/io/iceberg/common.py
@@ -9,7 +9,6 @@ import typing as pt
 from urllib.parse import parse_qs, urlparse
 
 import requests
-from numba.core import types
 from pyarrow.fs import FileSystem, FSSpecHandler
 
 from bodo.spawn.utils import run_rank0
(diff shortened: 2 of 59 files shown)


bodo/io/iceberg/common.py
```diff
@@ -50,6 +49,8 @@ def verify_pyiceberg_installed():
     try:
         return importlib.import_module("pyiceberg")
     except ImportError:
+        from bodo.utils.utils import BodoError
```
scott-routledge2, line 52: Should this raise a ValueError here too? I think this will fail if the compiler wasn't imported.

bodo/spawn/utils.py
```diff
@@ -321,3 +324,45 @@ def _get_data_sample(data):
     raise ValueError(
         "_get_data_sample only supports DataFrame, Series and ArrowExtensionArray input"
     )
+
+
+def run_rank0(func: Callable, bcast_result: bool = True, result_default=None):
```
IsaacWarren, line 329: Why put this here? It's still usable if spawn mode is disabled right?

bodo/__init__.py
```diff
@@ -342,6 +342,24 @@ def scatterv(data, send_counts=None, warn_if_dist=True, root=0, comm=None):
     return scatterv_nojit(data, root, comm)
 
 
+def get_start(total_size, pes, rank):  # pragma: no cover
+    """Same as bodo.libs.distributed_api.get_start() but avoiding JIT compiler import
+    here.
```
IsaacWarren, line 347: Can we replace bodo.libs.distributed_api.get_start to avoid duplication then?

bodo/io/iceberg/read_metadata.py
```diff
@@ -191,7 +191,8 @@ def get_iceberg_file_list_parallel(
                 ", ".join(x.path for x in pq_infos[:ICEBERG_TRACING_NUM_FILES_TO_LOG]),
             )
     except Exception as exc:  # pragma: no cover
-        raise BodoError(
+        # TODO: raise BodoError in case of compiler (not dataframe library)
+        raise ValueError(
```
IsaacWarren, line 195: Is there a reason BodoError depends on the compiler? I thought it was just a renamed ValueError?

IsaacWarren (approved):
Thanks Ehsan

bodo/spawn/utils.py
ehsantn: Yes, just had to be in a file that doesn't import JIT. This file is basically parallelism utilities. We need to reorganize the code base for JIT/no-JIT distinction.

bodo/__init__.py
ehsantn: That one is used inside JIT and don't want to mess with it right now.

bodo/io/iceberg/read_metadata.py
ehsantn: It is a subclass of `NumbaError`. The compiler pipeline treats it differently in critical places.

scott-routledge2 (approved):
Thanks @ehsantn !

bodo/io/iceberg/common.py
ehsantn: This only happens in the compiler code path since dataframe library import pyiceberg upfront.

PR #854, opened 2025-09-29
Set output schema in no-JIT gather/scatter

ehsantn:

As title. Fixes some existing issues with no-JIT gather/scatter.

## Testing strategy

Existing tests.

## User facing changes

Bug fix.

## Checklist

### DIFF
--- a/bodo/pandas/utils.py
+++ b/bodo/pandas/utils.py
@@ -506,7 +506,7 @@ def get_n_index_arrays(index):
         raise TypeError(f"Invalid index type: {type(index)}")
 
 
-def df_to_cpp_table(df):
+def df_to_cpp_table(df, return_schema=False):
     """Convert a pandas DataFrame to a C++ table pointer with column names and
     metadata set properly.
     """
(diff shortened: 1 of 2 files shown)


bodo/pandas/utils.py
```diff
@@ -537,6 +537,9 @@ def df_to_cpp_table(df):
     if any(col.num_chunks == 0 for col in arrow_table.columns):
         arrow_table = pa.Table.from_arrays(new_columns, schema=arrow_table.schema)
 
+    if return_schema:
```
DrTodd13, line 540: Isn't the normal pattern here to always return schema and then do "ret, _" in cases wehre you don't need it, particularly for cases like this where the function isn't used very many times?

DrTodd13 (approved):
I have a preference noted below but otherwise seems fine.  Thanks.

scott-routledge2 (approved):

bodo/pandas/utils.py
ehsantn: This pattern is also common. Just didn't want to change existing code. Will do in follow ups.

PR #855, opened 2025-09-30
Convert to BodoDataFrame/BodoSeries on fallback

scott-routledge2:

Generate BodoSeries and BodoDataFrames after running an unsupported function in Pandas. Also adds extra error checking to from_pandas

## User facing changes

Better error messages in from_pandas. Result of fallback methods returns BodoDataFrames/Series

## Checklist

### DIFF
--- a/bodo/pandas/base.py
+++ b/bodo/pandas/base.py
@@ -79,17 +79,15 @@ def from_pandas(df):
         raise BodoLibNotImplementedException(
             "from_pandas(): Hierarchical column names are not supported in Bodo yet."
         )
-
     new_columns = []
     for c in df.columns:
         if isinstance(df[c], pd.DataFrame):
             raise BodoLibNotImplementedException(
-                f"from_pandas(): Duplicate column names are not supported: '{c}'."
+                f"from_pandas(): Duplicate column name: '{c}'."
             )
         elif not isinstance(c, str):
             warnings.warn(
-                f"The column name '{c}' with type {type(c)} was converted to string "
-                + "and will not round trip correctly."
+                f"The column name '{c}' with type {type(c)} was converted to string."
             )
         new_columns.append(str(c))
 
@@ -113,7 +111,7 @@ def from_pandas(df):
         # TODO: add specific unsupported columns to message.
         raise BodoLibNotImplementedException(
             "from_pandas(): Could not convert DataFrame to Bodo: "
-            + "Unsupported datatype encountered in one or more columns:"
+            + "Unsupported datatype encountered in one or more columns: "
             + str(e)
         )
 
--- a/bodo/pandas/base.py
+++ b/bodo/pandas/base.py
@@ -54,6 +54,7 @@ from bodo.pandas.scalar import BodoScalar
 from bodo.pandas.series import BodoSeries, _get_series_func_plan
 from bodo.pandas.utils import (
     BODO_NONE_DUMMY,
+    BodoDictionaryTypeInvalidException,
     BodoLibNotImplementedException,
     arrow_to_empty_df,
     check_args_fallback,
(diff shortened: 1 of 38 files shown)


bodo/pandas/utils.py
```diff
@@ -1143,10 +1207,16 @@ def silenced_method(*args, **kwargs):
             with warnings.catch_warnings():
                 warnings.simplefilter("ignore", category=BodoLibFallbackWarning)
                 try:
-                    return attr(*args, **kwargs)
+                    with FallbackContext():
+                        py_res = attr(*args, **kwargs)
+
+                    # Convert objects to Bodo before returning them to the user.
+                    if FallbackContext.is_top_level():
+                        return convert_to_bodo(py_res)
```
scott-routledge2, line 1221: My observation was that Pandas methods call a lot of internal functions we do not support (example: `xs`, `copy`), so we can keep the DataFrame as Pandas for the internal calls and only convert when returning back to the user. 

This is also currently hiding a small bug in a lot of the tests that I haven't figured out yet, but seemed minor to me:
``` py
df = pd.DataFrame({"A": [1, 2, 3], "B": ["a", "b", "c"]}, index = [1,2,3])
bdf = bd.from_pandas(df)

bdf1 = bdf.rename_axis("index123")
bdf2 = bdf1.copy()
print("bodo result: ", bdf2.index.name)

pdf1 = df.rename_axis("index123")
pdf2 = pdf1.copy()
print("pandas result: ", pdf2.index.name)
```
Pandas result: "index123", Bodo result: `None` (index name doesn't propagate in some places)

bodo/tests/test_lazy/test_bodo_series.py
```diff
@@ -265,6 +265,9 @@ def test_slice(single_pandas_managers, head_s, collect_func, del_func):
     """Tests that slicing returns the correct value and does not trigger data fetch unnecessarily"""
     lazy_manager, pandas_manager = single_pandas_managers
 
+    if pandas_manager == SingleArrayManager:
```
scott-routledge2, line 269: These seem to be an existing issues that was exposed by this PR now that there is more conversion between pandas and bodo happening. I can open a followup to investigate but in my opinion it is not as big of a priority since BlockManager is the default and ArrayManager will be removed in Pandas 3.0

bodo/pandas/base.py
```diff
@@ -73,6 +75,24 @@ def from_pandas(df):
...
 
+    if isinstance(df.columns, pd.MultiIndex):
+        raise BodoLibNotImplementedException(
+            "from_pandas(): Hierarchical column names are not supported in Bodo yet."
+        )
+    new_columns = []
+    for c in df.columns:
+        if isinstance(df[c], pd.DataFrame):
+            raise BodoLibNotImplementedException(
+                f"from_pandas(): Duplicate column name: '{c}'."
```
DrTodd13: How does df[c] ever become itself a dataframe and why is that labelled as a duplicate column?

bodo/tests/test_spawn/test_spawn_mode.py
```diff
@@ -445,6 +445,7 @@ def test_spawn_input():
     assert sub.returncode == 0
 
 
+@pytest.mark.skip("TODO: Fix flakey test on CI.")
```
ehsantn: Let's open an issue and put on oncall board not to forget.

bodo/pandas/base.py
scott-routledge2: `df[c]` returns a dataframe with all columns named "c" in the case of duplicates

bodo/pandas/base.py
```diff
@@ -73,6 +75,24 @@ def from_pandas(df):
     if not isinstance(df, pd.DataFrame):
         raise TypeError("Input must be a pandas DataFrame")
 
+    if isinstance(df.columns, pd.MultiIndex):
+        raise BodoLibNotImplementedException(
+            "from_pandas(): Hierarchical column names are not supported in Bodo yet."
+        )
+    new_columns = []
+    for c in df.columns:
+        if isinstance(df[c], pd.DataFrame):
```
ehsantn: Using `df.columns.has_duplicates` is simpler and more reliable. `columns` is an Index, which is sort of a set and should have this info internally I think.

ehsantn (approved):
Thanks @scott-routledge2!

PR #856, opened 2025-09-30
Speed up TPCH.

DrTodd13:

Traverse right-side of cross-product so CTE can find duplicate computations.
Run to_datetime as cfunc over a map instead of Python.

## Testing strategy

run_ci

## User facing changes

None

## Checklist

### DIFF
--- a/bodo/pandas/base.py
+++ b/bodo/pandas/base.py
@@ -492,12 +492,12 @@ def to_datetime(
     # 2. Series Case
     if (
         errors == "raise"
-        and dayfirst == False
-        and yearfirst == False
-        and utc == False
-        and unit == None
+        and dayfirst is False
+        and yearfirst is False
+        and utc is False
+        and unit is None
         and origin == "unix"
-        and cache == True
+        and cache is True
     ):
         # If only options supported by Bodo JIT then run as cfunc over map.
         import bodo.decorators  # isort:skip # noqa
@@ -505,7 +505,6 @@ def to_datetime(
         if format is None:
 
             def bodo_df_lib_to_datetime(x):
-                print("bodo_df_lib_to_datetime", x)
                 return pd.to_datetime(x)
 
             return arg.map(bodo_df_lib_to_datetime, na_action="ignore")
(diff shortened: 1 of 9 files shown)


DrTodd13:
Is there a better way to flag the tests that have to_datetime in them as now requiring JIT?

bodo/pandas/base.py
```diff
@@ -490,14 +490,35 @@ def to_datetime(
...
+        import bodo.decorators  # isort:skip # noqa
+        from bodo.utils.utils import bodo_spawn_exec
+
+        # Declare function to be compiled to run to_datetime over series.
+        func = "def bodo_to_datetime(x):\n"
+        # Embed format string as constant in function.
+        func += f"    return pd.to_datetime(x, format='{in_kwargs['format']}')\n"
+        # Create the function from string.
+        to_datetime_func = bodo_spawn_exec(func, {"pd": pd}, {}, __name__)
+        return arg.map(to_datetime_func)
```
IsaacWarren: Would it be possible to use a c++ kernel for this maybe using Arrow instead of the compiler? Not saying it needs done now but maybe a comment for a followup

IsaacWarren (approved):
Thanks @DrTodd13

scott-routledge2:
> Is there a better way to flag the tests that have to_datetime in them as now requiring JIT?

Not that I can think of? Maybe we could change the test command to run each test module as a separate session? (I think `runtests.py` does this. That way would could tell the first function to import JIT in each file at least.

bodo/pandas/base.py
```diff
@@ -490,14 +490,35 @@ def to_datetime(
...
+        # If only options supported by Bodo JIT then run as cfunc over map.
+        import bodo.decorators  # isort:skip # noqa
+        from bodo.utils.utils import bodo_spawn_exec
+
+        # Declare function to be compiled to run to_datetime over series.
+        func = "def bodo_to_datetime(x):\n"
+        # Embed format string as constant in function.
+        func += f"    return pd.to_datetime(x, format='{in_kwargs['format']}')\n"
+        # Create the function from string.
+        to_datetime_func = bodo_spawn_exec(func, {"pd": pd}, {}, __name__)
```
scott-routledge2: Maybe this needs `bodo.jit(func, cache=True)` here?

PR #860, opened 2025-10-01
Fix scatter bug for dict arrays

ehsantn:
As title. Quick bug fix to unblock @scott-routledge2 .


scott-routledge2 (approved):
Thanks for the quick fix!

IsaacWarren (approved):

PR #861, opened 2025-10-02
Add C++ backend for BodoSQL

ehsantn:
## Checklist

### DIFF
--- a/BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/application/PythonEntryPoint.kt
+++ b/BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/application/PythonEntryPoint.kt
@@ -14,6 +14,7 @@ import com.bodosql.calcite.table.BodoSQLColumn
 import com.bodosql.calcite.table.BodoSQLColumnImpl
 import com.bodosql.calcite.table.ColumnDataTypeInfo
 import com.bodosql.calcite.table.LocalTable
+import org.apache.calcite.rel.RelNode
 import org.apache.commons.lang3.exception.ExceptionUtils
 import java.util.Properties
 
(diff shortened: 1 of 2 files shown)


BodoSQL/bodosql/tests/test_cpp_backend.py
```diff
@@ -0,0 +1,27 @@
+"""
+Test BodoSQL C++ backend
+"""
+
```
IsaacWarren, line 4: IMO the best way to test this is to add a marker to existing bodosql tests that is run in "no fallback" mode something like "require_cpp_backend" so we can reuse all our existing tests and don't need new ones

BodoSQL/bodosql/context.py
```diff
@@ -1240,6 +1261,47 @@ def _get_pandas_code(
...
+                dynamic_params_list
+            )
+            java_named_params_map = create_java_named_parameter_type_map(
+                named_params_dict
+            )
+            java_plan = JavaEntryPoint.getOptimizedPlan(
+                generator, sql, java_params_array, java_named_params_map
+            )
+            plan = java_plan_to_python_plan(self, java_plan)
+            out = bodo.pandas.plan.execute_plan(plan)
```
IsaacWarren: Won't this plan the query twice? Once in calcite and once in duckdb?

IsaacWarren (commented):
Thanks Ehsan, my main concern is the double planning, any ideas on how to avoid that?

BodoSQL/bodosql/tests/test_cpp_backend.py
ehsantn: I agree. This is just for merging this PR and having some tests for initial stages of the project (similar to what we did for dataframe library).

BodoSQL/bodosql/context.py
ehsantn: Set a flag to avoid optimizing in DuckDB.

scott-routledge2 (approved):
LGTM, thanks @ehsantn

PR #868, opened 2025-10-08
Overload array dunder method to convert BodoDataFrames of floats/ints to ndarrays with the correct dtype

scott-routledge2:

As titled

## Testing strategy

New unit test added, E2E nightly

## User facing changes

Converting DataFrame representing a NxM matrix to an NDArray will return an NDarray with the equivalent dtype instead of object, (assuming all column dtypes are equal)

## Checklist

### DIFF
--- a/bodo/tests/test_df_lib/test_end_to_end.py
+++ b/bodo/tests/test_df_lib/test_end_to_end.py
@@ -3778,3 +3778,26 @@ def test_series_str_match():
         pmask,
         check_pandas_types=False,
     )
+
+
+def test_asarray_df():
+    """
+    Tests that np.asarray(df) produces the correct ndarray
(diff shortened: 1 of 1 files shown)


bodo/pandas/frame.py
```diff
@@ -338,6 +338,21 @@ def __repr__(self):
             warnings.simplefilter("ignore", category=BodoLibFallbackWarning)
             return super().__repr__()
 
+    def __array__(self, dtype=None, copy=None):
```
IsaacWarren, line 341: Should we do this for series too?

scott-routledge2: Series doesn't have the same issue from testing (probably because there is only one dtype)

PR #873, opened 2025-10-10
BSE-5132: prepare_dataset

IsaacWarren:

Add prepare dataset to simplify training loops, updates examples, tests and docs to use it.

## Testing strategy
Unit test and running pandas ddp locally and on 2 gpu instance

## User facing changes
Prepare_dataset to simplify training loops

## Checklist

### DIFF
--- a/examples/ML/pandas_ddp_example.py
+++ b/examples/ML/pandas_ddp_example.py
@@ -25,7 +25,7 @@ class BertDataset(torch.utils.data.Dataset):
     def __getitem__(self, idx):
         row = self.df.iloc[idx]
         texts = row["label"]
-        texts = torch.tensor(batch_texts)
+        texts = torch.tensor(texts)
 
         tokenized = row["tokenized"]
         input_ids = torch.tensor(tokenized["input_ids"])


bodo/tests/test_df_lib/test_ai_funcs.py
```diff
@@ -399,9 +399,9 @@ def test_torch_train():
 
     df = bd.DataFrame(
         {
-            "feature1": [1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0],
-            "feature2": [2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0],
-            "label": [3.0, 5.0, 7.0, 9.0, 11.0, 13.0, 15.0, 17.0],
+            "feature1": pd.array([1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0], dtype="float32"),
```
scott-routledge2, line 402: Having a column for each dimension might be awkward since there are typically hundred or thousands of dimensions and we don't have a convenient way of going from some column that's a list of input sequences e.g. tokens to a dataframe where each position in the sequence has it's own column? 

Maybe letting users pass a custom Dataset class would be better (we could still keep PandasDataset for convenience and users can subclass? Thoughts?

IsaacWarren: Let me see if it works with list arrays, because then all the inputs can be a single column. The custom dataset would be nice but complicates the shuffle, our random shuffle can only do dataframes and I'm not really sure how pytorch shuffling works

scott-routledge2: Maybe we could enforce some requirement on the custom dataset like it has to have a "to_dataframe" and "from_dataframe" attribute so they are still shuffle-able? If it works for list arrays then it's not as big of a deal and can definitely be a follow up for later.

examples/ML/pandas_ddp_example.py
```diff
@@ -182,22 +125,31 @@ def train_one_epoch(model, train_loader, loss_fn, optimizer):
...
+        acc = (output.logits.argmax(dim=1) == train_label).sum().item()
         total_ddp_acc_loss_train[0] += acc
         total_ddp_acc_loss_train[1] += batch_loss.item()
         total_ddp_acc_loss_train[2] += train_label.size(0)
 
-        model.zero_grad()
         batch_loss.backward()
         optimizer.step()
+        scheduler.step()
+        scheduler.step()
```
scott-routledge2: scheduler.step is called twice
```suggestion
```

DrTodd13 (approved):

examples/ML/pandas_ddp_example.py
```diff
@@ -1,101 +1,38 @@
...
-        self.device = device
 
     def __len__(self):
         return len(self.df)
 
     def __getitem__(self, idx):
         row = self.df.iloc[idx]
         batch_texts = row["label"]
-        batch_texts = torch.tensor(batch_texts, device=self.device)
+        batch_texts = torch.tensor(batch_texts)
```
scott-routledge2: Wouldn't this be a single item not a batch?

examples/ML/pandas_ddp_example.py
```diff
@@ -145,12 +82,18 @@ def validation(model, val_loader, loss_fn):
...
+                                    labels=val_label)
 
-            batch_loss = loss_fn(output, val_label.long())
+            batch_loss = output.loss
             
             # Track local loss and accuracy stats
-            acc = (output.argmax(dim=1) == val_label).sum().item()
+            acc = (output.logits.argmax(dim=1) == val_label).sum().item()
             total_ddp_acc_loss_train[0] += acc
             total_ddp_acc_loss_train[1] += batch_loss.item()
```
scott-routledge2: I think this is still the average loss (similar change needed in val loop):
```suggestion
            total_ddp_acc_loss_train[1] += batch_loss.item() * BATCH_SIZE
```

examples/ML/pandas_ddp_example.py
```diff
@@ -1,101 +1,38 @@
...
-from torch.utils.data import DataLoader
-from transformers import BertModel, BertTokenizer
+from torch.optim import AdamW
+from transformers import BertTokenizer, BertForSequenceClassification, get_cosine_schedule_with_warmup
 import os
 
-LR = 16e-6
-EPOCHS = 5
+LR = 2e-5
+EPOCHS = 2
```
scott-routledge2, line 12: Change back number of epochs? Or is 2 enough

bodo/ai/train.py
```diff
@@ -154,3 +173,70 @@ def prepare_model(
 
             model = FSDP(model, **parallel_strategy_kwargs)
     return model
+
+
+def prepare_dataset(
```
scott-routledge2, line 178: Should we allow users to set the seed here (if shuffle)?

examples/ML/pandas_ddp_example.py
scott-routledge2, line 1: Do we need to set the seed for the `rebalance` calls?

IsaacWarren: By default it's deterministic

examples/ML/pandas_ddp_example.py
IsaacWarren: 2's enough, it hits 95% accuracy

PR #875, opened 2025-10-10
First batch of narwhals support.

DrTodd13:

Add flag to turn off all fallback warnings.
Improved floor support.
Mod/% support.

## Testing strategy

run_ci

## User facing changes

Floor support on series (not just expressions).
Mod support on series and expressions.
Flag to turn off all fallback warnings.

## Checklist


bodo/pandas/physical/expression.cpp
```diff
@@ -736,3 +736,185 @@ void PhysicalExpression::join_expr_batch(
...
+        for (int64_t i = 0; i < left.length; ++i) {
+            if (!is_valid_bit(left_valid_bits, left.offset, i) ||
+                !is_valid_bit(right_valid_bits, right.offset, i)) {
+                clear_valid(out_valid_bits, offset + i);
+            } else {
+                // Get corresponding ith elements of left and right arrays.
+                CType l = left_values[i];
+                CType r = right_values[i];
+                // Calculate modulus operator.
+                CType res = ModOp::apply(l, r);
```
IsaacWarren, line 830: Can't we apply the compute function to the whole array at a time instead of elementwise?

bodo/pandas/series.py
```diff
@@ -1360,6 +1371,9 @@ def where(
         plan = _create_series_binop_plan(lhs_plan, empty_data, expr)
         return wrap_plan(plan=plan)
 
+    def to_list(self):
+        return super().to_list()
```
IsaacWarren: Do we know why this wasn't just falling back?

DrTodd13: I did this to get rid of the fallback warning to make narwhals test happy but then ended up adding the other option to turn them all off.  Probably will remove this.

bodo/pandas/physical/expression.cpp
DrTodd13: Which compute function do you think we should apply to the whole array?

IsaacWarren: ModOp::apply, I thought arrow compute functions could operate on arrays by default

DrTodd13: ModOp::apply isn't an arrow compute function.  It is defined in the same file with all the new stuff.  We're making a new custom arrow compute function so have to do the iteration ourselves.

scott-routledge2 (approved):
Thanks @DrTodd13 , LGTM

PR #876, opened 2025-10-11
Todd/rmod fix

DrTodd13:

1. Additional uses of df lib warn flag.
2. Support "not" as last projection expression applied.
3. Make arith_binop true if at most one of the dtypes is bool and the other is numeric.
4. Apply bool casting of true/false to 1/0 in arith_binop and new cumsum function.
5. Remove int64, float64, and bool from datetime types.  is_numeric seems like it shouldn't have been saying something wasn't numeric when it is if it could also be a datetime.  Seems like datetime should have nothing to do with numeric and we should have an is_datetime check first before is_numeric is there is some priority we need to apply.

## Testing strategy

run_ci

## User facing changes

none

## Checklist


bodo/pandas/frame.py
```diff
@@ -1414,7 +1417,8 @@ def apply(
                     "Original error: "
                     f"{error_msg}."
                 )
-                warnings.warn(BodoCompilationFailedWarning(msg))
+                if bodo.dataframe_library_warn:
+                    warnings.warn(BodoCompilationFailedWarning(msg))
```
IsaacWarren, line 1411: Maybe refactor these to use a df_lib_warn function that intenrally checks if dataframe_library warn is set?

bodo/pandas/series.py
```diff
@@ -1371,6 +1388,26 @@ def where(
         plan = _create_series_binop_plan(lhs_plan, empty_data, expr)
         return wrap_plan(plan=plan)
 
+    @check_args_fallback(supported="none")
+    def cumsum(self, axis: Axis | None = None, skipna: bool = True, *args, **kwargs):
```
IsaacWarren, line 1391: Will this return an int series for bool input? Does that match pandas?

DrTodd13:
```
>>> import pandas as pd
>>> a = pd.Series([True, False, True])
>>> b = a.cumsum()
>>> b
0    1
1    1
2    2
dtype: int64
```

bodo/pandas/frame.py
DrTodd13: Thanks. Done.  Looks cleaner now.

bodo/pandas/series.py
```diff
@@ -1371,6 +1388,26 @@ def where(
...
+        if self.dtype == pd.ArrowDtype(pa.bool_()):
+            self = self.map({True: 1, False: 0})
+        msg = (
+            "Series.cumsum is not implemented in Bodo DataFrames yet. "
+            "Falling back to Pandas (may be slow or run out of memory)."
+        )
+        if bodo.dataframe_library_warn:
+            warnings.warn(BodoLibFallbackWarning(msg))
+        with bodo.pandas.utils.FallbackContext():
+            py_res = super().cumsum(axis, skipna, *args, **kwargs)
```
scott-routledge2: Do we need to filter deeper fallback warnings here? Maybe FallbackContext should do that.

bodo/pandas/series.py
```diff
@@ -1371,6 +1388,26 @@ def where(
         plan = _create_series_binop_plan(lhs_plan, empty_data, expr)
         return wrap_plan(plan=plan)
 
+    @check_args_fallback(supported="none")
+    def cumsum(self, axis: Axis | None = None, skipna: bool = True, *args, **kwargs):
+        # cumsum not supported for pyarrow boolean so convert to int
+        # Fix in pyarrow instead?
+        if self.dtype == pd.ArrowDtype(pa.bool_()):
+            self = self.map({True: 1, False: 0})
```
scott-routledge2, line 1395: Should we do `na_action = 'ignore'` here? Other wise NA will be converted to 0.

bodo/pandas/series.py
```diff
@@ -332,7 +332,21 @@ def __invert__(self):
...
-        if is_numeric(other):
+        self_bool = is_bool(self)
+        other_bool = is_bool(other)
+        if (
+            (is_numeric(self) or self_bool)
+            and (is_numeric(other) or other_bool)
+            and not (self_bool and other_bool)
+        ):
+            if self_bool:
+                self = self.map({True: 1, False: 0})
```
scott-routledge2, line 344: Should this map preserve NAs i.e. na_action="ignore"? 
Edit: It *Shouldn't* need that, I think our map is buggy.

scott-routledge2 (approved):
Thanks @DrTodd13, LGTM overall. I think we should test with bool Series with missing values.

bodo/pandas/series.py
scott-routledge2: Oh I think this might be a bug with our map:

``` py
>>> import pandas as pd
>>> pd.Series([True, None, False] * 3).map({True: 1, False: 0}) + 9
0    10.0
1     NaN
2     9.0
3    10.0
4     NaN
5     9.0
6    10.0
7     NaN
8     9.0
dtype: float64
>>> import bodo.pandas as pd
>>> pd.Series([True, None, False] * 3).map({True: 1, False: 0}) + 9
0    10
1     9
2     9
3    10
4     9
5     9
6    10
7     9
8     9
dtype: int64[pyarrow]
```

DrTodd13: Our map seems to ignore na_action=ignore.

bodo/pandas/series.py
DrTodd13:
```
import pandas as pd
  import bodo.pandas as bd


  pa = pd.Series([True, None, False]) * 3
  print(pa)
  a = bd.Series([True, None, False]) * 3
  print(a)
```

DrTodd13:
```
0      3
1    NaN
2      0
dtype: object
0    3
1    0
2    0
dtype: int64[pyarrow]
```

DrTodd13: That is after I added na_action='ignore'.  It does the same without it.  So, I think you are right.

bodo/pandas/series.py
scott-routledge2: I'll open an issue to fix.

PR #902, opened 2025-10-29
Support join filters in BodoSQL C++ backend

ehsantn:

As title.

## Testing strategy

Existing tests (especially TPC-H Q5).

## User facing changes

None.

## Checklist

### DIFF
--- a/bodo/pandas/_plan.h
+++ b/bodo/pandas/_plan.h
@@ -33,15 +33,15 @@ class LogicalJoinFilter : public duckdb::LogicalOperator {
         duckdb::LogicalOperatorType::LOGICAL_EXTENSION_OPERATOR;
 
     LogicalJoinFilter(duckdb::unique_ptr<duckdb::LogicalOperator> source,
-                      std::vector<int> filter_ids,
-                      std::vector<std::vector<int64_t>> filter_columns,
-                      std::vector<std::vector<bool>> is_first_locations)
+                      const std::vector<int> filter_ids,
+                      const std::vector<std::vector<int64_t>> filter_columns,
+                      const std::vector<std::vector<bool>> is_first_locations)
         : duckdb::LogicalOperator(
-              duckdb::LogicalOperatorType::LOGICAL_EXTENSION_OPERATOR) {
+              duckdb::LogicalOperatorType::LOGICAL_EXTENSION_OPERATOR),
+          filter_ids(std::move(filter_ids)),
+          filter_columns(std::move(filter_columns)),
+          is_first_locations(std::move(is_first_locations)) {
         this->children.push_back(std::move(source));
-        this->filter_ids = filter_ids;
-        this->filter_columns = filter_columns;
-        this->is_first_locations = is_first_locations;
     }
 
     duckdb::vector<duckdb::ColumnBinding> GetColumnBindings() override {
@@ -49,11 +49,11 @@ class LogicalJoinFilter : public duckdb::LogicalOperator {
     }
 
     // IDs of joins creating each filter
-    std::vector<int> filter_ids;
+    const std::vector<int> filter_ids;
     // Mapping columns of the join to the columns in the current table
-    std::vector<std::vector<int64_t>> filter_columns;
+    const std::vector<std::vector<int64_t>> filter_columns;
     // Indicating for which of the columns is it the first filtering site
-    std::vector<std::vector<bool>> is_first_locations;
+    const std::vector<std::vector<bool>> is_first_locations;
 
    protected:
     void ResolveTypes() override { types = children[0]->types; }
--- a/bodo/pandas/physical/join.h
+++ b/bodo/pandas/physical/join.h
@@ -626,7 +626,7 @@ class PhysicalJoin : public PhysicalProcessBatch, public PhysicalSink {
     /**
      * @brief Get pointer to JoinState used in join filters
      */
-    JoinState* getJoinStatePtr() { return join_state_.get(); }
+    JoinState* getJoinStatePtr() const { return join_state_.get(); }
 
    private:
     /**
--- a/bodo/pandas/physical/join_filter.h
+++ b/bodo/pandas/physical/join_filter.h
@@ -24,11 +24,12 @@ class PhysicalJoinFilter : public PhysicalProcessBatch {
    public:
     explicit PhysicalJoinFilter(
         bodo::LogicalJoinFilter& logical_filter,
-        std::shared_ptr<bodo::Schema> input_schema,
-        std::shared_ptr<std::unordered_map<int, JoinState*>> join_filter_states)
-        : filter_ids(logical_filter.filter_ids),
-          filter_columns(logical_filter.filter_columns),
-          is_first_locations(logical_filter.is_first_locations),
+        std::shared_ptr<bodo::Schema>& input_schema,
+        std::shared_ptr<std::unordered_map<int, JoinState*>>&
+            join_filter_states)
+        : filter_ids(std::move(logical_filter.filter_ids)),
+          filter_columns(std::move(logical_filter.filter_columns)),
+          is_first_locations(std::move(logical_filter.is_first_locations)),
           join_filter_states(join_filter_states) {
         this->output_schema = std::make_shared<bodo::Schema>();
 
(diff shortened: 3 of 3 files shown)


BodoSQL/bodosql/context.py
```diff
@@ -1299,6 +1302,8 @@ def execute_cpp_backend(
                     f"C++ backend execution failed with error:\n{message}"
                 ) from e
             out = CPP_BACKEND_EXECUTION_FAILED
+        finally:
+            self.join_filter_info = None
```
IsaacWarren, line 1306: Why do we always need to reset this?

bodo/pandas/_plan.h
```diff
@@ -19,6 +19,48 @@
...
+   public:
+    // TODO: add join filter to DuckDB operator types to allow more extension
+    // types
+    static constexpr const duckdb::LogicalOperatorType TYPE =
+        duckdb::LogicalOperatorType::LOGICAL_EXTENSION_OPERATOR;
+
+    LogicalJoinFilter(duckdb::unique_ptr<duckdb::LogicalOperator> source,
+                      std::vector<int> filter_ids,
+                      std::vector<std::vector<int64_t>> filter_columns,
+                      std::vector<std::vector<bool>> is_first_locations)
```
IsaacWarren: Can any of these be const, along with the class member variables?

bodo/pandas/physical/join.h
```diff
@@ -623,6 +623,11 @@ class PhysicalJoin : public PhysicalProcessBatch, public PhysicalSink {
 
     int64_t getOpId() const { return PhysicalSink::getOpId(); }
 
+    /**
+     * @brief Get pointer to JoinState used in join filters
+     */
+    JoinState* getJoinStatePtr() { return join_state_.get(); }
```
IsaacWarren:
```suggestion
    JoinState* getJoinStatePtr() const { return join_state_.get(); }
```

bodo/pandas/physical/join_filter.h
```diff
@@ -0,0 +1,202 @@
...
+        this->output_schema = std::make_shared<bodo::Schema>();
+
+        this->materialize_after_each_filter = true;
+        for (size_t i = 0; i < input_schema->ncols(); i++) {
+            std::unique_ptr<bodo::DataType> col_type =
+                input_schema->column_types[i]->copy();
+            // TODO[BSE-5176]: Handle DICT column casting for bloom filters
+            // See
+            // https://github.com/bodo-ai/Bodo/blob/4b6e5830cc9f16bba5fc40ba495f11955bbf15af/bodo/libs/streaming/join.py#L1477
+            if (col_type->array_type == bodo_array_type::DICT) {
```
IsaacWarren, line 43: I think this is an important case because otherwise the whole query will fail right?

bodo/pandas/physical/join_filter.h
```diff
@@ -0,0 +1,202 @@
...
+/**
+ * @brief Physical node for join filter.
+ *
+ */
+class PhysicalJoinFilter : public PhysicalProcessBatch {
+   public:
+    explicit PhysicalJoinFilter(
+        bodo::LogicalJoinFilter& logical_filter,
+        std::shared_ptr<bodo::Schema> input_schema,
+        std::shared_ptr<std::unordered_map<int, JoinState*>> join_filter_states)
```
IsaacWarren:
```suggestion
        std::shared_ptr<bodo::Schema>& input_schema,
        std::shared_ptr<std::unordered_map<int, JoinState*>>& join_filter_states)
```

bodo/pandas/physical/join_filter.h
```diff
@@ -0,0 +1,202 @@
...
+            // TODO[BSE-5176]: Handle DICT column casting for bloom filters
+            // See
+            // https://github.com/bodo-ai/Bodo/blob/4b6e5830cc9f16bba5fc40ba495f11955bbf15af/bodo/libs/streaming/join.py#L1477
+            if (col_type->array_type == bodo_array_type::DICT) {
+                throw std::runtime_error(
+                    "Join filter does not support input tables with DICT "
+                    "columns yet");
+            }
+
+            // Avoid materializing after each filter if there is any variable
```
IsaacWarren, line 49: String columns are variable length too, right? Also I don't understand why we set the col type to string

bodo/pandas/physical/join_filter.h
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    /**
+     * @brief The logic for filtering input batches.
+     *
+     * @param input_batch - the input data to be filtered
+     * @return the filtered batch from applying the expression to the input.
+     */
+    std::pair<std::shared_ptr<table_info>, OperatorResult> ProcessBatch(
+        std::shared_ptr<table_info> input_batch,
```
IsaacWarren, line 97:
```suggestion
        std::shared_ptr<table_info>& input_batch,
```

bodo/pandas/physical/join_filter.h
```diff
@@ -0,0 +1,202 @@
...
+            }
+            HashJoinState* join_state = (HashJoinState*)join_state_;
+
+            applied_any_filter =
+                applied_any_filter ||
+                join_state->RuntimeFilter(input_batch, row_bitmask,
+                                          this->filter_columns[i],
+                                          this->is_first_locations[i]);
+
+            if (this->materialize_after_each_filter && applied_any_filter) {
```
IsaacWarren, line 139: Minor optimization would be to not reset the bitmask the last iteration

bodo/pandas/physical/join_filter.h
```diff
@@ -0,0 +1,202 @@
...
+     *
+     * @return std::shared_ptr<bodo::Schema> physical schema
+     */
+    const std::shared_ptr<bodo::Schema> getOutputSchema() override {
+        // Filter's output schema is the same as the input schema
+        return output_schema;
+    }
+
+   private:
+    std::shared_ptr<bodo::Schema> output_schema;
```
IsaacWarren, line 179: All of these can be const right?

bodo/pandas/vendor/duckdb/src/include/duckdb/planner/operator/logical_comparison_join.hpp
```diff
@@ -39,6 +39,9 @@ class LogicalComparisonJoin : public LogicalJoin {
 	//! Scans where we should push generated filters into (if any)
 	unique_ptr<JoinFilterPushdownInfo> filter_pushdown;
 
+	// Bodo change: keep track of join id for join filters (only used in BodoSQL C++ backend and doesn't go through DuckDB optimizer)
+	int join_id = -1;
```
IsaacWarren, line 43: Instead of adding a change can we just subclass this with the join_id field and use our version instead?

scott-routledge2 (approved):
LGTM, thanks @ehsantn !

bodo/pandas/physical/join_filter.h
ehsantn: I think we don't use dictionary-encoded columns in dataframe library in general yet. We have to fill all of these gaps when we do.

bodo/pandas/physical/join_filter.h
ehsantn: This is checking the column type to be string not setting it. The code formatting is a bit misleading here.

bodo/pandas/physical/join_filter.h
IsaacWarren: Oh, won't bodosql designate columns as dict encoded or does that happen at runtime?

bodo/pandas/physical/join_filter.h
ehsantn: We are updating this variable locally so shouldn't use a reference. Also, in general, using references for shared pointers most probably is faster since it's just a pointer and a counter and efficient already.

bodo/pandas/physical/join_filter.h
IsaacWarren: Oh yeah, going too fast 🤦

BodoSQL/bodosql/context.py
ehsantn: Just in case and for a bit more code clarity that the scope is limited to one query.

bodo/pandas/_plan.h
ehsantn: Done.

bodo/pandas/physical/join.h
ehsantn: Done.

bodo/pandas/physical/join_filter.h
ehsantn: Don't remember the details but the new C++ runtime doesn't do dict-encoded reads yet I think.

bodo/pandas/physical/join_filter.h
ehsantn: Done.

bodo/pandas/physical/join_filter.h
ehsantn: Done.

bodo/pandas/physical/join_filter.h
ehsantn: Done.

bodo/pandas/vendor/duckdb/src/include/duckdb/planner/operator/logical_comparison_join.hpp
ehsantn: I think subclassing is overkill in this case. We can keep it simple and see if we have more extensive changes later.

bodo/pandas/physical/join_filter.h
IsaacWarren: I believe the problem with shared pointers is creating a new one allocates new memory but if we need it in this case , makes sense.

PR #903, opened 2025-10-29
Finetune on Iceberg Data

IsaacWarren:

Adds a new example.

## Testing strategy
Tested on my laptop.

## User facing changes
New example

## Checklist


examples/ML/llama_finetune_iceberg_data.py
IsaacWarren, line 1: @strangeloopcanon @scott-routledge2 Can you guys take a look at this and see if it makes sense to you before I clean up/start on the blog

strangeloopcanon: Few thoughts:
- attention mask didn't move to device? It's a CPU going into GPU 
- padding should prob use the tokenizer's id, not 0, otherwise it'll corrup. I'd also add an EOS to output_tokenizerd
- maybe compute loss average with the true batch size per step
- maybe also do model.config.use_cache=False for training
- load the tokenizer once and just pass as argument in the if __name__ == "__main__" block
- It iterates idxs once to find max_len, then iterate again to build tensors. For large batches I'd compute max_len in a single pass
-  save is with torch.distributed.checkpoint.save but then immediately call base_model.save_pretrained with rank 0, prob worth just using torch evwhere?
- oh and not sure this was intentional or not but the lora target now hits all linear projections, just note its quite aggressive sinceits training most of the model. maybe just do attention projections?
- i'd gradient clip! before optimizer step
- Minor but I'd also split checkpoint dirs for adapter and checkpoint

examples/ML/llama_finetune_iceberg_data.py
```diff
@@ -0,0 +1,246 @@
...
+        input_ids = torch.tensor(row["input_tokenized"])
+        output_ids = torch.tensor(row["output_tokenized"])
+        example = torch.cat([input_ids, output_ids])
+        labels = torch.cat([torch.full_like(input_ids, -100), output_ids])
+        attention_mask = torch.ones_like(example)
+
+        return example, labels, attention_mask
+
+    def __getitems__(self, idxs):
+        input_ids = []
```
scott-routledge2: Could we just do `tokenizer(list_batch_input_output, padding="max_len")` here/add a collate_fn? I guess the tradeoff would be Tokenizing an entire dataset up front might not be practical for large data but could potentially save some compute if the entire result can fit in memory.

examples/ML/llama_finetune_iceberg_data.py
```diff
@@ -0,0 +1,246 @@
...
+        
+        torch.distributed.checkpoint.save(
+            {"model_state_dict": adapter_state_dict}, # Save only the adapter
+            checkpoint_id=CHECKPOINT_DIR
+        )
+        
+    # Save in peft-preferred format on rank 0
+    if pytorch_rank == 0:
+       base_model.save_pretrained(CHECKPOINT_DIR) # Saves adapter_config.json etc.
+
```
scott-routledge2, line 243: Do some evaluation/inference at the end? I'm not sure if a validation set makes sense here but it would be good to confirm the model can at least perform pretty well on the training data (memorize some examples)

IsaacWarren: The problem is the training data is just random text so it won't actually improve

examples/ML/llama_finetune_iceberg_data.py
IsaacWarren: Yeah I was trying to use our tokenize function to show it off but it might not be the best fit

examples/ML/llama_finetune_iceberg_data.py
scott-routledge2: Oh makes sense. I think it should still be able memorize it's training data if it's small enough? more of a test to check the config is correct but not necessary

scott-routledge2: Also did we want to add the data gen script to the example?

IsaacWarren: The data is public, do you think we still should?

examples/ML/llama_finetune_iceberg_data.py
scott-routledge2: Maybe we could add some version of this to `datasets.py` called something like `TokensDataset`

examples/ML/llama_finetune_iceberg_data.py
IsaacWarren: Thanks! I've applied most of these, just a couple questions on a few

> It iterates idxs once to find max_len, then iterate again to build tensors. For large batches I'd compute max_len in a single pass

Not sure what this means, you have to iterate twice to find the max first if you need the max in the loop

> save is with torch.distributed.checkpoint.save but then immediately call base_model.save_pretrained with rank 0, prob worth just using torch evwhere?

It's two different formats, one is the torch one every epoch, the other is peft's format so it can easily be applied as a lora

> h and not sure this was intentional or not but the lora target now hits all linear projections, just note its quite aggressive sinceits training most of the model. maybe just do attention projections?

It's training about .5% of the parameters right now but that could make it even faster, I'll give it a shot!

> Minor but I'd also split checkpoint dirs for adapter and checkpoint

It only saves the adapter, just in different formats. I just wasn't sure how distributed model would interact with PEFT's saving so I only did it when training was done.

examples/ML/llama_finetune_iceberg_data.py
scott-routledge2: Oh didn't realize it was a public. It's probably not necessary then, unless maybe we want to show writing an iceberg table with Bodo for the blog?

IsaacWarren: > Oh didn't realize it was a public. It's probably not necessary then, unless maybe we want to show writing an iceberg table with Bodo for the blog? 

I did it with pyicberg+pyarrow 😂

examples/ML/llama_finetune_iceberg_data.py
strangeloopcanon: Ah misunderstood the intent re 2 saves, maybe a comment in the code to explain.

Re idx point it need two phases (find max, then pad), but you don't need two passes over the dataframe was my thought. In __getitems__, you iterate idxs twice and call df.iloc[idx] + tensor construction twice for the same indices. Prob academic for batch size 2 though, will only matter if it's >> higher.

examples/ML/llama_finetune_iceberg_data.py
IsaacWarren: Maybe, I just don't see a way to make it generic enough?

examples/ML/llama_finetune_iceberg_data.py
IsaacWarren: Gotcha, I reworked it to use the tokenizer so shouldn't be a problem anymore

examples/ML/llama_finetune_iceberg_data.py
```diff
@@ -0,0 +1,252 @@
+import bodo.pandas as pd
+
+# Use pandas for timestamp manipulations
+import pandas
```
ehsantn: Seems unnecessary now? Is there a gap for us to fill?
```suggestion
```

IsaacWarren: Good catch, I was using a timedelta to get an offset from the current date but realized that wouldn't work since the dates in the table are static so just used a set cutoff date instead

scott-routledge2: Just tried `pd.Timestamp.now()` and it doesn't work, since we wrap everything in `fallback_wrapper` by default I think.

ehsantn: Let's open issues for these and fix them soon since seem important.

IsaacWarren: https://bodo.atlassian.net/browse/BSE-5179?atlOrigin=eyJpIjoiMzBmMDlkMjViMDc2NDMxOWFmNDJkMDc1MGQ2NTljMDQiLCJwIjoiaiJ9

PR #905, opened 2025-10-29
Upgrade to arrow 22

scott-routledge2:

* Updates python lower bound to 3.10 (to actually drop 3.9, we would have to update meta.yaml in conda-forge and the PYTHON_VERSIONS variable in github settings, which will be a followup)
* Update iceberg parquet reader to always use st_executor and remove USE_BODO_ARROW_FORK variable from scripts
* Adds `set_time_adjusted_to_utc(true)` option to pq writer so we can write parquet data with TIME columns to snowflake (Arrow 22 changed the default to false). 

## Checklist


bodo/io/iceberg_parquet_reader.cpp
```diff
@@ -408,10 +408,7 @@ static std::shared_ptr<::arrow::dataset::ScanOptions> make_scan_options(
         builder->GetScanOptions(), "Error during GetScanOptions",
         std::shared_ptr<::arrow::dataset::ScanOptions> scan_options);
 
-#ifdef USE_BODO_ARROW_FORK
-    scan_options->exec_context = arrow::compute::ExecContext(
-        bodo::BufferPool::DefaultPtr(), cpu_executor);
-#endif
+    scan_options->cpu_executor = cpu_executor;
```
scott-routledge2, line 411: I had to change the upstream PR so that ScanOptions now has an Executor (as opposed to an ExecContext) since the other parts of the ExecContext, including the memory pool were not used anywhere in our fork/upstream PR.

ehsantn (approved):
Thanks @scott-routledge2 .

.github/actions/build-source/action.yml
```diff
@@ -30,10 +30,6 @@ runs:
 
       - name: Build Bodo
         run: |
-          if [[ "$RUNNER_OS" == 'Linux' && "$RUNNER_ARCH" == 'X64' ]]; then
-            # Build Bodo with our fork of Arrow
-            export USE_BODO_ARROW_FORK=1
```
IsaacWarren, line 35: What about the other change that hasn't been merged? Is it completely internal to our former arrow lib?

IsaacWarren (approved):
Thanks Scott

.github/actions/build-source/action.yml
scott-routledge2: Yeah it's an internal change.

PR #915, opened 2025-11-06
Timezone Support

scott-routledge2:

Keeps track of the Timezone in the C++ backend. The actual values themselves are still timezone naive but we keep track of the timezone to pass to arrow compute functions/run_func_on_table for datetime manipulation. Also updates `array_to_info` to pass timezone information for `DatetimeArray`.

## Testing strategy

New end to end tests added. CI, Narwhals tests.

## User facing changes

Users can now use the dt accessor with timezone types.

## Checklist

### DIFF
--- a/bodo/tests/test_df_lib/test_end_to_end.py
+++ b/bodo/tests/test_df_lib/test_end_to_end.py
@@ -3927,3 +3927,33 @@ def test_timezone_groupby(timezone_timestamp_df):
     pandas_out = df.groupby("A", as_index=False).min().B.dt.hour
 
     _test_equal(bodo_out, pandas_out, check_pandas_types=False)
+
+
+def test_timezone_merge(timezone_timestamp_df):
+    """Test merge works with timezone keys"""
+    df = timezone_timestamp_df
+    df2 = pd.DataFrame(
+        {
+            "C": pd.date_range(
+                "2023-01-01 00:00:00",
+                periods=5,
+                freq="h",
+                tz="America/Chicago",
+            )
+        }
+    )
+
+    bdf = bd.from_pandas(df)
+    bdf2 = bd.from_pandas(df2)
+
+    with assert_executed_plan_count(0):
+        bodo_out = bdf.merge(bdf2, left_on="A", right_on="C", how="inner")
+    pandas_out = df.merge(df2, left_on="A", right_on="C", how="inner")
+
+    _test_equal(
+        bodo_out,
+        pandas_out,
+        check_pandas_types=False,
+        sort_output=True,
+        reset_index=True,
+    )
--- a/bodo/utils/utils.py
+++ b/bodo/utils/utils.py
@@ -362,7 +362,7 @@ def numba_to_c_array_types(
         elif isinstance(arr_type, bodo.types.DatetimeArrayType):
             c_arr_types.append(numba_to_c_array_type(arr_type))
             # TODO: Serialize Timezone information here. See:
-            # https://github.com/bodo-ai/Bodo/blob/15f781f40664fba7fc5443917171c9a5cdfc1e18/bodo/libs/_bodo_common.cpp#L564
+            # https://github.com/bodo-ai/Bodo/blob/9e198ffd8fb1a554d3bdf324a01264ae0af9343f/bodo/libs/_bodo_common.cpp#L554
             c_arr_types.append(0)
         else:
             c_arr_types.append(numba_to_c_array_type(arr_type))


bodo/pandas/_plan.cpp
```diff
@@ -1642,9 +1642,7 @@ std::pair<duckdb::string, duckdb::LogicalType> arrow_field_to_duckdb(
                         break;
                 }
             } else {
-                // TODO: Do we need to check units here?
-                // Technically this is supposed to be in microseconds like
-                // TIMESTAMP
```
scott-routledge2, line 1647: Duckdb drops nanoseconds when querying Pandas DataFrames with timezone, so I think it is ok for us to do that too.

bodo/libs/groupby/_groupby_col_set.cpp
```diff
@@ -49,9 +49,11 @@ void BasicColSet::alloc_running_value_columns(
     int64_t num_categories = in_col->num_categories;
     std::tie(arr_type, dtype) =
         get_groupby_output_dtype(ftype, arr_type, dtype);
-    out_cols.push_back(alloc_array_top_level(
-        num_groups, 1, 1, arr_type, dtype, -1, 0, num_categories, false, false,
-        false, pool, std::move(mm)));
+    auto out_arr = alloc_array_top_level(num_groups, 1, 1, arr_type, dtype, -1,
+                                         0, num_categories, false, false, false,
+                                         pool, std::move(mm));
+    out_arr->timezone = in_col->timezone;
```
scott-routledge2: This function is how we alloc output columns for the ACC case and the other function (getRunningValueColumnTypes) is how we get column types for the AGG case.

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -3847,3 +3847,67 @@ def test_asarray_df():
...
+    _test_equal(bodo_out, pandas_out, check_pandas_types=False, reset_index=reset_index)
+
+
+def test_timezone_groupby(timezone_timestamp_df):
+    """Test groupby works with timezones"""
+    df = timezone_timestamp_df
+    bdf = bd.from_pandas(df)
+
+    with assert_executed_plan_count(0):
+        # TODO: fix as_index case
```
scott-routledge2: Unrelated to timezone but there is bug I found with adding a projection on top of a groupby (the value column gets duplicated in the index):

e.g. 
``` py
bdf = pd.DataFrame({
        "timestamp": timestamps,
        "value": range(5)
    })
bdf.groupby("value")["timestamp"].min().map(lambda x: x, engine="python")
```

```
value
1735707600000000000    2025-01-01 00:00:00-05:00
1735711200000000000    2025-01-01 01:00:00-05:00
1735714800000000000    2025-01-01 02:00:00-05:00
1735718400000000000    2025-01-01 03:00:00-05:00
1735722000000000000    2025-01-01 04:00:00-05:00
```

bodo/libs/groupby/_groupby_col_set.cpp
ehsantn: Maybe add `timezone` to `alloc_array_top_level` arguments to be cleaner?

bodo/tests/test_df_lib/test_end_to_end.py
ehsantn: Let's open an issue.

bodo/libs/_bodo_to_arrow.cpp
```diff
@@ -1454,8 +1456,13 @@ std::shared_ptr<array_info> arrow_array_to_bodo(
...
                     std::static_pointer_cast<arrow::TimestampArray>(casted_arr);
             }
-            return arrow_numeric_array_to_bodo<arrow::TimestampArray>(
-                ts_arr, Bodo_CTypes::DATETIME, src_pool);
+            std::shared_ptr<array_info> bodo_arr =
+                arrow_numeric_array_to_bodo<arrow::TimestampArray>(
+                    ts_arr, Bodo_CTypes::DATETIME, src_pool);
+
+            // Store original timezone info (for DataFrame Library)
+            bodo_arr->timezone = type->timezone();
```
ehsantn: Move to `arrow_numeric_array_to_bodo`?

bodo/libs/_array.cpp
```diff
@@ -224,6 +224,26 @@ array_info* interval_array_to_info(uint64_t n_items, char* left_data,
                           {left_buff, right_buff});
 }
 
+array_info* datetime_array_to_info(uint64_t n_items, char* data, int typ_enum,
```
ehsantn, line 227: What about the info to array path?

ehsantn (approved):
Thanks @scott-routledge2. See comments.

bodo/libs/_array.cpp
scott-routledge2: For info to Array we require specifying the output Array type, which should have the Timezone information.

ehsantn: Ok.

bodo/libs/_bodo_common.h
```diff
@@ -585,21 +586,25 @@ struct DataType {
     const Bodo_CTypes::CTypeEnum c_type;
     const int8_t precision;
     const int8_t scale;
+    const std::string
+        timezone;  // for DATETIME types, doesn't Serialize/DeSerialize
```
IsaacWarren: Don't we need it to serialize and deserialilze?

bodo/libs/_bodo_to_arrow.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include <cassert>
 #include <cstring>
+#include <iostream>
```
IsaacWarren:
```suggestion
```

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -3863,3 +3863,67 @@ def test_asarray_df():
...
+    bdf = bd.from_pandas(df)
+
+    val = pd.Timestamp("2023-01-01 00:02:00", tz="America/Chicago")
+    with assert_executed_plan_count(0):
+        bodo_out = bdf[bdf.A > val]
+    pandas_out = df[df.A > val]
+
+    reset_index = isinstance(index_val, pd.RangeIndex)
+    _test_equal(bodo_out, pandas_out, check_pandas_types=False, reset_index=reset_index)
+
```
IsaacWarren, line 3917: Maybe test them as join keys?

bodo/libs/_bodo_common.h
scott-routledge2: Serialize/Deserialize is for passing types between Python-C++ StreamingStates in BodoSQL, so I don't think it's needed at this time.

IsaacWarren: IIRC it's also used to send schemas during shuffle, are precision/scale serialized, I think we should match what we do got those

scott-routledge2: Oh I didn't see that when I searched for "Serialize", do you have a link to the code that does this? We do serialize/de-serialize decimal scale/precision so I'm happy to add it.

IsaacWarren: https://github.com/bodo-ai/Bodo/blob/isaac/subtree_duckdb/bodo/libs/streaming/_shuffle.cpp#L88-L88, not how I remember it being used but that's what I found

IsaacWarren: Plus we might want to pass tz from bodosql to the runtime in the future

PR #923, opened 2025-11-11
Run Narwhals tests on a single worker and fix issues

scott-routledge2:

Some tests like those in `unique_test.py` have non-deterministic results that are both correct, to avoid too many testing fixes, we can mark these types of tests as "skip if multiple workers". We can also catch different issues running tests on one worker vs multiple. Narwhals tests are pretty quick to run (~3 minutes to run both single and multiple works)

* Fixes slicing issue with negative step that returns an empty slice if stop is unbounded and start falls within len(md_head)

## Testing strategy

Extended `test_slice` to add a case for negative step

## User facing changes

Bug with slicing a Series with negative step fixed

## Checklist

### DIFF
--- a/.github/workflows/_test_python_source.yml
+++ b/.github/workflows/_test_python_source.yml
@@ -36,7 +36,7 @@ on:
         type: string
         description: Number of workers to use for Bodo tests.
         required: false
-        default: ''
+        default: "2"
 
 jobs:
   run:
@@ -102,17 +102,8 @@ jobs:
               git clone https://github.com/bodo-ai/narwhals.git
               cd narwhals
               python -c "import bodo; print('Detected Bodo version:', bodo.__version__)"
-
-              # Run tests on one worker
-              BODO_NUM_WORKERS=1 pytest -s -v -Wignore -k bodo
-              one_rank_exit_code=$?
-
-              # Run tests on multiple workers
-              python -c "import bodo; print('num workers:', bodo.spawn.spawner.get_num_workers())"
               pytest -s -v -Wignore -k bodo
-              multi_rank_exit_code=$?
-
-              exit $((one_rank_exit_code + multi_rank_exit_code))
+              exit $?
             else
               # Bodo Tests: Coverage is collected
               pytest -s -v -Wignore \


.github/workflows/_test_python_source.yml
```diff
@@ -97,8 +97,17 @@ jobs:
               git clone https://github.com/bodo-ai/narwhals.git
               cd narwhals
               python -c "import bodo; print('Detected Bodo version:', bodo.__version__)"
+
+              # Run tests on one worker
+              BODO_NUM_WORKERS=1 pytest -s -v -Wignore -k bodo
```
DrTodd13: Would it be useful to see at a glance if the single-worker or multi-worker case is failing?  I'm not sure what is best but could make this a different top-level test category and then control BODO_NUM_WORKERS like the WARN env var below.

bodo/pandas/utils.py
```diff
@@ -64,6 +64,25 @@ def get_lazy_single_manager_class() -> type[
...
+    Args:
+        slobj (slice): The slice object
+        nrows (int): Total number of rows in the DataFrame/Series
+
+    Returns:
+        tuple[int, int | None, int]: A tuple of normalized (start, stop, step)
+            without negative start/stop indices. stop is None implies the slice
+            goes from start to the beginning in reverse order.
+    """
+    start, stop, step = slobj.indices(nrows)
```
DrTodd13, line 81: Will .loc also go through this routine?  If so, then is stop = None correct?  Should start here also get a similar negative check that stop has?

scott-routledge2: Since the stop is exclusive, there is no way to represent the beginning of a list in reverse order since 0 would exclude the first element and -1 would imply the last element, which is why I am using None here. Since the start is inclusive then I don't think there needs to be a similar negative check.

From testing I think loc takes this same code path but it seems to work. Any particular reason it wouldn't work here?

scott-routledge2: In the tests, we cover missing/None indices, positive indices, and negative indices, the part that was missing was negative step, which I add in this PR, so we should have all the cases covered.

PR #934, opened 2025-11-17
Numba 0.62.1 upgrade

scott-routledge2:

Numba upgrade and changes to testing infrastructure to account for higher memory footprint of running spawn/dataframes tests after upgrading. 

## Testing strategy

Nightly, everything appears to be passing except [Ollama test, which I think is flakey](https://dev.azure.com/bodo-inc/Bodo/_build/results?buildId=26120&view=logs&j=5e0e995c-bc8d-58fb-ee95-9cdce8fe5fa7&t=9d3ce10f-bc91-5a89-8dcc-53df89aad876&l=837)

## Checklist

### DIFF
--- a/bodo/tests/test_df_lib/test_groupby_udf.py
+++ b/bodo/tests/test_df_lib/test_groupby_udf.py
@@ -191,7 +191,9 @@ def test_agg_mix_udf_builtin(groupby_df):
         pytest.param(
             lambda df, func: df.groupby(by=["A"]).apply(func, include_groups=False),
             id="apply",
-            marks=pytest.mark.weekly,
+            marks=pytest.mark.skip(
+                "Skip tests on Nightly CI to reduce memory footprint."
+            ),
         ),
     ],
 )


azure-release-pipelines.yml
```diff
@@ -100,11 +103,11 @@ stages:
         ${{ each slice in split(variables.bodo_df_1p_each, ',') }}:
           ${{ format('BODO_1P_{0}_OF_{1}', slice, variables.bodo_df_1p_total) }}:
             NP: 1
-            PYTEST_MARKER: "df_lib and not weekly and ${{ format('bodo_df_{0}of{1} and ({2})', slice, variables.bodo_df_1p_total, parameters.PYTEST_MARKER) }}"
```
IsaacWarren, line 103: Why get rid of not weekly?

bodo/tests/test_df_lib/test_groupby_udf.py
```diff
@@ -191,6 +191,7 @@ def udf(x):
         pytest.param(
             lambda df, func: df.groupby(by=["A"]).apply(func, include_groups=False),
             id="apply",
+            marks=pytest.mark.weekly,
```
IsaacWarren:
```suggestion
            marks=pytest.mark.weekly,
```
Are the weekly tests actually run weekly?

azure-release-pipelines.yml
scott-routledge2: Not weekly is applied to all tests by default.

bodo/tests/test_df_lib/test_groupby_udf.py
scott-routledge2: Good point, I actually don't think we have weekly tests. I was thinking of just removing this param since the code paths are essentially the same and it adds a lot of memory to our tests (we have other specific tests for apply also), but adding the weekly marker lets us run them still just in case. WDYT?

IsaacWarren: Mark it slow so it runs on nightly?

scott-routledge2: Running them on nightly was what I was trying to avoid since they were using a lot of memory.

IsaacWarren: Ok, if we don't want them on CI at all, just skip them with a message saying not running on CI because of memory. That's the cleanest IMO

PR #935, opened 2025-11-18
Add metrics for Iceberg Read/Parquet Read

scott-routledge2:

Add Iceberg read/init stage metrics.

## Testing strategy

PR CI, local testing.

## User facing changes

No

## Checklist

### DIFF
--- a/bodo/pandas/physical/read_iceberg.cpp
+++ b/bodo/pandas/physical/read_iceberg.cpp
@@ -1,6 +1,5 @@
 #include "read_iceberg.h"
 #include <arrow/util/key_value_metadata.h>
-#include <iostream>
 #include "../../libs/_utils.h"
 #include "../_util.h"
 #include "physical/operator.h"


bodo/pandas/_executor.h
```diff
@@ -69,6 +69,7 @@ class Executor {
    public:
     explicit Executor(std::unique_ptr<duckdb::LogicalOperator> plan,
                       std::shared_ptr<arrow::Schema> out_schema) {
+        QueryProfileCollector::Default().Init();
```
scott-routledge2, line 72: This change allows us to collect metrics from the init stages of Parquet/Iceberg read.

bodo/io/iceberg_parquet_reader.h
```diff
@@ -130,6 +130,14 @@ class IcebergParquetReader : public ArrowReader {
 
     IcebergParquetReaderMetrics iceberg_reader_metrics;
 
+    /**
+     * @brief Report Read Stage metrics if they haven't already been reported.
+     * Note that this will only report the metrics to the QueryProfileCollector
+     * the first time it's called.
+     *
+     */
+    void ReportReadStageMetrics(std::vector<MetricBase>& metrics_out) override;
```
scott-routledge2, line 139: Instead of having to call this method directly from each PhysicalOperator that has an arrow reader, maybe `ArrowReader::read_batch` could call this when is_last=true?

DrTodd13: Prefer current approach instead of integrating when is_last=true.

IsaacWarren (approved):
Thanks @scott-routledge2

PR #941, opened 2025-11-24
Support np.ufunc calls on BodoSeries

ehsantn:

As title.

## Testing strategy

Added unit test.

## User facing changes

Numpy ufuncs won't fall back.

## Checklist

### DIFF
--- a/bodo/pandas/series.py
+++ b/bodo/pandas/series.py
@@ -507,7 +507,9 @@ class BodoSeries(pd.Series, BodoLazyWrapper):
         from bodo.pandas.base import _empty_like
 
         if method != "__call__" or len(inputs) != 1 or inputs[0] is not self or kwargs:
-            raise NotImplementedError("ufunc not implemented for BodoSeries yet")
+            raise BodoLibNotImplementedException(
+                "ufunc not implemented for BodoSeries yet"
+            )
 
         new_metadata = _get_empty_series_arrow(ufunc(_empty_like(self)))
 


bodo/pandas/series.py
```diff
@@ -498,6 +499,27 @@ def __getitem__(self, key):
...
 
+    @check_args_fallback(unsupported="none")
+    def __array_ufunc__(
+        self, ufunc: np.ufunc, method: str, *inputs: pt.Any, **kwargs: pt.Any
+    ):
+        """Adds support for simple numpy ufuncs on BodoSeries like np.func(Series)."""
+        from bodo.pandas.base import _empty_like
+
+        if method != "__call__" or len(inputs) != 1 or inputs[0] is not self or kwargs:
+            raise NotImplementedError("ufunc not implemented for BodoSeries yet")
```
scott-routledge2: Can we make this `BodoLibNotImplementedException` to fallback for unsupported arguments?

ehsantn: Good catch. AI-generated error.

DrTodd13: Oh wait.  Isn't it part of __array_ufunc__ API that it return NotImplemented if it can't handle something?  So, I would expect it to return and not raise.  Definitely need to test the failure case on this.  @ehsantn

ehsantn: This falls back to the Pandas `__array_ufunc__` which in turn may return `NotImplemented` if necessary. I just tested the fallback with `np.divide` and it works.

PR #942, opened 2025-11-24
Todd/prune fix

DrTodd13:

1. Fix nested join callback so that the bitmap is at the block level rather than using global indices.
2. Save new BoundColumnRefExpressions for CTE nodes in remove_unused_columns pass so that references to them stay around while they are being used.

## Testing strategy

run_ci

## User facing changes

None.

## Checklist


CMakeLists.txt
```diff
@@ -229,7 +229,7 @@ if (CMAKE_BUILD_TYPE STREQUAL "Debug" OR CMAKE_BUILD_TYPE STREQUAL "DebugSanitiz
   if (NOT WIN32)
     add_compile_options(
       "-g"
-      "-O1"
+      "-O0"
```
DrTodd13, line 232: Fix to let me use this mode on Linux.

bodo/pandas/physical/expression.cpp
```diff
@@ -857,7 +857,8 @@ void PhysicalExpression::join_expr_batch(
     int64_t right_index_start, int64_t right_index_end) {
     for (int64_t j = right_index_start; j < right_index_end; j++) {
         for (int64_t i = left_index_start; i < left_index_end; i++) {
-            SetBitTo(match_arr, i + j,
+            SetBitTo(match_arr,
+                     (i - left_index_start) + (j - right_index_start),
```
DrTodd13, line 861: match_arr only stores matches for one combination of the blocks of the probe and build side.  So, need to subtract the start indices of both sides from the global indices to get the per-block indices.

ehsantn (approved):
Thanks @DrTodd13 .

PR #948, opened 2025-12-01
Join filter pass

IsaacWarren:

Adds a join filter pass that runs after optimization. Currently supports pushing through Joins and Projections. No support for pushing into IO currently.
## Testing strategy
PR CI, mostly used TPCH locally for debugging.

## User facing changes
Performance when join filters can push past expensive projections or in front of other joins

## Checklist


bodo/pandas/optimizer/runtime_join_filter.cpp
```diff
@@ -0,0 +1,284 @@
...
+#include "duckdb/planner/logical_operator.hpp"
+#include "duckdb/planner/operator/logical_comparison_join.hpp"
+#include "duckdb/planner/operator/logical_projection.hpp"
+
+duckdb::unique_ptr<duckdb::LogicalOperator>
+RuntimeJoinFilterPushdownOptimizer::VisitOperator(
+    duckdb::unique_ptr<duckdb::LogicalOperator> &op) {
+    switch (op->type) {
+        case duckdb::LogicalOperatorType::LOGICAL_COMPARISON_JOIN: {
+            return this->VisitCompJoin(op);
```
DrTodd13, line 18: I slightly prefer the cast here and the function then taking the known subtypes.

bodo/pandas/optimizer/runtime_join_filter.cpp
```diff
@@ -0,0 +1,284 @@
...
+        } break;
+        case duckdb::LogicalOperatorType::LOGICAL_PROJECTION: {
+            op = this->VisitProjection(op);
```
DrTodd13, line 24: Same as above.

bodo/pandas/_plan.h
```diff
@@ -48,6 +47,28 @@ class LogicalJoinFilter : public duckdb::LogicalOperator {
         return children[0]->GetColumnBindings();
     }
 
+    duckdb::string GetName() const override { return "LogicalJoinFilter"; }
+    duckdb::InsertionOrderPreservingMap<duckdb::string> ParamsToString()
```
DrTodd13: Unused?

bodo/pandas/_executor.h
```diff
@@ -77,17 +78,32 @@ class Executor {
         // Otherwise, mark it as seen
         seen.insert(cur);
 
+        std::vector<std::shared_ptr<Pipeline>> inserted_dependents;
         for (auto it = cur->run_before_begin(); it != cur->run_before_end();
              ++it) {
-            fillPipelinesTopoSort(*it, seen);
+            if (fillPipelinesTopoSort(*it, seen)) {
+                // Dependent was added so remove it from run_before
+                inserted_dependents.emplace_back(*it);
```
DrTodd13, line 92: I'm a little confused by this algorithm.  My instinct here is to use a visited set (like the "seen" set but not reset at the end of the func like seen is) and first do a check of this set if the node has previously been added then just return.  If two nodes A and B reference some common node C and A must go before B then you'll get to C through A first and the pipeline order will be fine.  If B goes before A then you get to C through B first and again it is fine.  That algorithm seems idiomatic in these situations but I'm not sure how yours is working because node C doesn't know about A and B.  Anyway, for this particular case and the overall join filter creation and pushdown, I'd like there to be some higher-level algorithmic descriptions in appropriate places.

bodo/pandas/optimizer/runtime_join_filter.cpp
```diff
@@ -0,0 +1,284 @@
...
+        }
+    }
+    for (auto &i : op->children) {
+        i = VisitOperator(i);
+    }
+    return std::move(op);
+}
+
+duckdb::unique_ptr<duckdb::LogicalOperator>
+RuntimeJoinFilterPushdownOptimizer::insert_join_filters(
```
DrTodd13, line 52: Docs.

bodo/pandas/optimizer/runtime_join_filter.cpp
```diff
@@ -0,0 +1,284 @@
...
+        make_join_filter(op,
+                         /*filter_ids=*/filter_ids,
+                         /*filter_columns=*/filter_columns,
+                         /*is_first_locations=*/is_first_locations);
+    // Replace current operator with join filter
+    return std::move(join_filter);
+}
+
+duckdb::unique_ptr<duckdb::LogicalOperator>
+RuntimeJoinFilterPushdownOptimizer::VisitCompJoin(
```
DrTodd13, line 76: Docs.

bodo/pandas/optimizer/runtime_join_filter.cpp
```diff
@@ -0,0 +1,284 @@
...
+    // Visit probe side
+    op->children[0] = VisitOperator(op->children[0]);
+    this->join_state_map = right_join_state_map;
+    // Visit build side
+    op->children[1] = VisitOperator(op->children[1]);
+    return this->insert_join_filters(op, out_join_state_map);
+}
+
+duckdb::unique_ptr<duckdb::LogicalOperator>
+RuntimeJoinFilterPushdownOptimizer::VisitProjection(
```
DrTodd13, line 274: Docs.

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -1354,7 +1355,8 @@ def test_merge_non_equi_cond():
     pre, post = bd.plan.getPlanStatistics(nan_bdf4._mgr._plan)
 
     _test_equal(pre, 5)
-    _test_equal(post, 4)
+    # The filter node gets pushed into join and then a join filter is inserted
+    _test_equal(post, 5)
```
DrTodd13, line 1359: Is comment on 1354 still valid?

bodo/pandas/optimizer/runtime_join_filter.cpp
```diff
@@ -0,0 +1,284 @@
...
+                    new_filter_columns.end()) {
+                    new_filter_columns.push_back(child_col);
+                    new_is_first_locations.push_back(true);
+                } else {
+                    // Duplicate column reference in projection, cannot push
+                    new_filter_columns.push_back(-1);
+                    new_is_first_locations.push_back(false);
+                }
+            } else {
+                // Projection expression is not a column ref, cannot push down
```
DrTodd13, line 307: Do we know if this condition is triggered in TPCH?  This isn't strictly an impossibility right?  You could apply an inverse operation of the expression and push that down?  Perhaps mark as a TO-DO if that is indeed the case and it may be important.

bodo/pandas/optimizer/runtime_join_filter.cpp
IsaacWarren: I agree that's cleaner but it leads to difficulty returning itself, the cast produces a value instead of a unique ptr so it can't be returned. We can't return a bare value because I think that would require a copy. I tried doing a pointer cast but they all consume the unique pointer

bodo/tests/test_df_lib/test_end_to_end.py
IsaacWarren: Yep, it still gets pushed into the join since join filter creation/pushdown runs last

bodo/pandas/optimizer/runtime_join_filter.cpp
IsaacWarren: Will add a todo, I didn't see us doing this in BodoSQL so I didn't think it was initially important. It's tough because you have to weigh the expense of evaluating the expression an extra time vs the benefit of rows filtered

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -1329,7 +1329,8 @@ def test_merge_non_equi_cond():
     pre, post = bd.plan.getPlanStatistics(bdf4._mgr._plan)
 
     _test_equal(pre, 5)
-    _test_equal(post, 4)
+    # The filter node gets pushed into join and then a join filter is inserted
+    _test_equal(post, 5)
```
ehsantn, line 1333: Is there a better way to test join filters? Since the number of nodes are the same here before and after, this test passes if the optimizer is not working. Could be a follow up PR.

bodo/pandas/_executor.h
```diff
@@ -61,11 +61,18 @@ class Executor {
     std::vector<std::shared_ptr<Pipeline>> pipelines;
     /*
      * @brief Do topological sort and fill in the pipelines vector.
+     * https://en.wikipedia.org/wiki/Topological_sorting#Kahn's_algorithm
+     * with the modification that when a dependency is removed it is only
+     * removed from one of its parents' run_before list since we don't have an
+     * easy reference to remove it from all parents. As a result we have to
+     * check that a pipeline hasn't akready been inserted since it can be
```
ehsantn: Minor typo
```suggestion
     * check that a pipeline hasn't already been inserted since it can be
```

bodo/pandas/physical/join_filter.h
```diff
@@ -138,21 +143,28 @@ class PhysicalJoinFilter : public PhysicalProcessBatch {
 
             if (this->materialize_after_each_filter && applied_any_filter) {
                 input_batch = RetrieveTable(input_batch, row_bitmask);
-                // Reset row_bitmask for next filter if not last iteration
-                if (i < this->filter_ids.size() - 1) {
```
ehsantn, line 142: Does this check cause a bug? Just curious.

bodo/pandas/optimizer/runtime_join_filter.h
```diff
@@ -0,0 +1,66 @@
...
+
+   private:
+    size_t cur_join_filter_id = 0;
+    using JoinFilterProgramState = std::unordered_map<int, JoinColumnInfo>;
+    JoinFilterProgramState join_state_map;
+
+    /**
+     * @brief Inserts join filters for all entries in join_state_map above the
+     * given operator op. Returns the resulting top operator, either op or the
+     * join filters. params
```
ehsantn: Minor typo.
```suggestion
     * join filters.
```

bodo/pandas/optimizer/runtime_join_filter.cpp
```diff
@@ -0,0 +1,288 @@
...
+                if (key_iter != left_keys.end()) {
+                    int64_t key_idx =
+                        std::distance(left_keys.begin(), key_iter);
+                    right_filter_cols.push_back(right_child_colref_map[{
+                        right_keys[key_idx].table_index,
+                        right_keys[key_idx].column_index}]);
+                    right_is_first_locations.push_back(true);
+                } else {
+                    right_filter_cols.push_back(-1);
+                    right_is_first_locations.push_back(false);
```
ehsantn, line 153: Could you explain these lines? What does -1 mean exactly again?

bodo/pandas/optimizer/runtime_join_filter.cpp
```diff
@@ -0,0 +1,288 @@
...
+                         /*filter_ids=*/filter_ids,
+                         /*filter_columns=*/filter_columns,
+                         /*is_first_locations=*/is_first_locations);
+    // Replace current operator with join filter
+    return std::move(join_filter);
+}
+
+duckdb::unique_ptr<duckdb::LogicalOperator>
+RuntimeJoinFilterPushdownOptimizer::VisitCompJoin(
+    duckdb::unique_ptr<duckdb::LogicalOperator> &op) {
```
ehsantn, line 77: I think this function could use a bit more comments throughout since quite complicated.

ehsantn (approved):
Thanks @IsaacWarren ! Looks good to me overall.

bodo/tests/test_df_lib/test_end_to_end.py
IsaacWarren: I wasn't trying to test join filters here, just modifying an existing test to pass with the new nodes. Testing the number of nodes in a plan isn't ideal IMO and we've known this. We should provide a way to inspect the nodes in a plan and compare to what we expect (in this case confirm we have a filter node before and no filter node after/the join's condition reflects the filter's condidtion)

bodo/pandas/physical/join_filter.h
IsaacWarren: It did because I made the bitmask allocation cache so it needs to be reset after every call so if the bitmask isn't reallocated the bitmask is in the correct state (all 1s)

bodo/pandas/optimizer/runtime_join_filter.cpp
IsaacWarren: It means the filter's already been evaluated

bodo/pandas/_executor.h
IsaacWarren: Sorry I thought I replied to this. I added a link to the algorithm I used, with a small modification. I used a vector instead of a set because I thought the number of pipelines would be small enough that hashing would outway the cost of just searching the vector

PR #951, opened 2025-12-02
TPCH benchmark scripts

scott-routledge2:

Adds/modifies scripts for running (Multi Node) Bodo, Dask, PySpark TPCH queries under the `tpch/<impl>/` directories
Adds/modifies scripts for running (Single node) Pandas, Polars, Dask, PySpark, Bodo, DuckDB under the `tpch/pds-benchmark` directory 

Adds explanation of directory structure, documentation on running scripts to top level README

## Testing strategy

Running everything locally/cluster with a mix of S3 and local paths

## User facing changes

No

## Checklist


benchmarks/tpch/README.md
```diff
@@ -47,54 +47,127 @@ This script assumes `tpch-dbgen` is in the same directory. If you downloaded it
...
+To run all queries for a specific library, you can use:
+``` shell
+SCALE_FACTOR=SF PATH_DATA_FOLDER=/path/to/you/tpch/data python -m queries.<IMPL>
+```
 
-Use
+When running all queries, each query will be run as a separate Python process and measured based on end-to-end time including IO. Each query will be run at least twice, the first time is considered a "cold start" and will not be logged when reporting the specific query time, although both hot and cold runs will be included in the "total time" reported at the end.
 
-`mpiexec -n N python bodo_queries.py --folder folder_path`
+Some implementations like Polars may have multiple modes (streaming, eager, etc) to run Polars with their new streaming mode:
```
IsaacWarren:
```suggestion
Some implementations like Polars may have multiple modes (streaming, eager, etc). To run Polars with their new streaming mode:
```

benchmarks/tpch/pds-benchmark/queries/pyspark/q1.py
```diff
@@ -0,0 +1,39 @@
+from queries.pyspark import utils
+
+Q_NUM = 1
+
+
+def q() -> None:
+    query_str = """
```
IsaacWarren: Ideally I think we'd use pyspark's python bindings instead of SQL for a more fair comparison, what would the effort be for that? I guess we could try and do one and see if the performance is any different

scott-routledge2: Maybe a couple hours at least? We probably won't need to implement from scratch e.g. use: https://github.com/dotnet/spark/blob/main/benchmark/python/tpch_functional_queries.py but would need to verify they are correct. Glancing at these queries, they use a lot of udfs which I'm guessing would have worse performance?

scott-routledge2: 15/22 queries ran after replacing the import and making some minor changes, so I'll replace spark sql with those queries for now and we can do the rest later

PR #1119, opened 2026-04-10
[GPU] create scalar fix for tpch q8.

DrTodd13:

We were creating string scalars in such a way that they were not valid and thus filters were matching no rows.

## Testing strategy

run_ci

## User facing changes

None

## Checklist

### DIFF
--- a/bodo/pandas/physical/gpu_expression.h
+++ b/bodo/pandas/physical/gpu_expression.h
@@ -103,10 +103,9 @@ std::unique_ptr<cudf::scalar> make_cudf_scalar_from_value(
     using T = std::decay_t<U>;
 
     if constexpr (std::is_same_v<T, std::string>) {
-        return std::make_unique<cudf::string_scalar>(value, se->stream);
+        return cudf::make_string_scalar(value, se->stream);
     } else if constexpr (std::is_same_v<T, const char *>) {
-        return std::make_unique<cudf::string_scalar>(std::string(value),
-                                                     se->stream);
+        return cudf::make_string_scalar(std::string(value), se->stream);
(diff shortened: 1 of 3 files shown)


bodo/pandas/physical/gpu_expression.h
```diff
@@ -103,10 +103,9 @@ std::unique_ptr<cudf::scalar> make_cudf_scalar_from_value(
     using T = std::decay_t<U>;
 
     if constexpr (std::is_same_v<T, std::string>) {
-        return std::make_unique<cudf::string_scalar>(value, se->stream);
+        return cudf::make_string_scalar(value, se->stream);
```
ehsantn: Why is using the constructor problematic? libcudf docs says you can use either:
https://github.com/rapidsai/cudf/blob/1ed06c6105fb31bba18fc7cf69e2529f9c6a7a22/cpp/doxygen/developer_guide/DEVELOPER_GUIDE.md#L388

ehsantn (commented):
See comment.

bodo/pandas/physical/gpu_expression.h
ehsantn: Should we replace all other scalar constructor uses?

DrTodd13: make_string_scalar passes true.  That is the only diff so reverting to old approach with true param like the others in this func.

IsaacWarren (approved):

ehsantn (approved):

bodo/pandas/physical/gpu_expression.h
ehsantn: Why didn't the compiler complain about data type mismatch?!

PR #1122, opened 2026-04-10
[GPU] BSE-5362: GPU Sort limit and offset

IsaacWarren:
Adds limit and offset to cuda sort, enabling TopN.

## Testing strategy
New unit test and tpch on pr ci

## User facing changes
Full GPU sort support

## Checklist

### DIFF
--- a/bodo/libs/streaming/cuda_sort.cpp
+++ b/bodo/libs/streaming/cuda_sort.cpp
@@ -16,11 +16,14 @@ CudaSortState::CudaSortState(
     std::shared_ptr<bodo::Schema> schema,
     std::vector<cudf::size_type> const& key_indices,
     std::vector<cudf::order> const& column_order,
-    std::vector<cudf::null_order> const& null_precedence)
+    std::vector<cudf::null_order> const& null_precedence, int64_t limit,
+    int64_t offset)
     : schema(std::move(schema)),
       key_indices(key_indices),
       column_order(column_order),
-      null_precedence(null_precedence) {}
+      null_precedence(null_precedence),
+      limit(limit),
+      offset(offset) {}
 
 void CudaSortState::ConsumeBatch(std::shared_ptr<cudf::table> table,
                                  std::shared_ptr<StreamAndEvent> input_se) {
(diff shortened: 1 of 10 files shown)


bodo/libs/streaming/cuda_sort.cpp
```diff
@@ -244,8 +294,19 @@ std::unique_ptr<cudf::table> CudaSortState::GetOutputBatch(
                 views.push_back(table->view());
             }
             // Multi-way Merge
-            final_result = cudf::merge(views, key_indices, column_order,
-                                       null_precedence, stream);
+            auto merged = cudf::merge(views, key_indices, column_order,
```
DrTodd13, line 325: So, just so I understand the algorithm.  If we know we don't have any portion of the output then we do nothing but if we have any portion of the output then we merge everything and do all the same work and only do something else at the end?  I guess other systems try to take advantage of this extra information somehow to reduce the work but we aren't.  I guess that would have to be a cudf option to merge right?  A brief search didn't show cudf as having this ability but can we use the samples in any way to slice the views before merging to potentially same some time?  Maybe not worth it to do now but maybe leave a note in the code?

DrTodd13 (approved):
Thanks!

bodo/libs/streaming/cuda_sort.cpp
IsaacWarren: Yeah exactly, cudf has a top_k but only on a contiguous table. I just added an optimization for ranks to only accumulate limit+offset rows so that will help when the offset is small but at large offsets it's still an issue. Will add a comment

docs/docs/guides/dataframes/gpu_acceleration.md
```diff
@@ -144,6 +144,10 @@ The listed aggregations (sum, count, mean, min, max, var, std, size, skew, nuniq
 
 Inner equi-joins are supported on GPU. Joins with non-equality predicates (range joins, inequality joins, or arbitrary expressions) are not supported on GPU and will run on CPU.
 
+### Sorting
```
scott-routledge2, line 145: I think this needs to be updated above too? (line 123) And we have been indicating GPU support on the API pages as well.

scott-routledge2 (approved):
LGTM, thanks @IsaacWarren !

docs/docs/guides/dataframes/gpu_acceleration.md
IsaacWarren: I didn't see where we have GPU support on the API pages? I looked at join and groupby

scott-routledge2: I think I was looking at https://docs.bodo.ai/latest/api_docs/dataframe_lib/io/

PR #1130, opened 2026-04-20
[GPU] Enable more dataframe tests to run on GPU.

DrTodd13:

Fix nested column type conversion from gpu to arrow bug.
Add a bunch more dataframe library test to be marked to run on GPU.

## Testing strategy

run_ci

## User facing changes

None

## Checklist

### DIFF
--- a/bodo/pandas/physical/operator.cpp
+++ b/bodo/pandas/physical/operator.cpp
@@ -313,14 +313,31 @@ GPU_DATA convertArrowTableToGPU(std::shared_ptr<arrow::Table> arrow_table,
     return GPU_DATA{std::move(result), arrow_batch->schema(), se};
 }
 
+cudf::column_metadata build_meta_from_arrow(
+    std::shared_ptr<arrow::Field> field) {
+    cudf::column_metadata meta;
+    meta.name = field->name();
+
+    if (field->type()->id() == arrow::Type::STRUCT) {
+        auto struct_type =
+            std::static_pointer_cast<arrow::StructType>(field->type());
+        for (int i = 0; i < struct_type->num_fields(); ++i) {
+            meta.children_meta.push_back(
+                build_meta_from_arrow(struct_type->field(i)));
+        }
+    }
+
+    return meta;
+}
+
 std::shared_ptr<arrow::Table> convertGPUToArrow(GPU_DATA batch) {
     cudf::table_view view = batch.table->view();
     // Setup Metadata (Arrow requires column names)
     // We must create a cudf::column_metadata hierarchy matching the table
     // structure.
     std::vector<cudf::column_metadata> meta;
-    for (const auto &name : batch.schema->field_names()) {
-        meta.emplace_back(name);
+    for (const auto &field : batch.schema->fields()) {
+        meta.push_back(build_meta_from_arrow(field));
     }
 
     cudf::unique_schema_t unique_schema = cudf::to_arrow_schema(view, meta);
--- a/bodo/tests/test_df_lib/test_direct_series_methods.py
+++ b/bodo/tests/test_df_lib/test_direct_series_methods.py
@@ -8,6 +8,7 @@ from bodo.tests.test_df_lib.series_test_generator import generate_series_test
 from bodo.tests.utils import _test_equal
 
 
+@pytest.mark.gpu
 @pytest.mark.parametrize("use_index1", [True, False])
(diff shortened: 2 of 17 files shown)


bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -195,6 +199,7 @@ def test_read_parquet_series_len_shape(datapath):
         assert bodo_out.shape == py_out.shape
 
 
+@pytest.mark.gpu(allow_fallback=True)
```
scott-routledge2: It would be good to note what is falling back so we can remove the marker once we enable more features. Here it looks like it's because of `to_datetime`?

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -645,6 +661,7 @@ def test_filter_datetime(datapath, op):
     )
 
 
+@pytest.mark.gpu(allow_fallback=True)
```
scott-routledge2: Assuming head tests are falling back because of limit pushdown? We may want to wait on enabling tests like this until something non-trivial is happening to save CI resources.

bodo/tests/test_s3.py
```diff
@@ -899,7 +899,7 @@ def read_table():
     )
 
 
-@pytest.mark.gpu
+@pytest.mark.gpu(allow_fallback=True)
```
scott-routledge2: Why do we need to enable fallback?

scott-routledge2 (approved):
Thanks @DrTodd13. I think it would be good to note why certain tests are falling back and ideally avoid tests where the main operator/behavior tested doesn't occur on GPU yet.

bodo/tests/test_df_lib/test_end_to_end.py
DrTodd13: There is a GPU-unimplemented count_star.

bodo/tests/test_s3.py
DrTodd13: Pretty sure I saw some kind of a fallback here but when I run it locally again I don't so removing it for now.

bodo/tests/test_df_lib/test_end_to_end.py
DrTodd13: None of the other places did it so I just followed the pattern and did the same.  I can see both sides on this one that it could be useful but also possible the comments atrophy.  Let's see what Ehsan thinks.

bodo/pandas/physical/operator.cpp
```diff
@@ -313,14 +313,31 @@ GPU_DATA convertArrowTableToGPU(std::shared_ptr<arrow::Table> arrow_table,
     return GPU_DATA{std::move(result), arrow_batch->schema(), se};
 }
 
+cudf::column_metadata build_meta_from_arrow(
+    std::shared_ptr<arrow::Field> field) {
+    cudf::column_metadata meta;
+    meta.name = field->name();
+
+    if (field->type()->id() == arrow::Type::STRUCT) {
```
ehsantn, line 321: I assume Arrow's Map and List types need to be handled here too. Let's handle them or just throw an error for now to avoid silent failures.

bodo/tests/test_df_lib/test_direct_series_methods.py
```diff
@@ -38,6 +39,7 @@ def test_series_isin(index_val, use_index1, use_index2):
     )
 
 
+@pytest.mark.gpu(allow_fallback=True)
```
ehsantn: I'd be good to add a comment on what is falling back to make it easier to revisit when more functionality is implemented.

bodo/tests/test_df_lib/test_tpch.py
```diff
@@ -183,7 +183,6 @@ def test_tpch_q19():
     run_tpch_query_test(tpch.tpch_q19, plan_executions=1)
 
 
-@pytest.mark.gpu
```
ehsantn, line 187: Q20 is now fixed with this PR: https://github.com/bodo-ai/Bodo/pull/1132/changes

bodo/tests/test_df_lib/test_end_to_end.py
ehsantn: I think it'd be good to comment on what is falling back even if other places haven't added comments yet. It'll make future development just a bit easier.

ehsantn (approved):
Thanks, Todd. See comments.

PR #1137, opened 2026-04-27
[GPU] Support dropna and avoid string sum in groupby

ehsantn:

As title. Fixes some test issues.

## Testing strategy

Enabled unit tests.

## User facing changes

Fixes groupby issues on GPU.

## Checklist

### DIFF
--- a/bodo/pandas/physical/gpu_aggregate.h
+++ b/bodo/pandas/physical/gpu_aggregate.h
@@ -36,6 +36,10 @@ inline bool gpu_capable(duckdb::LogicalAggregate& logical_aggregate) {
         return gpu_capable_reduce(logical_aggregate);
     }
 
+    std::map<std::pair<duckdb::idx_t, duckdb::idx_t>, size_t> col_ref_map =
+        getColRefMap(logical_aggregate.children[0]->GetColumnBindings());
+    logical_aggregate.children[0]->ResolveOperatorTypes();
+
     for (size_t i = 0; i < logical_aggregate.expressions.size(); i++) {
         const auto& expr = logical_aggregate.expressions[i];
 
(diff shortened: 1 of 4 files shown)


bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -1607,7 +1608,7 @@ def test_series_groupby(dropna, as_index):
         df1 = pd.DataFrame(
             {
                 "B": ["a1", "b11", "c111"] * 2,
-                "E": pd.array([1.1, pd.NA, 13.3, pd.NA, pd.NA, 13.3], "Float64"),
+                "E": pd.array([1.1, 4.1, 13.3, 1.9, 3.5, 13.3], "Float64"),
```
ehsantn, line 1611: All NA data in a group produces NA in cudf but zero in Pandas. Not a big deal right now I think.

bodo/pandas/physical/gpu_aggregate.h
```diff
@@ -51,6 +55,26 @@ inline bool gpu_capable(duckdb::LogicalAggregate& logical_aggregate) {
...
+        }
+        auto& colref = child_expr->Cast<duckdb::BoundColumnRefExpression>();
+
+        uint64_t col_idx = col_ref_map.at(
+            {colref.binding.table_index, colref.binding.column_index});
+
+        // cudf doesn't support sum aggregation on string columns
+        if (agg_expr.function.name == "sum" &&
+            logical_aggregate.children[0]->types[col_idx].id() ==
+                duckdb::LogicalTypeId::VARCHAR) {
```
DrTodd13, line 75: Isn't there a fixed length string and blob we should include here?  Also, other functions don't support strings either so add those above?

DrTodd13 (approved):

scott-routledge2 (approved):
Thanks @ehsantn !

bodo/pandas/physical/gpu_aggregate.h
ehsantn: We only use DuckDB's `VARCHAR` data type. Also, string sum is an exception that works on CPUs and not GPUs. I don't know any other cases (count of strings works I think for example).

PR #1141, opened 2026-04-29
[GPU] Different GPU batch size adaptation.

DrTodd13:

Divide GPU memory size by a constant to get batch size.  Simple adaptation for different GPUs.

## Testing strategy

run_ci

## User facing changes

Added BODO_GPU_STREAMING_BATCH_SIZE_DIVISOR to control how many GPU bytes get allocated per batch row.  So, can control GPU batch size this way or by setting the absolute size with the existing env var.  This way allows you more control when your GPUs have differing memory amounts.

## Checklist

### DIFF
--- a/bodo/pandas/physical/operator.cpp
+++ b/bodo/pandas/physical/operator.cpp
@@ -5,6 +5,7 @@
 #include <string>
 
 #ifdef USE_CUDF
+#include <cuda_runtime.h>
 #include <cudf/concatenate.hpp>
 #include <cudf/copying.hpp>
 #include <cudf/table/table.hpp>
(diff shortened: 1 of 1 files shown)


bodo/pandas/physical/operator.cpp
```diff
@@ -24,11 +25,43 @@ int get_streaming_batch_size() {
...
+        // pretty consistently has problems particularly if there are CTEs
+        // involved.  Allocating 2000 per row + overheads is about 10% slower
+        // but with even less likelihood of memory problems.
+        int divisor = (env_str != nullptr) ? std::stoi(env_str) : 2000;
+        default_result = total_bytes / divisor;
+    });
+
     char *env_str = std::getenv("BODO_GPU_STREAMING_BATCH_SIZE");
-    return (env_str != nullptr) ? std::stoi(env_str) : 24.e6;
+    return (env_str != nullptr) ? std::stoi(env_str) : default_result;
```
ehsantn, line 54: Does this calculation result in something similar to the 24m number on typical cloud instances?

ehsantn (approved):

bodo/pandas/physical/operator.cpp
DrTodd13: My understanding is that we previously looked at 24GB memory on L4 instances and guessed a nice round 1000 number and so ended up with 24 million rows.  We did a little testing with it and then hard-coded it.  On L4, this PR would select 12 million.  On the GB200's, it would be 2x what it currently is (48M instead of 24M).

ehsantn: Sounds good.

scott-routledge2 (approved):
Thanks @DrTodd13 !

PR #1142, opened 2026-04-29
[GPU] Handle case of unmatched_build_rows being empty.

DrTodd13:

Handle join case where unmatched_build_rows is nullptr.

## Testing strategy

run_ci

## User facing changes

None

## Checklist

### DIFF
--- a/bodo/libs/streaming/cuda_join.cpp
+++ b/bodo/libs/streaming/cuda_join.cpp
@@ -193,6 +193,10 @@ std::unique_ptr<cudf::table> CudaJoin::produce_unmatched_build_rows(
 
     cudf::table_view build_kept_view = _build_table->select(
         this->build_kept_cols.begin(), this->build_kept_cols.end());
+
+    if (unmatched_build_rows == nullptr) {
+        return table;
+    }
     // For right and outer joins, we need to output unmatched build rows
     // at the end. We can identify these using the matched_build_rows
     // boolean mask.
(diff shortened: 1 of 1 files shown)


IsaacWarren (commented):
I don't think this case should ever happen, if it's a join that propagates build unmatched_build_rows should not be a nullptr, otherwise it'll just return at line 161. If you want me help debug why this happens I'm happy to take a look

IsaacWarren (approved):

scott-routledge2:
> I don't think this case should ever happen, if it's a join that propagates build unmatched_build_rows should not be a nullptr, otherwise it'll just return at line 161. If you want me help debug why this happens I'm happy to take a look

This bug started happening after [this line](https://github.com/bodo-ai/Bodo/blob/3c92280fc8e60b30fcd6aa0b03a2e69d45798f76/bodo/pandas/_executor.cpp#L138) was added. Maybe a type mismatch somewhere?

scott-routledge2 (approved):
Thanks Todd

IsaacWarren:
> > I don't think this case should ever happen, if it's a join that propagates build unmatched_build_rows should not be a nullptr, otherwise it'll just return at line 161. If you want me help debug why this happens I'm happy to take a look
> 
> This bug started happening after [this line](https://github.com/bodo-ai/Bodo/blob/3c92280fc8e60b30fcd6aa0b03a2e69d45798f76/bodo/pandas/_executor.cpp#L138) was added. Maybe a type mismatch somewhere?

Sorry should've updated my comment, it only happens when the table is empty so it makes sense for it to be nullptr

PR #1143, opened 2026-04-30
[GPU] Fix groupby test hang

ehsantn:

Fixing the hang needed duration data type support.

## Testing strategy

Enabled unit test.

## User facing changes

Bug fix. More timedelta support.

## Checklist

### DIFF
--- a/bodo/tests/test_df_lib/test_end_to_end.py
+++ b/bodo/tests/test_df_lib/test_end_to_end.py
@@ -1700,9 +1700,9 @@ def test_groupby_fallback():
 def groupby_agg_df(request):
     return pd.DataFrame(
         {
-            "A": pd.array([1, 2, pd.NA, 2147483647] * 3, "Int32"),
+            "A": pd.array([1, 2, 0, 2147483647] * 3, "Int32"),
             "D": pd.array(
-                [i * 2 if (i**2) % 3 == 0 else pd.NA for i in range(12)], "Int32"
+                [i * 2 if (i**2) % 3 == 0 else 0 for i in range(12)], "Int32"
             ),
             "B": pd.array(["A", "B", pd.NA] * 4),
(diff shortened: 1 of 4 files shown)


bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -1700,9 +1700,9 @@ def test_groupby_fallback():
 def groupby_agg_df(request):
     return pd.DataFrame(
         {
-            "A": pd.array([1, 2, pd.NA, 2147483647] * 3, "Int32"),
+            "A": pd.array([1, 2, 0, 2147483647] * 3, "Int32"),
```
DrTodd13, line 1703: What precedent are we setting by removing the NA tests?  That's a pretty big thing right and doing this because we will never get that to work on GPU?

DrTodd13 (approved):

bodo/tests/test_df_lib/test_end_to_end.py
ehsantn: The problem is a minor behavior difference between Pandas and cuDF. Pandas produces zero for all-NA groups but cuDF produces NA. This is minor I think. NA handling is fine otherwise.

IsaacWarren: Can we just fix it after cudf runs? IMO this isn't complicated enough to warrant breaking compatibility

ehsantn: I don't think it's easy to fix it. We'd have to detect all-NA groups and replace the output with zero of the proper data type for sum.

IsaacWarren: Ok then I think we at least need to document it

ehsantn: Done.

IsaacWarren (approved):

PR #1145, opened 2026-05-01
[GPU] BSE-5381: Iceberg read

IsaacWarren:

Adds iceberg read GPU operator.

## Testing strategy
Existing df lib iceberg tests

## User facing changes
As title, updated docs.

## Checklist


DrTodd13 (approved):
Awesome.  Nice to have this!

bodo/pandas/physical/gpu_read_iceberg.cpp
```diff
@@ -0,0 +1,1094 @@
...
+    if (leftover_tbl) {
+        std::size_t n = leftover_tbl->num_rows();
+        if (n <= target_rows_) {
+            rows_accum += n;
+            gpu_tables.emplace_back(std::move(leftover_tbl));
+            leftover_tbl = nullptr;
+        } else {
+            cudf::table_view tv = leftover_tbl->view();
+            std::vector<cudf::table_view> sliced =
+                cudf::slice(tv, {0, (int)target_rows_, (int)n}, se->stream);
```
scott-routledge2: For stream aware operations I think we will eventually need to sync batch stream (se) with the stream the chunked_parquet_reader is using. (For example: https://github.com/bodo-ai/Bodo/blob/01d594372b112595a8386a4d6fc7cb03d4e6b234/bodo/pandas/physical/gpu_read_parquet.h#L237)

bodo/pandas/physical/gpu_read_iceberg.cpp
```diff
@@ -0,0 +1,1094 @@
...
+    opts.set_column_names(file_col_names);
+
+    if (filter_ast_tree.size() > 0) {
+        opts.set_filter(filter_ast_tree.back());
+    }
+
+    std::vector<std::unique_ptr<cudf::io::datasource>> sources;
+    if (filesystem_->type_name() != "local") {
+        std::shared_ptr<arrow::io::RandomAccessFile> arrow_file =
+            filesystem_->OpenInputFile(path).ValueOrDie();
```
scott-routledge2: In my experience creating 1 chunked reader per file can be slow. Is it possible to group pieces with the same schema into a single chunked reader? (Assuming grouped pieces are relatively small)

bodo/pandas/physical/gpu_read_iceberg.cpp
```diff
@@ -0,0 +1,1094 @@
...
+                curr_reader->read_chunk();
+            std::unique_ptr<cudf::table> tbl =
+                std::move(table_and_metadata.tbl);
+
+            if (!tbl || tbl->num_rows() == 0) {
+                continue;
+            }
+
+            if (rows_to_skip > 0) {
+                // Some pieces may have a row offset (e.g. from file-level
```
scott-routledge2: For file level filtering is row offset always equal to number of rows in the file?

scott-routledge2 (approved):
Thanks @IsaacWarren !

bodo/pandas/physical/gpu_read_iceberg.cpp
IsaacWarren: I did try doing that originally based on schema groups but cudf complained about mismatched schemas, it must not do upcasting in the same way arrow does. I do see an option that might fix it though, let me try it

bodo/pandas/physical/gpu_read_iceberg.cpp
IsaacWarren: I think that comment's just wrong, file level filtering should just remove the file from the list, thanks for catching

bodo/pandas/physical/gpu_read_iceberg.cpp
IsaacWarren: I now have it group on physical schemas instead of schema group, let me know what you think

bodo/pandas/physical/gpu_read_iceberg.cpp
```diff
@@ -318,18 +327,28 @@ void GPUIcebergRankBatchGenerator::init_next_reader(
     cudf::io::parquet_reader_options opts =
         cudf::io::parquet_reader_options::builder(cudf::io::source_info());
     opts.set_column_names(file_col_names);
+    opts.enable_allow_mismatched_pq_schemas(true);
```
scott-routledge2, line 334: Is this still needed if schemas are now identical?

bodo/pandas/physical/gpu_read_iceberg.cpp
scott-routledge2: Nice! I think chunked_parquet_reader requires memory equal to some fraction of the total size of all pieces (chunked_reader_limit only controls the output size). So if we expect a lot of large files with identical schemas this could run out of memory.  

For parquet read we use the estimated average file size to create batches of files and then create chunked readers for each of those batches. https://github.com/bodo-ai/Bodo/blob/10e0ac825057d47e609f485d36d7ca7b1e7bbeb7/bodo/pandas/physical/gpu_read_parquet.h#L390

But otherwise looks good.

bodo/pandas/physical/gpu_read_iceberg.cpp
IsaacWarren: It is, I think the number of columns can not match

bodo/pandas/physical/gpu_read_iceberg.cpp
IsaacWarren: Good call, done

PR #1146, opened 2026-05-01
[GPU] Fix broadcast join segfault from OpenMPI/UCX

scott-routledge2:

* Replace CUDA-aware MPI_Iallreduce with copy to host.
* Sort build table in the broadcast join case to make sure build_indices are consistent across all ranks

## Testing strategy

TPCH Q22 runs and SF10 results are correct on a 4 GPU Instance.

## Checklist

### DIFF
--- a/bodo/libs/streaming/cuda_join.cpp
+++ b/bodo/libs/streaming/cuda_join.cpp
@@ -166,19 +166,15 @@ std::unique_ptr<cudf::table> CudaJoin::produce_unmatched_build_rows(
     const MPI_Comm comm = this->probe_shuffle_manager->get_mpi_comm();
     if (this->is_broadcast_join && !this->build_matches_synced) {
         if (this->sync_build_matches_req == MPI_REQUEST_NULL) {
-            this->n_build_rows = this->unmatched_build_rows->size();
-            this->unmatched_build_rows_contents =
-                std::make_unique<cudf::column::contents>(
-                    this->unmatched_build_rows->release());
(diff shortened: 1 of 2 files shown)


bodo/libs/streaming/cuda_join.cpp
```diff
@@ -291,6 +305,24 @@ std::pair<std::unique_ptr<cudf::table>, bool> CudaJoin::materialize_and_output(
...
+            "initialized");
+    }
+
+    auto unsorted_build = std::move(this->_build_table);
+    auto order = cudf::sorted_order(
+        unsorted_build->view(),
+        std::vector<cudf::order>(unsorted_build->num_columns(),
+                                 cudf::order::ASCENDING),
+        std::vector<cudf::null_order>(unsorted_build->num_columns(),
+                                      cudf::null_order::AFTER));
```
IsaacWarren, line 321: Probably doesn't matter much but we could only sort on a single column, right?

IsaacWarren (approved):
Thanks Scott

bodo/libs/streaming/cuda_join.cpp
DrTodd13: Single column doesn't sound safe to me.

scott-routledge2: It depends. If there are duplicates in the column we choose then the sorted outputs are not guaranteed to be the same across ranks.

DrTodd13 (approved):

PR #1153, opened 2026-05-07
[GPU] Avoid cuda calls on non-GPU ranks.

DrTodd13:

1. Incorporate ucx env var.
2. Eliminate non-GPU ranks from allocations on a GPU.

## Testing strategy

run_ci

## User facing changes

Don't need to manually assign the environment variable anymore for OMPI.

## Checklist


bodo/__init__.py
```diff
@@ -283,6 +283,8 @@ def get_sql_config_str() -> str:
 os.environ["OMP_NUM_THREADS"] = "1"
 os.environ["MKL_NUM_THREADS"] = "1"
 
+if gpu_enabled:
```
scott-routledge2, line 286: Have you tested this with `mpiexec -n 1 ...` also? I remember at some point seeing workers not inheriting the environment from the spawner if launched via mpiexec.

scott-routledge2 (approved):
Thanks @DrTodd13 !

bodo/pandas/physical/operator.cpp
```diff
@@ -32,22 +32,28 @@ int get_gpu_streaming_batch_size() {
...
+            // tables yields a good performance with low-risk of crashing or
+            // hanging due to memory related problems.  Allocating 500 bytes per
+            // row + overheads pretty consistently has problems particularly if
+            // there are CTEs involved.  Allocating 2000 per row + overheads is
+            // about 10% slower but with even less likelihood of memory
+            // problems.
+            int divisor = (env_str != nullptr) ? std::stoi(env_str) : 2000;
+            default_result = total_bytes / divisor;
+        } else {
+            default_result = 0;
```
ehsantn: Is it ok for non-GPU ranks to have GPU batch size of 0? No problems in CPU/GPU exchange operators? Maybe some non-zero default value to be safe?

CMakeLists.txt
```diff
@@ -904,6 +904,27 @@ if(USE_CUDF)
...
+
+  # Fallback: also search default system paths
+  if(NOT NVML_LIB)
+      find_library(NVML_LIB NAMES nvidia-ml)
+  endif()
+
+  if(NOT NVML_LIB)
+      message(FATAL_ERROR "Could not find libnvidia-ml.so (NVML). Install NVIDIA driver.")
+  endif()
+  message(STATUS "libnvidia-ml lib: ${NVML_LIB}")
```
ehsantn, line 926: Is it safe to try to find the NVML library in system paths? Conda doesn't provide it? Not part of cuda-toolkit package for example?

ehsantn (commented):
Thanks Todd. My main concern is finding NVML library in the system paths. Is there a better way?

bodo/pandas/physical/operator.cpp
DrTodd13: done

CMakeLists.txt
DrTodd13: done

ehsantn (approved):

bodo/__init__.py
DrTodd13: Seems to be propagated but likewise I see the issue you saw for BODO_GPU and some other env vars.

scott-routledge2: Sounds good. Also, do we need to update our documentation for setting this variable?

CMakeLists.txt
DrTodd13: Actually, I don't know why I thought this worked at first but subsequent runs were failing because nvidia-ml paradoxically doesn't provide libnvidia-ml.so.  Do you know a clean way to search all of conda forge?  I tried a few things but nothing worked.  I'm back to thinking that this isn't in any conda package and need to rely on the local nvidia installation.

ehsantn (commented):
What about this package?
https://anaconda.org/channels/nvidia/packages/cuda-nvml-dev/overview

DrTodd13:
> What about this package? https://anaconda.org/channels/nvidia/packages/cuda-nvml-dev/overview

I think that is a stubs-only library.

PR #1162, opened 2026-05-13
Support test_comparison.py in BodoSQL C++ backend

ehsantn:

As title. Added Series.fillna() support too but didn't end up reusing it.

## Testing strategy

Enabled unit tests.

## User facing changes

None.

## Checklist


BodoSQL/bodosql/tests/test_comparison.py
```diff
@@ -231,12 +232,12 @@ def time_comparison_args(comparison_query_args):
     data = [
         None,
         Time(0, 0, 0, precision=9),
-        Time(12, 30, 15, nanosecond=13, precision=9),
```
ehsantn, line 234: Pandas testing infrastructure doesn't support nano/micro/milli differences for Time but otherwise the code works.

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -496,9 +523,22 @@ def java_literal_to_python_literal(java_literal, input_plan, out_pa_type=None):
...
+        dummy_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.duration("ns")))
+        # getValue2() returns an integer representing milliseconds
+        val = pd.to_timedelta(int(java_literal.getValue2()), unit="ms")
+        return ConstantExpression(dummy_empty_data, input_plan, val)
+
     if lit_type_name.equals(SqlTypeName.NULL):
-        assert out_pa_type is not None, "Output type must be provided for NULL literals"
-        dummy_empty_data = pd.Series(dtype=pd.ArrowDtype(out_pa_type))
+        dummy_empty_data = pd.Series(
+            dtype=pd.ArrowDtype(sql_type_to_pa_type(lit_type.getSqlTypeName()))
```
ehsantn, line 541: Looks like this type info is enough.

bodo/pandas/physical/expression.cpp
```diff
@@ -374,8 +395,10 @@ std::shared_ptr<array_info> do_arrow_compute_case(
             "do_arrow_compute else is neither array nor scalar.");
     }
 
+    // NOTE: Arrow's "if_else" doesn't match our Python and SQL semantics since
+    // it propagates nulls in the condition.
     arrow::Result<arrow::Datum> cmp_res =
-        arrow::compute::CallFunction("if_else", {src1, src2, src3});
+        arrow::compute::CallFunction("case_when", {src1, src2, src3});
```
DrTodd13, line 401: Should we perhaps specialize the common case of a single condition because I believe that kernel is faster than the generalized case?

bodo/pandas/series.py
```diff
@@ -1277,6 +1277,20 @@ def isin(self, values):
         values = pa.array(values, type=self.head(0).dtype.pyarrow_dtype)
         return _get_series_func_plan(self._plan, new_metadata, "isin", (values,), {})
 
+    @check_args_fallback(supported=["value"])
+    def fillna(self, value, *, axis=None, inplace: bool = False, limit=None):
```
DrTodd13, line 1281: Docs update for this?

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -353,6 +368,22 @@ def java_binop_to_python_expr(kind, op_exprs):
     raise NotImplementedError(f"Binary operator {kind.toString()} not supported yet")
 
 
+def java_case_to_python_case(operands, input_plan):
+    """Convert a BodoSQL Java CASE operator call to a DataFrame library CaseExpression.
+    operands has the form [when1, then1, when2, then2, ..., else].
+    """
+    assert len(operands) >= 3, "CASE operator should have at least 3 operands"
```
DrTodd13, line 375: Could check len(operands) % 2 == 1 as well.

DrTodd13 (approved):
Thanks Ehsan!

bodo/pandas/physical/expression.cpp
```diff
@@ -291,6 +291,20 @@ arrow::Datum do_arrow_compute_unary(
         return invert_res.ValueOrDie();
     }
 
+    // Special handling for is_true since it is not supported directly
+    // by Arrow compute.
+    if (comparator == "is_true") {
+        auto arrow_false = arrow::MakeScalar(false);
+        arrow::Result<arrow::Datum> is_true_res = arrow::compute::CallFunction(
+            "coalesce", {src1, arrow_false}, func_options);
```
IsaacWarren, line 299: Logical AND might be more straightforward?

bodo/pandas/vendor/duckdb/src/include/duckdb/common/enums/expression_type.hpp
```diff
@@ -151,7 +151,10 @@ enum class ExpressionType : uint8_t {
 	LAMBDA = 231,
 	POSITIONAL_REFERENCE = 232,
 	BOUND_LAMBDA_REF = 233,
-	BOUND_EXPANDED = 234
+	BOUND_EXPANDED = 234,
+
+	// Bodo change: added new operator
+	OPERATOR_IS_TRUE = 235
```
IsaacWarren:
```suggestion
	OPERATOR_IS_TRUE = 250
```
To make merging easier if they add more expression types

bodo/pandas/_plan.cpp
```diff
@@ -141,6 +141,18 @@ duckdb::unique_ptr<duckdb::Expression> make_const_timestamp_ns_expr(
...
+    int64_t val) {
+    if (val % 1000 != 0) {
+        throw std::runtime_error(
+            "make_const_timedelta_ns_expr only supports values with "
+            "microsecond precision since duckdb::Value::INTERVAL only supports "
+            "microsecond precision");
+    }
+    return duckdb::make_uniq<duckdb::BoundConstantExpression>(
+        duckdb::Value::INTERVAL(duckdb::Interval::FromMicro(val / 1000)));
+}
```
IsaacWarren:
```suggestion
    std::lldiv_t div_res = std::div(val, 1000);
    if (div_res.rem % 1000 != 0) {
        throw std::runtime_error(
            "make_const_timedelta_ns_expr only supports values with "
            "microsecond precision since duckdb::Value::INTERVAL only supports "
            "microsecond precision");
    }
    return duckdb::make_uniq<duckdb::BoundConstantExpression>(
        duckdb::Value::INTERVAL(duckdb::Interval::FromMicro(div_res.quot)));
}
```

bodo/pandas/physical/expression.cpp
DrTodd13: If the bodosql side just makes the multi-case instances into a series of if/then/else then why the conversion to case_when when the comment says it doesn't handle nulls correctly?

BodoSQL/bodosql/tests/test_comparison.py
IsaacWarren: There's no way to do it with arrow?

IsaacWarren (approved):
Thanks @ehsantn

bodo/pandas/physical/expression.cpp
ehsantn: `if_else` doesn't work since it propagates condition nulls to output so we have to use `case_when`. Is the comment not clear? BodoSQL makes multi-when a single expression. We can optimize multi-when case by converting into a single DuckDB custom expression but not high priority right now.

bodo/pandas/physical/expression.cpp
ehsantn: We want nulls to be replaced by False so AND wouldn't work.

BodoSQL/bodosql/tests/test_comparison.py
ehsantn: Pandas converts Arrow data types to Python objects in testing functions unfortunately.

bodo/pandas/series.py
ehsantn: Done.

bodo/pandas/vendor/duckdb/src/include/duckdb/common/enums/expression_type.hpp
ehsantn: Done.

PR #1163, opened 2026-05-13
[GPU][BSE-5416] Fix issues related to multiple workers per GPU and re-enabled tests

scott-routledge2:

Fix segfaults due to non-GPU workers returning nullptr for data/streams to avoid creating contexts
Enable some GPU tests for multiple workers-per GPU case. More tests will be enabled in a followup PR

## Testing strategy

PR CI

## Checklist

### DIFF
--- a/bodo/tests/test_df_lib/test_gpu/test_gpu_end_to_end.py
+++ b/bodo/tests/test_df_lib/test_gpu/test_gpu_end_to_end.py
@@ -122,11 +122,10 @@ def test_cpu_to_gpu_exchange(datapath):
     assert count_gpu_plan_nodes(bdf3._plan) == 4, (
         "Expected 4 GPU nodes (ReadParquet, JoinFilter, Join, Project)"
     )
-    bdf3 = bdf3.sort_values(bdf3.columns.to_list())
 
     pdf2 = pd.read_parquet(path)
     pdf3 = pdf1.merge(pdf2, how="inner", on="A")
-    _test_equal(pdf3, bdf3)
+    _test_equal(bdf3, pdf3)
 
     # Case 2: CPU (read pandas) -> GPU sink (write parquet)
     with tempfile.TemporaryDirectory() as tmp:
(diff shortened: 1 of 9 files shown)


bodo/tests/test_df_lib/test_gpu/test_gpu_end_to_end.py
```diff
@@ -126,7 +126,7 @@ def test_cpu_to_gpu_exchange(datapath):
 
     pdf2 = pd.read_parquet(path)
     pdf3 = pdf1.merge(pdf2, how="inner", on="A")
-    _test_equal(pdf3, bdf3)
+    _test_equal(bdf3, pdf3)
```
IsaacWarren, line 129: What does changing the order do?

IsaacWarren (approved):
Thanks Scott

bodo/tests/test_df_lib/test_gpu/test_gpu_end_to_end.py
scott-routledge2: We convert the first argument (bodo_result) to Pandas. Without this step then the results won't be sorted properly because of the other operations we do after.

bodo/tests/utils.py
```diff
@@ -1330,6 +1330,17 @@ def _test_equal(
     except ImportError:
         csr_matrix = type(None)
 
+    if bodo.gpu_enabled:
+        import cupy as cp
+
+        num_devices = cp.cuda.runtime.getDeviceCount()
```
DrTodd13, line 1336: I guess extra allocations on GPU don't matter during these tests?  I think you can use nvml here from Python as well (pynvml?) if we want to avoid those extra allocations.  Let you decide if it is worth it.

DrTodd13 (approved):
Thanks Scott!

bodo/tests/utils.py
scott-routledge2: Good point, I think this would be pretty simple change right?

scott-routledge2: Although I was imagining we would check workers for extra CUDA contexts somehow, and ignore extra contexts from the Spawner?

PR #1167, opened 2026-05-14
Support aggregation function lists in agg function dict [GPU]

dmsa-or:
We now handle the case where there are multiple aggregations for a column, specified using the dictionary syntax where the value is the list of aggregation functions for that column. A test case was added to spot when this functionality does not work.

## User facing changes
A new aggregation function combination is supported.

## Checklist


DrTodd13:
Maybe permission problem due to non-Bodo account?

scott-routledge2:
> Maybe permission problem due to non-Bodo account?

PRs from forks need explicit permission to run workflows. I just enabled it.

PR #1171, opened 2026-05-15
[GPU] Fix shuffle hangs in limit pipelines

scott-routledge2:

* Fix GPU limit to ensure that all ranks have at least N rows (or prev op is finished) 
* Clean shuffle state after early termination via a limit
* Avoid sorting on dict columns in testing utilities 
* Logic bug in GPU reduce

## Testing strategy

PR CI

## Checklist

### DIFF
--- a/bodo/libs/streaming/_shuffle.h
+++ b/bodo/libs/streaming/_shuffle.h
@@ -192,23 +192,6 @@ class AsyncShuffleSendState {
         return total_size;
     }
 
-    /**
-     * @brief Cancel and cleanup inflight send requests.
-     *
-     */
-    void CancelSends() {
-        for (MPI_Request& req : send_requests) {
-            CHECK_MPI(MPI_Cancel(&req),
-                      "AsyncShuffleSendState::CancelSends: MPI error on "
-                      "MPI_Cancel: ");
-        }
-
-        CHECK_MPI(
-            MPI_Waitall(send_requests.size(), send_requests.data(),
-                        MPI_STATUSES_IGNORE),
-            "AsyncShuffleSendState::CancelSends: MPI error on MPI_Waitall: ");
-    }
-
    private:
     using comm_infos_t =
         std::tuple<std::vector<mpi_comm_info>, std::vector<mpi_str_comm_info>>;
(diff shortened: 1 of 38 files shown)


DrTodd13 (approved):
LGTM.  Thanks!

bodo/libs/streaming/_shuffle.h
```diff
@@ -192,6 +192,23 @@ class AsyncShuffleSendState {
         return total_size;
     }
 
+    /**
+     * @brief Cancel and cleanup inflight send requests.
+     *
+     */
+    void CancelSends() {
```
scott-routledge2: While MPI_Cancel + Wait should be a local operation this seems to hang when I try many workers. I removed the CancelShuffle() call and everything passed, so maybe it is ok to remove this part for now and we can add it back only when necessary. 

For now, it seems like the limit termination condition + finish the barrier if necessary is sufficient.

bodo/pandas/physical/operator.cpp
```diff
@@ -555,7 +555,20 @@ RankDataExchange::RankDataExchange(int64_t op_id_) : op_id(op_id_) {
     }
 }
 
-RankDataExchange::~RankDataExchange() { MPI_Comm_free(&this->shuffle_comm); }
+RankDataExchange::~RankDataExchange() {
+    if (!finished) {
+        while (!sync_is_last_non_blocking(is_last_state.get(), 1)) {
+            // Wait for all ranks to get to global is_last barrier before
+            // finalizing shuffle state
+        };
```
IsaacWarren, line 563: Could this deadlock if two operators use rankdataexchange? i.e. one rank ends up in the desctructor for this 1 operator and one rank ends up in the desctructor of the other

IsaacWarren (approved):
Thanks @scott-routledge2

bodo/pandas/physical/operator.cpp
scott-routledge2: You're right this is possible. This assumes each rank destroys the rankdataexchange object in the same order, which I think is usually true because the pipelines are the same across ranks. Maybe a safer place to do this would be an explicit Finalize function (though that would require a lot of changes potentially). Thoughts?

IsaacWarren: We have finalize for operators though right? Can we call a rankdataexchange finalize inside of the operator finalize. I think operator finalization order should be guaranteed

scott-routledge2: Correct, I meant we'd call the rankdataexchange finalize in the base operator Finalize and then all subclasses would have to be changed to call the base class finalize.

scott-routledge2: In my opinion at this point the RankDataExchanges would be their own physical operators, but we can probably do that at a later point.

IsaacWarren (approved):

PR #1173, opened 2026-05-18
Refactor function expression creation in plans

ehsantn:

Refactors function expressions in the Python plan to take input expressions instead of a source plan and column indices. This provides a lot more flexibility which is necessary for translating Java plans to Python.

## Testing strategy

Existing tests.

## User facing changes

None.

## Checklist

### DIFF
--- a/bodo/pandas/plan.py
+++ b/bodo/pandas/plan.py
@@ -77,11 +77,11 @@ class LazyPlan:
             col_index = args[1]
             return f"ColRefExpression({col_index})"
         elif isinstance(self, PythonScalarFuncExpression):
-            func_name, col_indices = args[1][0], args[2]
-            return f"PythonScalarFuncExpression({func_name}, {col_indices})"
+            func_name = args[1][0]
+            return f"PythonScalarFuncExpression({func_name})"
         elif isinstance(self, ArrowScalarFuncExpression):
-            func_name, col_indices = args[2], args[1]
-            return f"ArrowScalarFuncExpression({func_name}, {col_indices})"
+            func_name = args[1]
+            return f"ArrowScalarFuncExpression({func_name})"
 
         out = f"{self.plan_class}: \n"
         args_str = ""
@@ -679,8 +679,8 @@ class PythonScalarFuncExpression(Expression):
     """Expression representing a Python scalar function call in the query plan."""
 
     @property
-    def source(self):
-        """Return the source of the expression."""
+    def input_exprs(self):
+        """Return the input expressions of the function."""
         return self.args[0]
 
     @property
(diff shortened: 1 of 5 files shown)


bodo/pandas/plan.py
```diff
@@ -749,35 +766,46 @@ class ArrowScalarFuncExpression(Expression):
...
+    def source(self):
+        """Return the source of the function expression assuming all input expressions share the same source."""
+        assert len(self.input_exprs) != 0, (
+            "ArrowScalarFuncExpression::source: expected at least one input expression"
+        )
+
+        source = self.input_exprs[0].source
+        for expr in self.input_exprs[1:]:
+            if expr.source != source:
+                return None
```
DrTodd13: source can return None now but I don't see any new checks added anywhere for this possibility.  There are some places that replace the source and those seem to check new source with old source but if the old source is mixed and therefore returns None then they won't be equal and you'll change all of the source to the new specified source.  Any reason that this is guaranteed to be safe or correct?  Seems like we should have some new tests for the case where all the sources aren't the same.

DrTodd13 (approved):
Isaac mentioned something about the issue Daniel filed but it seems like this PR is a step in support for that as well.  So, is the old requirement that the exprs all come directly from the source dataframe still hold?

bodo/pandas/plan.py
ehsantn: Currently, `source` should be the same for all expressions since this is just refactoring the previous code which had a source operator argument. This check is just in case for future and should not be possible right now.

ehsantn:
> Isaac mentioned something about the issue Daniel filed but it seems like this PR is a step in support for that as well. So, is the old requirement that the exprs all come directly from the source dataframe still hold?

This PR is just refactoring and doesn't change expression requirements. We have transformations elsewhere for specific cases that make sure the source dataframe is matched.

bodo/pandas/plan.py
```diff
@@ -679,40 +679,51 @@ class PythonScalarFuncExpression(Expression):
...
+    def source(self):
+        """Return the source of the function expression assuming all input expressions share the same source."""
+        assert len(self.input_exprs) != 0, (
+            "PythonScalarFuncExpression::source: expected at least one input expression"
+        )
+
+        source = self.input_exprs[0].source
+        for expr in self.input_exprs[1:]:
+            if expr.source != source:
+                return None
```
IsaacWarren: Should this throw an error?

bodo/pandas/plan.py
IsaacWarren: If it shouldn't be possible shouldn't we just throw an error if it happens?

IsaacWarren (approved):
Thanks Ehsan

bodo/pandas/plan.py
ehsantn: Done.

bodo/pandas/plan.py
ehsantn: Done.

PR #1178, opened 2026-05-19
[GPU] BSE-5428: Fix GPU Nunique

IsaacWarren:

Handle aggregations with pre_agg_fns like nunique per column in separate aggregation steps, then combine with the remaining non-pre-agg columns. 

## Testing strategy

CI

## Checklist


scott-routledge2 (approved):

bodo/libs/streaming/cuda_groupby.cpp
```diff
@@ -47,13 +47,140 @@ std::unique_ptr<cudf::table> CudaGroupbyState::do_groupby(
...
+                }
+                keys_stored = true;
+            }
+            // Store the deduplicated result column
+            for (auto& agg_result : result.second) {
+                for (auto& c : agg_result.results) {
+                    if (post_agg_fns[i] != nullptr) {
+                        c = post_agg_fns[i](c->view(), stream);
+                    }
+                    deduped_cols.push_back(std::move(c));
```
DrTodd13, line 116: I guess my fear here is that subsequent dedup groupbys might not output keys in the same order or drop some keys.

bodo/libs/streaming/cuda_groupby.cpp
```diff
@@ -82,7 +209,11 @@ std::unique_ptr<cudf::table> CudaGroupbyState::do_groupby(
         aggregation_requests[i].values = col_to_use;
     }
 
-    // Run the groupby
+    // Run the groupby.
+    // If any pre_agg_table_fns were processed, accumulation happened
```
DrTodd13, line 213: If any pre-agg happened then isn't it returning there and so won't get here?  I'm concerned about the complexity of this.  Could we have a simpler loop that does aggregations one at a time and if it needs a pre-agg runs it there and then runs the singular aggregation on the keys plus result of the pre-agg.  I guess the pre-agg stuff would need to be changed to operate on an extracted columns instead of assuming a table but that isn't hard.  Then, you stich together the keys plus the individual aggregation results into one table at the end?    Again, I'd have concerns with guaranteeing that keys appear in the same order for each aggregation but if that is guaranteed then would this be simpler?

DrTodd13 (approved):
Not sure on this one.  Seems to be getting pretty complicated and there ought to be a simpler way.

PR #1179, opened 2026-05-20
[GPU] workers > gpus, groupby fallback

DrTodd13:

Resolves BSE-5430.
If groupby falls back on bodo dataframe then convert dataframe to pandas and run groupby on that rather than letting pandas work on the bodo dataframe piecemeal.

## Testing strategy

run_ci

## User facing changes

None

## Checklist


bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -4513,3 +4513,24 @@ def test_join_filter_pushdown_aggregate_split_keys():
         sort_output=True,
         reset_index=True,
     )
+
+
+@pytest.mark.gpu
```
scott-routledge2: I think this test is a duplicate of `test_groupby_fallback`

bodo/pandas/utils.py
scott-routledge2: I believe we also need to convert `self._obj` back to pandas here for cases where the groupby succeeds but some other method e.g. `apply()` falls back? Thinking specifically of `bodo/tests/test_df_lib/test_groupby_udf.py::test_apply_fallback`, which was also failing in the num workers > num gpus case.

scott-routledge2 (approved):
Thanks @DrTodd13 , Can we unskip `bodo/tests/test_df_lib/test_groupby_udf.py::test_apply_fallback` and `test_end_to_end.py::test_groupby_fallback` now and make sure they are passing?

bodo/tests/test_df_lib/test_end_to_end.py
DrTodd13: Fixed. thanks

bodo/pandas/utils.py
DrTodd13: I tried to be minimalistic in this change.  We think there could be broader issues but unsure how best to address them.  test_apply_fallback works with this PR so I'm inclined not to make changes to the part you mention unless we have a proven case of it failing.

ehsantn (approved):
Thanks, Todd.

bodo/pandas/utils.py
DrTodd13: Finally found a configuration where test_apply_fallback did fail so fixed to convert _obj to pandas and run groupby on that.

PR #1180, opened 2026-05-21
[GPU] Only reduce on ranks with data.

DrTodd13:

GPU reduce will only reduce on ranks with data.

## Testing strategy

run_ci

## User facing changes

None

## Checklist


ehsantn (approved):
Thanks, Todd. There are changes from the other PR that need removed here.

DrTodd13:
> Thanks, Todd. There are changes from the other PR that need removed here.

Yes, I'll merge the other one first and then update this branch.

bodo/pandas/physical/gpu_reduce.cpp
```diff
@@ -152,26 +152,34 @@ void GPUReductionFunction::Finalize(MPI_Comm comm) {
...
+            // NOTE: OpenMPI collectives are not CUDA-aware
+            CHECK_MPI(
+                MPI_Allreduce(MPI_IN_PLACE, result_ptr, 1, mpi_dtype,
+                              this->mpi_reduce_op, has_data_comm),
+                "GPUReductionFunction::Finalize: MPI error on MPI_Allreduce:");
+
+            // Copy the reduced CPU result back to cudf scalar
+            std::shared_ptr<arrow::Scalar> arrow_scalar =
+                arrow_table->column(0)->GetScalar(0).ValueOrDie();
+            this->results[i] = arrow_scalar_to_cudf(arrow_scalar);
```
scott-routledge2, line 181: What happens if results is not populated for a given rank, it will just contain nullptrs?

scott-routledge2 (approved):
Thanks @DrTodd13

bodo/pandas/physical/gpu_reduce.cpp
DrTodd13: Yeah, it contains nullptr.  The ones that contain nullptr should be the non-GPU ranks which don't read from that to produce.  Might add an assert for that.

scott-routledge2: Wouldn't it also be nullptr if a GPU rank with empty data?

DrTodd13: result is unique_ptr to cudf::scalar so only the ptr can be null, the scalar can't be empty.

DrTodd13: don't think scalar NA is possible either.

PR #1182, opened 2026-05-21
Build platform package in container for compat 

scott-routledge2:

We need to build our platform package with GLIBC <= 2.34 (i.e. the same version as our conda-forge package and other linux packages) for compatibility with the new OS on our platform (EL9). This glibc was previously the version we built against before our native github runners upgraded their OS (which upgraded GLIBC to 2.38), which led to us upgrade our platform to EL10.

However, EL10 is not compatible with EFA, so we need to downgrade to EL9 on the platform, and therefore downgrade glibc in conda-platform as well. 

## Checklist


.github/workflows/_build_bodo_conda_linux_comm.yml
```diff
@@ -19,7 +24,7 @@ on:
 jobs:
   build-bodo:
     runs-on: ${{ inputs.arm && 'ubuntu-24.04-arm' || 'ubuntu-latest' }}
-    container: ${{ inputs.arm && 'condaforge/linux-anvil-aarch64:alma8' || 'condaforge/linux-anvil-x86_64:alma8' }}
+    container: ${{ inputs.arm && 'condaforge/linux-anvil-aarch64:alma9' || 'condaforge/linux-anvil-x86_64:alma9' }}
```
scott-routledge2: This is the same image used in our conda-forge workflow:
https://github.com/conda-forge/bodo-feedstock/blob/501008865d39410e04fa2000a8506d9e0b05d951/.github/workflows/conda-build.yml#L29 

Though if we still want to keep the older version on our channel then we can revert this change.

buildscripts/bodo/conda-recipe/meta.yaml
```diff
@@ -33,7 +33,8 @@ requirements:
     - {{ compiler('cuda') }}  # [gpu]
 
   host:
-    - python
+    - python {{ PY_VER }}
```
scott-routledge2, line 36: Avoid "no gil python" (314t), otherwise our packages won't be compatible with conda, which will cause issues for the platform.

See: https://github.com/conda-forge/numba-feedstock/blob/286a0d5dc9f925532a18de1a48922240c22228f1/.ci_support/linux_64_python3.14.____cp314t.yaml#L26

DrTodd13 (approved):

buildscripts/bodo/conda-recipe/meta.yaml
ehsantn: Let's add this comment to the code. Do we need to update our conda-forge too?

ehsantn (approved):

buildscripts/bodo/conda-recipe/meta.yaml
scott-routledge2: No, we have the same thing in our feedstock repo CI except we don't have a no-GIL build yet:
https://github.com/conda-forge/bodo-feedstock/blob/96f11f3e743897af6e4c8a487bdac1925638e463/.ci_support/linux_64_libarrow23.0python3.14.____cp314.yaml#L34

PR #1183, opened 2026-05-21
Support BodoDataFrame.get() without fallback message [GPU]

dmsa-or:
`get()` is now a native `BodoDataFrame` method. This reuses the pandas implementation which is fine for us because it is basically just a wrapper for `__getitem__`.

## Testing strategy
Test cases for different input combinations, some that should return the default and some that shouldn't.

## User facing changes

`BodoDataFrame.get()` is officially supported, and there is no fallback message suggesting there might be a performance penalty when there isn't.

## Checklist

### DIFF
--- a/bodo/pandas/frame.py
+++ b/bodo/pandas/frame.py
@@ -1118,6 +1118,10 @@ class BodoDataFrame(pd.DataFrame, BodoLazyWrapper):
             )
         return self.__getitem__(columns_to_use)
 
+    @check_args_fallback("none")
+    def get(self, key, default=None):
+        return pd.DataFrame.get(self, key, default)
+
     @check_args_fallback("all")
     def __getitem__(self, key):
         """Called when df[key] is used."""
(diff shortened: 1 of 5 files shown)


DrTodd13 (approved):

bodo/pandas/frame.py
```diff
@@ -1119,6 +1118,10 @@ def drop(
             )
         return self.__getitem__(columns_to_use)
 
+    @check_args_fallback("none")
+    def get(self, key, default=None):
+        return pd.DataFrame.get(self, key, default)
```
ehsantn: This is reusing the Pandas implementation but their code is very simple. I'd copy their code here for clarity and add a link for reference.
https://github.com/pandas-dev/pandas/blob/366ccdfcd8ed1e5543bfb6d4ee0c9bc519898670/pandas/core/generic.py#L4427

bodo/pandas/series.py
```diff
@@ -960,6 +960,10 @@ def sort_values(
             ),
         )
 
+    @check_args_fallback(unsupported="none")
+    def get(self, key, default=None):
+        return pd.Series.get(self, key, default)
```
ehsantn: `key` here is the Index value for the row to retrieve (like a map key). We don't support this in getitem so adding `get` here doesn't really add value and may make the fallback warnings more confusing.

ehsantn (commented):
See comments.

ehsantn (approved):
Thank you, Daniel.

PR #1186, opened 2026-05-23
Support BodoDataFrame.filter() [GPU]

dmsa-or:

`filter()` is implemented on `BodoDataFrame`. `BodoDataFrame.filter()` only supports the column axis for now, due to row index challenges.

## Testing strategy

Tests are added to verify both proper and improper usage of `filter()`. One test case is skipped due to an existing inconsistency in behavior versus pandas (maybe worth taking a look at).

## User facing changes

The `filter()` chaining method is now natively supported (for `BodoDataFrame`).

## Checklist

### DIFF
--- a/bodo/pandas/frame.py
+++ b/bodo/pandas/frame.py
@@ -1477,6 +1477,45 @@ class BodoDataFrame(pd.DataFrame, BodoLazyWrapper):
             self._plan, empty_series, "apply", (func,), apply_kwargs
         )
 
+    @check_args_fallback(supported=["items", "like", "regex"])
+    def filter(
+        self,
+        items=None,
+        like: str | None = None,
+        regex: str | None = None,
+        axis: Axis | None = None,
+    ) -> BodoDataFrame | None:
+        if sum(x is not None for x in [items, like, regex]) > 1:
+            raise TypeError(
+                "Keyword arguments `items`, `like`, or `regex` are mutually exclusive"
+            )
+
+        if items is not None:
+            items = pd.Index(items).intersection(self.columns).to_list()
+        elif like:
+
+            def search(label):
+                if not isinstance(label, str):
+                    label = str(label)
+                return like in label
+
+            items = list(np.array(self.columns)[self.columns.map(search)])
+        elif regex:
+            import re
+
+            matcher = re.compile(regex)
+
+            def search(label):
+                if not isinstance(label, str):
+                    label = str(label)
+                return matcher.search(label) is not None
+
+            items = list(np.array(self.columns)[self.columns.map(search)])
+        else:
+            raise TypeError("Must pass either `items`, `like`, or `regex`")
+
+        return self.__getitem__(items)
+
     @check_args_fallback(supported=["by", "ascending", "na_position", "kind"])
     def sort_values(
         self,
--- a/bodo/pandas/frame.py
+++ b/bodo/pandas/frame.py
@@ -1493,24 +1493,14 @@ class BodoDataFrame(pd.DataFrame, BodoLazyWrapper):
         if items is not None:
             items = pd.Index(items).intersection(self.columns).to_list()
         elif like:
-
-            def search(label):
-                if not isinstance(label, str):
-                    label = str(label)
-                return like in label
-
-            items = list(np.array(self.columns)[self.columns.map(search)])
+            items = [label for label in self.columns if like in str(label)]
         elif regex:
             import re
 
             matcher = re.compile(regex)
-
-            def search(label):
-                if not isinstance(label, str):
-                    label = str(label)
-                return matcher.search(label) is not None
-
-            items = list(np.array(self.columns)[self.columns.map(search)])
+            items = [
+                label for label in self.columns if matcher.search(label) is not None
+            ]
         else:
             raise TypeError("Must pass either `items`, `like`, or `regex`")
 
--- a/bodo/pandas/frame.py
+++ b/bodo/pandas/frame.py
@@ -1477,14 +1477,19 @@ class BodoDataFrame(pd.DataFrame, BodoLazyWrapper):
             self._plan, empty_series, "apply", (func,), apply_kwargs
         )
 
-    @check_args_fallback(supported=["items", "like", "regex"])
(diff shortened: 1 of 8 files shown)


dmsa-or:
This is a similar situation to `BodoSeries.get()` in that `BodoSeries.filter()` seems to just trigger a fallback. I can remove it if necessary.

bodo/pandas/frame.py
```diff
@@ -1477,6 +1477,45 @@ def apply(
             self._plan, empty_series, "apply", (func,), apply_kwargs
         )
 
+    @check_args_fallback(supported=["items", "like", "regex"])
```
DrTodd13: Say we support axis but throw exception if the row axis is selected.

bodo/pandas/frame.py
```diff
@@ -1477,6 +1477,45 @@ def apply(
...
+        if items is not None:
+            items = pd.Index(items).intersection(self.columns).to_list()
+        elif like:
+
+            def search(label):
+                if not isinstance(label, str):
+                    label = str(label)
+                return like in label
+
+            items = list(np.array(self.columns)[self.columns.map(search)])
```
DrTodd13: items = [x for x in self.columns if like in str(x)]

Is this equivalent and therefore better due to being shorter?

bodo/pandas/frame.py
```diff
@@ -1477,6 +1477,45 @@ def apply(
...
+            import re
+
+            matcher = re.compile(regex)
+
+            def search(label):
+                if not isinstance(label, str):
+                    label = str(label)
+                return matcher.search(label) is not None
+
+            items = list(np.array(self.columns)[self.columns.map(search)])
```
DrTodd13: Can do the same here as above but substituting regex for like?

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -667,6 +667,60 @@ def test_filter_datetime(datapath, op):
...
+            {"items": []},
+            None,
+            0,
+            marks=pytest.mark.skip(reason="Existing minor Bodo difference"),
+        ),
+        pytest.param(
+            {"items": [], "axis": 1},
+            None,
+            1,
+            marks=pytest.mark.skip(reason="Currently crashes"),
```
DrTodd13: This is a fallback because we don't support axis?  Anyway, this should be supported by filter because 1 is columns right?  I guess we test empty rows plenty of places but this is testing empty columns, right?

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -667,6 +667,60 @@ def test_filter_datetime(datapath, op):
...
+        pytest.param({"items": [], "like": "A"}, TypeError, 0),
+        pytest.param({"items": ["A_3", "2-A", "2-A", "not_a_column"]}, None, 0),
+        pytest.param({"items": {"1AA", "B-", "A_3"}}, None, 0),
+        pytest.param({"like": "A"}, None, 0),
+        pytest.param({"regex": ".*[-4].*"}, None, 0),
+        pytest.param(
+            {"items": []},
+            None,
+            0,
+            marks=pytest.mark.skip(reason="Existing minor Bodo difference"),
```
DrTodd13: How are there differences if there are no columns?

DrTodd13 (changes requested):

bodo/pandas/frame.py
dmsa-or: I think so, although my version above is more similar to what the Pandas implementation does.

bodo/tests/test_df_lib/test_end_to_end.py
dmsa-or: This is BSE-5438. It only falls back at the moment because of the axis argument. If we detected `axis=1` and didn't fall back, this test would not be a problem. However, it might hide a deeper issue that is causing the `filter` fallback to crash in this case.

bodo/pandas/frame.py
dmsa-or: Can we still fall back if the row axis is selected, or is it too late if it passed `check_args_fallback`? Would I have to manually call the Pandas version?

bodo/tests/test_df_lib/test_end_to_end.py
dmsa-or: Pandas keeps the indices and Bodo doesn't. The dataframes are not considered equal by `_test_equal`.

bodo/pandas/frame.py
DrTodd13: Look at other cases that does this sort of thing and if you throw the right exception then it does fallback still.

bodo/tests/test_df_lib/test_end_to_end.py
DrTodd13: Right.  I remember now.  Do you think keeping the indices is easily doable?

dmsa-or: I have no clue why the indices are dropped, so not sure how easy that is to fix. Would need to do more digging.

DrTodd13 (approved):
Thanks.

bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -667,6 +667,58 @@ def test_filter_datetime(datapath, op):
...
+    """Basic test for DataFrame.filter()"""
+    with assert_executed_plan_count(epc):
+        df = pd.DataFrame(
+            {
+                "A1": [1, 2, 3, 4, 5],
+                "1AA": [10, 9, 8, 7, 6],
+                "2-A": [5, 5.9, -0.13, 1, 1],
+                "B-": [1, 10, 2, 9, 3],
+                "A_3": [8.2, 1.43, 7.3, 9.7, 8.21],
+                4: [2.5, 3, 3.5, 4, 4.02],
```
ehsantn, line 703: I think we don't fully support non-string column names (since we use Arrow among other issues), but it may be ok in this limited case.

ehsantn (approved):
Looks good to me. Thanks.

PR #1199, opened 2026-06-01
BSE-5443: Datetime offset

IsaacWarren:
Adds support for sql intervals, note we don't support nanosecond precision because duckdb intervals don't, snowflake doesn't either so it should be ok.

## Testing strategy
Marking existing tests bodosql_cpp

## User facing changes
Interval expression support

## Checklist

### DIFF
--- a/BodoSQL/bodosql/plan_conversion.py
+++ b/BodoSQL/bodosql/plan_conversion.py
@@ -738,13 +738,15 @@ def java_literal_to_python_literal(ctx, java_literal, input_plan):
         if _is_year_month_interval(lit_type_name):
             # getValue() returns a BigDecimal representing total months
             months = int(java_literal.getValue())
-            val = pa.scalar((months, 0, 0), type=interval_type)
+            # cloudpickle can't serialize pyarrow month_day_nano_interval objects so we pass the values as integers and construct the interval in the C++ backend
+            val = ("MonthDayNanoInterval", months, 0, 0)
         else:
             # Day/second subtypes: getValue2() returns a BigDecimal (Py4J
             # converts to decimal.Decimal) representing milliseconds.
             millis = float(str(java_literal.getValue2()))
             nanos = int(millis * 1_000_000)
-            val = pa.scalar((0, 0, nanos), type=interval_type)
+            # cloudpickle can't serialize pyarrow month_day_nano_interval objects so we pass the values as integers and construct the interval in the C++ backend
+            val = ("MonthDayNanoInterval", 0, 0, nanos)
         return ConstantExpression(dummy_empty_data, input_plan, val)
 
     if (
--- a/BodoSQL/bodosql/plan_conversion.py
+++ b/BodoSQL/bodosql/plan_conversion.py
@@ -737,8 +737,7 @@ def java_literal_to_python_literal(ctx, java_literal, input_plan):
         return ConstantExpression(dummy_empty_data, input_plan, val)
 
     if _is_interval_type(lit_type_name):
-        interval_type = pa.month_day_nano_interval()
-        dummy_empty_data = pd.Series(dtype=pd.ArrowDtype(interval_type))
+        dummy_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.duration("ns")))
         if _is_year_month_interval(lit_type_name):
             # getValue() returns a BigDecimal representing total months
             months = int(java_literal.getValue())
@@ -1082,7 +1081,7 @@ def sql_type_to_pa_type(ctx, sql_type_name):
         tz = ctx.default_tz if ctx.default_tz is not None else "UTC"
         return pa.timestamp("ns", tz=tz)
     if _is_interval_type(sql_type_name):
-        return pa.month_day_nano_interval()
+        return pa.duration("ns")
     if sql_type_name.equals(SqlTypeName.BOOLEAN):
         return pa.bool_()
     if sql_type_name.equals(SqlTypeName.CHAR):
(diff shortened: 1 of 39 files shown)


scott-routledge2 (approved):
Thanks @IsaacWarren

bodo/pandas/_plan.cpp
```diff
@@ -158,15 +158,20 @@ duckdb::unique_ptr<duckdb::Expression> make_const_timestamp_ns_expr(
...
+    int64_t micros = val / 1000;
+    return duckdb::make_uniq<duckdb::BoundConstantExpression>(
+        duckdb::Value::INTERVAL(duckdb::Interval::FromMicro(micros)));
+}
+
+duckdb::unique_ptr<duckdb::Expression> make_const_date_offset_expr(
+    int32_t months, int64_t nanos) {
+    duckdb::interval_t interval_val;
+    interval_val.months = months;
+    interval_val.days = 0;
```
ehsantn: We don't have use cases where we need to pass days too?

bodo/pandas/plan_optimizer.pyx
```diff
@@ -742,6 +743,8 @@ cdef unique_ptr[CExpression] make_const_expr(object const_schema, val):
         return move(make_const_timestamp_ns_expr(val.value))
     elif isinstance(val, pd.Timedelta):
         return move(make_const_timedelta_ns_expr(val.value))
+    elif isinstance(val, pd.DateOffset):
+        return move(make_const_date_offset_expr(val.months, val.nanos))
```
ehsantn: `DateOffset` doesn't seem to normalize to months and nanos. I suggest using Arrows `month_day_nano_interval` scalars to match Calcite.
```python
In [7]: d = pd.DateOffset(years=3)

In [8]: d.months
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[8], line 1
----> 1 d.months

AttributeError: 'DateOffset' object has no attribute 'months'

In [9]: d.nanos
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[9], line 1
----> 1 d.nanos

File pandas/_libs/tslibs/offsets.pyx:963, in pandas._libs.tslibs.offsets.BaseOffset.nanos.__get__()
--> 963 'Could not get source, probably due dynamically evaluated source code.'

ValueError: <DateOffset: years=3> is a non-fixed frequency
```

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -702,10 +732,21 @@ def java_literal_to_python_literal(ctx, java_literal, input_plan):
         val = pd.Timestamp(java_literal.getValue2(), unit="ms")
         return ConstantExpression(dummy_empty_data, input_plan, val)
 
-    if lit_type_name.equals(SqlTypeName.INTERVAL_DAY_SECOND):
+    if _is_year_month_interval(lit_type_name):
+        # getValue() returns a BigDecimal representing total months
+        months = int(java_literal.getValue())
+        dummy_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.duration("ns")))
```
ehsantn: We probably need to support Arrow's calendar-aware interval data type:
https://arrow.apache.org/docs/python/generated/pyarrow.month_day_nano_interval.html

ehsantn (commented):
See comments. We need to handle relative calendar-aware offsets vs fixed durations properly. For example, 3 months added to Jan 31 should land on Apr 30 and not just 90 days later.

bodo/pandas/plan_optimizer.pyx
```diff
@@ -742,6 +743,19 @@ cdef unique_ptr[CExpression] make_const_expr(object const_schema, val):
...
+        nanos = 0
+        try:
+            nanos = val.nanos
+        except (AttributeError, ValueError):
+            raise NotImplementedError("Only DateOffsets with a fixed frequency are supported as constants, if a variable frequency is needed use pyarrow's MonthDayNanoInterval type instead.")
+        return move(make_const_date_offset_expr(0, 0, nanos))
+    elif isinstance(val, pa.MonthDayNanoIntervalScalar):
+        py_val = val.as_py()
+        return move(make_const_date_offset_expr(
+            py_val.months, py_val.days, py_val.nanoseconds))
```
ehsantn, line 756: These two branches are unused right now?

bodo/pandas/_util.cpp
```diff
@@ -105,12 +105,11 @@ extractValue(const duckdb::Value &value) {
             return arrow::MakeScalar(date_type, extracted.days).ValueOrDie();
         } break;
         case duckdb::LogicalTypeId::INTERVAL: {
-            auto interval_type = arrow::duration(arrow::TimeUnit::NANO);
             duckdb::interval_t extracted = value.GetValue<duckdb::interval_t>();
-            return arrow::MakeScalar(
-                       interval_type,
-                       duckdb::Interval::GetMicro(extracted) * 1000)
-                .ValueOrDie();
+            // Convert to nanoseconds total, dropping month/day information
```
ehsantn, line 109: Can we use Arrows calendar-aware type here if the DuckDB value had month/day?

ehsantn (approved):
Looks good to me.

bodo/pandas/_util.cpp
IsaacWarren: I tried this originally but if the result of the expression is the calendar aware type we don't have a way to store since we don't have an equivalent Bodo array and my understanding of the issue was that arrays were out of scope

ehsantn: The scalar becomes an array down the line? Should we throw errors for month/day case to avoid wrong results?

IsaacWarren: We could, but previously in bodosql we also just cast to duration https://github.com/bodo-ai/Bodo/blob/b1668cf6803a540a51588bd612311ef459a38f45/BodoSQL/bodosql/tests/test_literals.py#L213 if you look at the expected results

PR #1201, opened 2026-06-02
Add suffix to query profile directory to avoid collisions

scott-routledge2:

When queries are executed in rapid succession, such as a simple test script, query profile directories with the same path can be created leading to hangs. 

This PR avoids this by adding a suffix to the end of subsequent query profile directories with the same name. 

This directory name will be generated by rank 0 and broadcasted out.

## Testing strategy
CI, 
Used small test script on the platform:
``` py
import numpy as np

import bodo.pandas as pd

df = pd.DataFrame({"A": np.arange(1000), "B": np.random.rand(1000)})
df2 = pd.DataFrame({"A": np.arange(900, 10000), "C": np.random.rand(9100)})

print(df.merge(df2))
```

Which hangs on main but runs on my branch. 

## Checklist

### DIFF
--- a/bodo/libs/_query_profile_collector.cpp
+++ b/bodo/libs/_query_profile_collector.cpp
@@ -82,15 +82,30 @@ void QueryProfileCollector::Init() {
             std::chrono::duration_cast<std::chrono::minutes>(seconds - hour);
         auto second = std::chrono::duration_cast<std::chrono::seconds>(
             seconds - hour - minute);
-        output_dir = fmt::format("{}/run_{}{:02}{:02}_{:02}{:02}{:02}",
-                                 parent_dir, year, month, day, hour.count(),
-                                 minute.count(), second.count());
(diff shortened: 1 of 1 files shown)


bodo/libs/_query_profile_collector.cpp
```diff
@@ -82,15 +82,30 @@ void QueryProfileCollector::Init() {
...
+            hour.count(), minute.count(), second.count());
+
+        int run_suffix = 0;
+        while (true) {
+            std::string candidate =
+                run_suffix == 0
+                    ? output_dir_base
+                    : output_dir_base + "_" + std::to_string(run_suffix);
+
+            if (mkdir(candidate.c_str(), 0700) == 0) {
```
scott-routledge2, line 96: Rank 0 creates this directory and then broadcasts name to other ranks

IsaacWarren (approved):
Thanks Scott

ehsantn (approved):

PR #1211, opened 2026-06-08
Update PeriodIndex from_ordinals comment

yu-liu24:
Updates a stale PeriodIndex comment to match the current PeriodIndex.from_ordinals call.\n\nThe implementation already calls PeriodIndex.from_ordinals with the ordinals keyword, but the nearby comment still referenced the deprecated PeriodIndex ordinal keyword. This keeps the comment aligned with the current pandas API recommendation.\n\nValidation: comment-only change, inspected diff.


ehsantn (approved):
Thanks for your contribution!

PR #1213, opened 2026-06-08
Fix parsing for unary operators in JIT dataframe query expression

dmsa-or:

This PR fixes a correctness issue when unary operators (such as `not`) are used in the `query()` expression string on the JIT side. The existing code seems to assume that there will be 2 or more operands; if there is just one operand it drops the operator.

## Testing strategy

Added a test to cover this case.

## Checklist

### DIFF
--- a/bodo/tests/test_df_query.py
+++ b/bodo/tests/test_df_query.py
@@ -17,6 +17,7 @@ pytestmark = pytest_pandas
     [
         "`B B` > @a + 1 & 5 > index > 1",
         pytest.param("(A == @a) | (C == 'AA')", marks=pytest.mark.slow),
+        pytest.param("not A < 5", marks=pytest.mark.slow),
         pytest.param("C in ['AA', 'C']", marks=pytest.mark.slow),
         pytest.param("C not in ['AA', 'C']", marks=pytest.mark.slow),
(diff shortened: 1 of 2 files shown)


DrTodd13 (approved):

bodo/tests/test_df_query.py
```diff
@@ -17,6 +17,7 @@
     [
         "`B B` > @a + 1 & 5 > index > 1",
         pytest.param("(A == @a) | (C == 'AA')", marks=pytest.mark.slow),
+        pytest.param("not A < 5", marks=pytest.mark.slow),
```
ehsantn, line 20: What is the operator precedence here?

ehsantn (approved):
Looks good to me. Thanks, Daniel.

bodo/tests/test_df_query.py
dmsa-or: I can't find a source exactly stating this, but it seems to act similarly to "not" used in Python, e.g. "not A < 5 and not B > 3" would be parsed as "not (A < 5) and not (B > 3)". On the other hand, if ~ was used instead here, you would need to add parentheses to avoid it being interpreted as "(~A) < 5".

PR #1217, opened 2026-06-09
Support BodoDataFrame.query() [GPU]

dmsa-or:

`query` is implemented natively on `BodoDataFrame`. We reuse the query plans from `__getitem__` by parsing the query expression string, converting to bracket syntax, and calling eval(). Much of the parsing logic was borrowed from the JIT side.

## Testing strategy

New unit test cases were added for `query()`.

## User facing changes

Dataframe `query()` is now supported aside from the arguments `parser`, `engine`, and `inplace`.

## Checklist


bodo/pandas/frame.py
```diff
@@ -1511,6 +1512,355 @@ def filter(
...
+            for series_accessor in ["str", "dt", "cat", "sparse"]:
+                parsed_expr_str = parsed_expr_str.replace(
+                    f"{clean_column_name}.{series_accessor}",
+                    f"self['{column_name}'].{series_accessor}",
+                )
+
+        try:
+            return eval(f"self.__getitem__({parsed_expr_str})")
+        except Exception as e:
+            raise BodoLibNotImplementedException("Unsupported query usage: " + str(e))
```
IsaacWarren, line 1688: If we get to this point the query should be supported, right? I'm just worried a user could write an invalid query and this swallows it

bodo/tests/test_df_lib/test_end_to_end.py
IsaacWarren, line 1: Do any of these test fallback?

IsaacWarren (commented):
A couple questions, thanks @dmsa-or!

bodo/pandas/frame.py
```diff
@@ -1511,6 +1512,355 @@ def filter(
...
+            cleaned_columns,
+            join_cleaned_cols=(),
+        ):
+            """Parses expression string for BodoDataFrame.query() call handling.
+            Patches Pandas query expr parsing code to avoid evaluating values during parsing.
+
+            join_cleaned_cols are used by general join condition to detect
+            non-valid identifiers.
+
+            Tweaked from the original bodo.hiframes.dataframe_impl._parse_query_expr() for conversion to bracket syntax
```
IsaacWarren: What are the differences between this and bodo.hiframes.dataframe_impl._parse_query_expr(). Can we refactor any of that functionality into standalone functions and call in both places to avoid code duplication?

bodo/pandas/frame.py
dmsa-or: Most likely, yes. I pretty much added this just in case. The test case for "A >= @comparison_list" is one example of a query that works in pandas but not for us because `__getitem__` doesn't accept elementwise comparison like that.

bodo/tests/test_df_lib/test_end_to_end.py
dmsa-or: Yes, the index test falls back, as well as the one mentioned above since we catch the error and raise `BodoLibNotImplementedException`.

bodo/pandas/frame.py
dmsa-or: I have changed `math__str__`, `op__str__`, and the `except pd.errors.UndefinedVariableError` case.

ehsantn (approved):
Looks good to me overall. Thanks, Daniel. It would be great if we could avoid code duplication a bit if possible.

IsaacWarren (approved):
Thanks @dmsa-or

PR #1219, opened 2026-06-10
SEARCH and new literal_agg implementation.

DrTodd13:

Implement most common case for SEARCH.
New implementation for convert of calcite literal aggregations.
Support direct creation of LogicalTopN.
Mark currently passing tpch bodosql tests.
Support Calcite CachedSubPlans.

## Testing strategy

run_ci

## User facing changes

None

## Checklist


BodoSQL/bodosql/plan_conversion.py
```diff
@@ -900,6 +904,108 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+            sarg = search_expr.value
+            sarg_rangeSet = sarg.getClass().getDeclaredField("rangeSet").get(sarg)
+            ranges_collection = sarg_rangeSet.asRanges()
+            search_options = []
+            try:
+                it = ranges_collection.iterator()
+                while it.hasNext():
+                    r = it.next()
+                    search_options.append(r)
+            except Exception:
```
IsaacWarren: Maybe cleaner to except whatever the specific index error is instead

IsaacWarren (approved):
Thanks @DrTodd13

BodoSQL/bodosql/plan_conversion.py
DrTodd13: AI suggested that exception region even though I expect it will always go through the try portion so I just simplified it and got rid of the try/except.

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -900,6 +904,102 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+    if operator_class_name == "SqlSearchOperator":
+        operands = java_call.getOperands()
+        op_exprs = [java_expr_to_python_expr(ctx, o, input_plan) for o in operands]
+        func_name = op.getName().upper()
+
+        if func_name == "SEARCH" and len(op_exprs) == 2:
+            src = op_exprs[0]
+            search_expr = op_exprs[1]
+            bool_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.bool_()))
+            sarg = search_expr.value
```
ehsantn, line 918: What is the data type of `search_expr`? Does this need an assertion?

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -900,6 +904,102 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+            sarg_rangeSet = sarg.getClass().getDeclaredField("rangeSet").get(sarg)
+            ranges_collection = sarg_rangeSet.asRanges()
+            search_options = []
+            it = ranges_collection.iterator()
+            while it.hasNext():
+                r = it.next()
+                search_options.append(r)
+
+            def range_type_to_python(x):
```
ehsantn, line 937: This doesn't look robust. What is the data type of `x`? Why can we convert decimal to int without losing fractions?

bodo/pandas/_physical_conv.cpp
```diff
@@ -615,6 +615,10 @@ duckdb::vector<duckdb::idx_t> gen_split_filter_projection_map(
  */
 std::unique_ptr<duckdb::LogicalOperator> SplitNonEquiFromComparisonJoin(
     duckdb::LogicalComparisonJoin& comp_join) {
+    if (comp_join.join_type != duckdb::JoinType::INNER) {
```
ehsantn, line 621: Add a comment? For outer non-equi joins, we will use whatever support we have in `PhysicalJoin`?

ehsantn (approved):
Thanks, Todd. See comments.

BodoSQL/bodosql/plan_conversion.py
DrTodd13: Added comments and assertions.

BodoSQL/bodosql/plan_conversion.py
DrTodd13: fixed.

bodo/pandas/_physical_conv.cpp
DrTodd13: The actual support is in streaming/_join but yes it will do what it was doing before.  Now that I think about it, we should make a task soon for fixing this issue at its core rather than adding hacks.

PR #1231, opened 2026-06-16
Support more string operations in BodoSQL C++ backend (second batch)

dmsa-or:

Added conversions to Arrow compute functions for a few more SQL string operations and marked the corresponding tests with `pytest.mark.bodosql_cpp`.
Added logic on the C++ side to redirect to using the binary versions of Arrow string compute functions if the input is binary.

## Testing strategy

CI

## Checklist


bodo/pandas/physical/expression.h
```diff
@@ -1431,17 +1455,165 @@ class PhysicalArrowExpression : public PhysicalExpression {
...
+                    std::make_shared<ScalarExprResult>(
+                        ConvertDatumToArrayInfo(empty_string));
+                std::shared_ptr<ExprResult> captured_field_expr_result =
+                    std::make_shared<ArrayExprResult>(
+                        ConvertDatumToArrayInfo(arrow::Datum(captured_field)),
+                        "regexp_substr_captured_field");
+
+                captured_field_datum = do_arrow_compute_multi_input_datum(
+                    {captured_field_expr_result, empty_string_expr_result},
+                    "nullif");
```
dmsa-or: It's kind of ugly having to do this. In general, I think we should have a common way of operating on `ExprResult`s and `arrow::Datum`s and automatically handling the necessary conversions and return types.

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -1225,23 +1346,91 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+                occurences_replaced_expr = ArrowScalarFuncExpression(
+                    src.empty_data,
+                    [without_start_expr],
+                    "replace_substring",
+                    (
+                        match_expr.value,
+                        first_char_replacement + match_expr.value[1:],
+                        occurrence_num - 1,
+                    ),
+                )
```
IsaacWarren: I'm not sure this is correct in all cases, if we're looking for the second occurrence of "-\_" in "--\_-\_" then first_char_replacement is "\_" so occurrences replaced becomes "-\_-\_" and we match the first case

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -1123,20 +1123,141 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+            # Chop off the start so that searching begins after the provided position
+            without_start_expr = ArrowScalarFuncExpression(
+                src.empty_data,
+                [src],
+                "utf8_slice_codeunits",
+                (start, None, 1),
+            )
+
+            # Remove earlier occurrences so that extract_regex can find the correct occurrence/substring matching the regexp
+            # TODO: How should this behave for overlapping occurrences?
```
IsaacWarren: What do we do in the jit kernel/do Snowflake's docs say?

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -1225,23 +1346,91 @@ def java_call_to_python_call(ctx, java_call, input_plan):
                 "utf8_length",
                 (),
             )
-        elif func_name == "INSTR" and len(op_exprs) == 2:
+        elif (func_name == "INSTR" and len(op_exprs) in (2, 3, 4)) or (
+            func_name == "CHARINDEX" and len(op_exprs) in (2, 3)
+        ):
+            # TODO: Investigate what the proper outputs should be in various exceptional conditions
```
IsaacWarren: Can we throw errors in those exceptional cases for now?

IsaacWarren (commented):
Thanks @dmsa-or

BodoSQL/bodosql/plan_conversion.py
dmsa-or: Surprisingly, the Snowflake docs have an example that seems to illustrate that recognized occurrences are not overlapping. So the technique of removing earlier occurrences seems to be correct here.

BodoSQL/bodosql/plan_conversion.py
dmsa-or: Good point, there may be a solution but I think I'll just delete the occurrence parameter because it looks like Snowflake doesn't support it for POSITION/CHARINDEX.

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -1123,20 +1123,140 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
 
-        if func_name == "REPLACE" and len(op_exprs) == 3:
+        if func_name == "REPLACE" and len(op_exprs) in (2, 3):
             src = op_exprs[0]
             search_expr = op_exprs[1]
-            replacement_expr = op_exprs[2]
 
             ensure_type_of_expr(src, "src", str)
             ensure_arg_is_const_expr_of_type(search_expr, "search_expr", str)
-            ensure_arg_is_const_expr_of_type(replacement_expr, "replacement_expr", str)
```
ehsantn, line 1133: The arguments are not necessarily constant but ok to limit them for now. https://github.com/bodo-ai/Bodo/blob/2c5a21d02e0830c570d3355f08a1d9c721b4b544/BodoSQL/bodosql/kernels/string_array_kernels.py#L1540

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -1123,20 +1123,140 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+                    )
+            else:
+                occurrence_num = 1
+
+            if len(op_exprs) >= 5:
+                regex_params_expr = op_exprs[4]
+                ensure_arg_is_const_expr_of_type(
+                    regex_params_expr, "regex_params_expr", str
+                )
+                if "c" in regex_params_expr.value and "i" in regex_params_expr.value:
```
ehsantn: Cleaner to refactor regex processing in a separate helper function.

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -1123,20 +1123,140 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+                if group_num > 0:
+                    group_num -= 1  # Convert from 1-based to 0-based
+                regex_params = (
+                    regex_params + "e"
+                )  # 'e' is implied if group_num is passed
+            else:
+                group_num = 0
+
+            # Chop off the start so that searching begins after the provided position
+            without_start_expr = ArrowScalarFuncExpression(
```
ehsantn: For performance, we can apply this if start is not zero.

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -1225,23 +1345,71 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+            # We need to return 0 in this case
+            bool_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.bool_()))
+            negative_one_expr = ConstantExpression(int_empty_data, input_plan, -1)
+            substring_not_found = ComparisonOpExpression(
+                bool_empty_data, substring_pos_expr, negative_one_expr, operator.eq
+            )
+
+            # Add the ignored start index to the result only if substring was found
+            offset_expr = ArrowScalarFuncExpression(
+                int_empty_data, [substring_not_found], "if_else", (0, start)
```
ehsantn: Why `if_else`? Null semantics are case don't match?

bodo/pandas/physical/expression.h
```diff
@@ -1431,17 +1455,165 @@ class PhysicalArrowExpression : public PhysicalExpression {
...
+                                        : "utf8_slice_codeunits";
+
             auto [start, stop, step] = get_py_slice_args(scalar_func_data.args);
 
             arrow::compute::SliceOptions opts(start, stop, step);
-            result = do_arrow_compute_unary(res, "utf8_slice_codeunits", &opts);
-        } else if (scalar_func_data.arrow_func_name == "utf8_trim") {
+            result = do_arrow_compute_unary(res, func_name, &opts);
+        } else if (scalar_func_data.arrow_func_name == "regexp_substr") {
+            auto [pattern, regex_params, group_to_extract] =
```
ehsantn, line 1593: This argument code is long so better to move it to a helper function.

ehsantn (approved):
Looks good to me overall. See comments.

BodoSQL/bodosql/plan_conversion.py
dmsa-or: Don't they have to be scalar if we use `ArrowScalarFuncExpression`?

ehsantn: We use `ArrowScalarFuncExpression` with multiple array inputs too. Probably outdated name that needs changed.

ehsantn: Actually, "Scalar" here means "element-wise".

dmsa-or: In this particular case, it looks like Arrow's `replace_substring` only supports constant `pattern` and `replacement`: https://arrow.apache.org/docs/python/generated/pyarrow.compute.replace_substring.html

So we may need another strategy to match the JIT backend.

bodo/pandas/physical/expression.h
dmsa-or: Are you referring to the whole REGEXP_SUBSTR case or just the processing until `extract_regex` is called? Because there is quite a lot after `extract_regex` too.

BodoSQL/bodosql/plan_conversion.py
ehsantn: If Arrow is limited then it's ok for now.

bodo/pandas/physical/expression.h
ehsantn: I think the whole `REGEXP_SUBSTR` could be in a separate function for readability. It could even be broken into multiple functions as needed.

ehsantn (approved):
Thanks for the fixes.

PR #1232, opened 2026-06-17
[GPU] GPU backend support for bodosql.

DrTodd13:

1. Support size with no column in gpu_aggregate.
2. Support is_null and is_true in gpu_expression.
3. Ignore join filters for join that have been converted to cross-products.
4. Support date32 and date64 conversion from arrow scalar to cudf for non-null scalars.
5. Fix left join problem when build table is empty.

## Testing strategy

run_ci

## User facing changes

None

## Checklist

### DIFF
--- a/bodo/pandas/_executor.cpp
+++ b/bodo/pandas/_executor.cpp
@@ -110,6 +110,11 @@ bool is_supported_cudf_type(const duckdb::LogicalType &type) {
     return true;
 }
 
+bool bodosql_enabled() {
+    char *env_str = std::getenv("BODOSQL_CPP_BACKEND");
+    return env_str != nullptr && std::string(env_str) != "0";
+}
+
 class DevicePlanNode {
    private:
     duckdb::LogicalOperator &op;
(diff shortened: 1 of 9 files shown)


bodo/pandas/_executor.cpp
```diff
@@ -265,14 +270,19 @@ class DevicePlanNode {
                 op.estimated_cardinality =
                     (uint64_t)(limit.limit_val.GetConstantValue());
             } else {
+                if (bodosql_enabled()) {
+                    op.has_estimated_cardinality = true;
+                    op.estimated_cardinality = rows_in;
+                } else {
```
IsaacWarren, line 281: Why do we need this?

IsaacWarren (approved):
Thanks @DrTodd13

bodo/pandas/_executor.cpp
DrTodd13: Our GPU code uses cardinality estimates to make the CPU/GPU decisions.  Well, at least that is the way it was designed.  At the moment, it doesn't really use those estimates for anything but based on the original design the code asserts that cardinality estimates are there.  From bodosql, we create a duckdb plan but don't run the optimizer on it and it is the optimizer that propagates the cardinality estimates.  Alternative is turning off cardinality checks.

bodo/pandas/_util.cpp
```diff
@@ -1614,6 +1614,18 @@ std::unique_ptr<cudf::scalar> arrow_scalar_to_cudf(
...
 
+        case arrow::Type::DATE32: {
+            auto ds = std::static_pointer_cast<arrow::Date32Scalar>(s);
+            return std::make_unique<cudf::timestamp_scalar<cudf::timestamp_D>>(
+                ds->value, true);
+        }
+
+        case arrow::Type::DATE64: {
+            auto ds = std::static_pointer_cast<arrow::Date64Scalar>(s);
+            return std::make_unique<cudf::timestamp_scalar<cudf::timestamp_ms>>(
```
ehsantn: Why ms resolution here and not days?

bodo/pandas/_executor.cpp
ehsantn: I think this is confusing. At least, let's add this comment to the code.

ehsantn (approved):
Looks good to me. Thanks, Todd.

bodo/pandas/_util.cpp
DrTodd13: I was just copying the code in the above section of that function that also uses timestamp_ms for date64.  I agree makes sense to change both to days.  Agreed?

ehsantn: Not sure, these libraries may have non-intuitive behavior based on precision so has to be tested.

PR #1245, opened 2026-06-22
Support bitwise operations in C++ backend

dmsa-or:

Bitwise operations are now supported in the C++ backend, and enabled for the scalar cases.
This PR also features:
- Support for calling Arrow's `if_else` from the Python side (used in GETBIT conversion).
- A new function `CastIntDatumsToCommonType` in `_util.h` which is called in `do_arrow_compute_case` and optionally in `do_arrow_compute_binary`. This is mainly aimed to resolve issues in Arrow's cast function when passing in a `uint64`.
- A new util function `get_scalar_py_object_as_datum` which is helpful when the input (e.g. to `if_else`) can be any arbitrary type.
- Plan conversion for MOD.

## Testing strategy

CI

## Checklist


bodo/pandas/_util.cpp
```diff
@@ -1244,6 +1488,14 @@ int64_t get_py_object_as_int64(PyObject *py_int, const char *err_context) {
     return PyLong_AsLongLong(py_int);
 }
 
+uint64_t get_py_object_as_uint64(PyObject *py_int, const char *err_context) {
+    if (!PyLong_Check(py_int)) {
+        throw std::runtime_error("Object is not a Python int: " +
+                                 std::string(err_context));
+    }
+    return PyLong_AsUnsignedLongLong(py_int);
```
IsaacWarren:
```suggestion
    uint64_t val = PyLong_AsUnsignedLongLong(py_intj);
    if (PyErr_Occurred()) {
        PyErr_Clear();
        throw std::runtime_error("Integer conversion overflowed 64-bit boundaries.");
    }
    return val;
```

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -1272,7 +1280,65 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+            if func_name == "BITAND":
+                arrow_equivalent_func = "bit_wise_and"
+            elif func_name == "BITOR":
+                arrow_equivalent_func = "bit_wise_or"
+            elif func_name == "BITXOR":
+                arrow_equivalent_func = "bit_wise_xor"
+            elif func_name == "BITSHIFTLEFT":
+                arrow_equivalent_func = "shift_left"
+                # Left shift can overflow, so promote to INT64
+                empty_data = pd.Series(dtype=pd.ArrowDtype(pa.int64()))
```
IsaacWarren: What if it's already uint64?

bodo/pandas/_util.cpp
```diff
@@ -1236,6 +1358,128 @@ void assert_py_args_is_tuple(PyObject *args, const char *err_context) {
...
+int32_t py_object_to_date(PyObject *obj) {
+    // Assume we have already verified that PyDate_Check(obj) == true
+    struct tm time_struct = {0};
+    time_struct.tm_year = PyDateTime_GET_YEAR(obj) - 1900;
+    time_struct.tm_mon = PyDateTime_GET_MONTH(obj) - 1;
+    time_struct.tm_mday = PyDateTime_GET_DAY(obj);
+    time_struct.tm_isdst = -1;
+
+    // Convert time struct down to standard epoch seconds
+    time_t epoch_seconds = mktime(&time_struct);
```
IsaacWarren: I think mktime is in localtime but arrow's date type is in UTC

bodo/pandas/_util.cpp
```diff
@@ -19,6 +19,128 @@
 #include "duckdb/planner/filter/optional_filter.hpp"
 #include "physical/expression.h"
 
+std::vector<arrow::Datum> CastIntDatumsToCommonTypeImpl(
```
IsaacWarren: Let's see if Arrow has something like this already

bodo/pandas/_util.cpp
```diff
@@ -19,6 +19,128 @@
...
+    const char *err_context, const std::vector<arrow::Datum> &datums) {
+    if (datums.empty()) {
+        return {};
+    }
+
+    auto get_bit_width = [](arrow::Type::type type_id) -> int {
```
IsaacWarren: https://arrow.apache.org/docs/cpp/api/datatype.html#_CPPv4NK5arrow8DataType9bit_widthEv

IsaacWarren (commented):
Thanks @dmsa-or!

BodoSQL/bodosql/plan_conversion.py
dmsa-or: Thinking more about this, I believe int64 empty data is correct here.

This is what Snowflake says: "_When the input expression contains an INTEGER value, returns a signed 128-bit (16-byte) integer, regardless of the size or data type of the input data value._" Also, "_If a signed INTEGER value is returned, and the value of the high bit changes (from 0 to 1, or from 1 to 0), the sign of the result is reversed. For example, BITSHIFTLEFT(1, 127) returns a negative number._"

We can't easily return int128 as far as I know, so int64 is the next best. We can replicate Snowflake semantics by doing an unsafe cast from uint64 to int64 at the end.

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -1269,7 +1277,76 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+            elif func_name == "BITSHIFTLEFT":
+                arrow_equivalent_func = "shift_left"
+                # For bitshifting, result type should be INT64 to match Snowflake and output on C++ side.
+                empty_data = int64_empty_data
+            elif func_name == "BITSHIFTRIGHT":
+                arrow_equivalent_func = "shift_right"
+                empty_data = int64_empty_data
+
+            result = ArrowScalarFuncExpression(
+                empty_data, [left_expr], arrow_equivalent_func, (right_expr.value,)
```
ehsantn: This implementation assumes the right side is always a scalar which is not the case on the Python side:
https://github.com/bodo-ai/Bodo/blob/2c5a21d02e0830c570d3355f08a1d9c721b4b544/BodoSQL/bodosql/kernels/numeric_array_kernels.py#L1322
The SQL tests are technically integration tests and unit tests for Python kernels are elsewhere like: https://github.com/bodo-ai/Bodo/blob/2c5a21d02e0830c570d3355f08a1d9c721b4b544/BodoSQL/bodosql/tests/test_kernels/test_numeric_array_kernels.py#L200

bodo/pandas/physical/expression.h
```diff
@@ -1394,11 +1398,95 @@ class PhysicalArrowExpression : public PhysicalExpression {
...
+            // between the args (as chosen by CastIntDatumsToCommonType()) needs
+            // to be signed to ensure an arithmetic right shift is performed.
+
+            // Cast the input to uint64 first (just in case).
+            // We don't want shift_left to be limited by the precision of the
+            // input (at least for INTEGER input).
+            arrow::Datum res_datum = ConvertExprResultToDatum(
+                res, scalar_func_data.arrow_func_name + " input");
+            arrow::Result<arrow::Datum> casted_res_datum;
+            if (scalar_func_data.arrow_func_name == "shift_left" ||
```
ehsantn: Let's do all casting in Python side to simplify the code.

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -1269,7 +1277,76 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+            )
+
+            bool_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.bool_()))
+            zero_expr = ConstantExpression(int64_empty_data, input_plan, 0)
+            is_bit_set = ComparisonOpExpression(
+                bool_empty_data, src_with_mask, zero_expr, operator.ne
+            )
+
+            return ArrowScalarFuncExpression(
+                int64_empty_data, [is_bit_set], "if_else", (1, 0)
```
ehsantn: Why is `if_else` necessary instead of regular case? Null semantics?

ehsantn (commented):
See comments.

bodo/pandas/physical/expression.h
```diff
@@ -1443,23 +1443,50 @@ class PhysicalArrowExpression : public PhysicalExpression {
...
+            // Note that it is important that y_datum is int64 here as opposed
+            // to uint64. When the shift_right input is signed, the common type
+            // between the args (as chosen by CastIntDatumsToCommonType()) needs
+            // to be signed to ensure an arithmetic right shift is performed.
+
+            // Cast the input to uint64 first (just in case).
+            // We don't want shift_left to be limited by the precision of the
+            // input (at least for INTEGER input).
+            arrow::Datum res_datum = ConvertExprResultToDatum(
+                res, scalar_func_data.arrow_func_name + " input");
```
IsaacWarren: We've found arrow casting to be quite slow in the past, if there's anything we can do it avoid it, even only in certain cases, that would be ideal

IsaacWarren (approved):
Thanks @dmsa-or!

BodoSQL/bodosql/plan_conversion.py
dmsa-or: I do call `ensure_arg_is_const_expr_of_type(right_expr, "right_expr", int)` above here. This restriction does make many of the tests fail, but I don't know a way around it. Is there something I'm missing?

ehsantn: You can just pass both the arguments to `ArrowScalarFuncExpression` as we do elsewhere. "Scalar" here means "element-wise" and you can have as many array argument as the actual Arrow function supports.

dmsa-or: Do you mean passing them as `input_exprs`? So that would then go through `do_arrow_compute_multi_input`, correct? I'm assuming you would have to add a special case in `do_arrow_compute_multi_input` so that the `input_exprs` are passed as separate arguments to the Arrow compute function instead of in a `std::vector` passed as a single argument to Arrow, which is the default? I feel like we need more infrastructure to make this easier, but it may be difficult in C++ to automatically unpack the `input_exprs` as positional arguments.

dmsa-or: I think I understand now. `arrow::compute::CallFunction` arguments are always passed as vector where each element corresponds to a different argument. `binary_join_element_wise` may be an exception in that regard, so it wasn't a great example for me to be referencing.

ehsantn: Yes, pass `input_exprs` to go through `do_arrow_compute_multi_input`. I don't think you need to do anything with input arguments since `CallFunction` takes a vector. Have you tried and seen any issues?

bodo/pandas/physical/expression.h
dmsa-or: I believe the casting is necessary in a lot of cases for correctness. In my updated code I've made an attempt to reduce unnecessary casts. I also added a check in `do_arrow_compute_cast` to make sure that the Datum isn't already the desired type. Not sure if Arrow checks that internally or not. This should make it cleaner to cast on the Python side without constantly verifying the input type.

BodoSQL/bodosql/tests/test_bitwise.py
```diff
@@ -36,37 +36,95 @@ def bitwise_df():
...
+                pd.DataFrame(
+                    {
+                        0: pd.Series(
+                            [100, 0, -1, -128, 8, None],
+                            dtype=pd.Int32Dtype(),
+                        )
+                    }
+                ),
+            ),
+            id="all_vector_mixed_int8",
```
dmsa-or, line 99: I added `all_vector_mixed_int8` test cases for BITAND/BITOR/BITXOR when the bit width is the same but the sign is different. The existing BITOR test cases show that when one input is signed and the other is unsigned, the result should be signed. That's the behavior I've stuck with for BITAND/BITOR/BITXOR. Note that for this test case (BITAND of int8 and uint8), DuckDB actually returned an unsigned result. I thought that contradicted the other examples, so I made the expected output of this test the signed result. I don't know who the authority is here.

BodoSQL/bodosql/plan_conversion.py
dmsa-or: That's right, due to null semantics.

bodo/pandas/_util.h
```diff
@@ -465,6 +508,17 @@ void assert_py_args_is_tuple(PyObject *args, const char *err_context);
...
+/**
+ * @brief Get a single argument from a Python function call
+ * and convert it to a uint64 (unsigned long long).
+ *
+ * @param py_int Python object
+ * @param err_context String used for error messages, often the name of the
+ * function py_int came from
+ * @return uint64 representation of the Python string argument
+ */
+uint64_t get_py_object_as_uint64(PyObject *py_int, const char *err_context);
```
ehsantn: Is this function unused?

bodo/pandas/physical/expression.cpp
```diff
@@ -430,35 +430,38 @@ std::shared_ptr<array_info> do_arrow_compute_multi_input(
 std::shared_ptr<array_info> do_arrow_compute_binary(
     std::shared_ptr<ExprResult> left_res, std::shared_ptr<ExprResult> right_res,
     const std::string& comparator,
-    const std::shared_ptr<arrow::DataType> result_type) {
+    const std::shared_ptr<arrow::DataType> result_type,
+    bool sync_input_int_types) {
```
ehsantn: Can we avoid the `sync_input_int_types` input (cast only in Python side) and remove all the C++ code to simplify the code?

bodo/pandas/physical/expression.cpp
```diff
@@ -624,25 +636,32 @@ std::shared_ptr<array_info> do_arrow_compute_case(
     arrow::Datum src3 =
         ConvertExprResultToDatum(else_res, "do_arrow_compute else");
 
+    // Make input integer datums unsigned so that Arrow doesn't attempt a
+    // safe conversion to signed int that isn't possible
+    std::tie(src2, src3) =
+        CastIntDatumsToCommonType("do_arrow_compute_case arg", src2, src3);
```
ehsantn: Same comment about casting upstream in Python instead.

bodo/pandas/_util.cpp
```diff
@@ -19,6 +19,109 @@
 #include "duckdb/planner/filter/optional_filter.hpp"
 #include "physical/expression.h"
 
+std::vector<arrow::Datum> CastIntDatumsToCommonTypeImpl(
```
ehsantn: Same comment about casting upstream in Python and removing this code.

ehsantn (approved):
Thanks, Daniel.

bodo/pandas/physical/expression.cpp
dmsa-or: I tried taking this out of `do_arrow_compute_case`, but looks like it is still needed for `test_bitnot[case_uint64]`. To remove this completely from the C++ side I would need to look into doing something similar in Python `CaseExpression`s. What do you think?

ehsantn: I think handling types for `CaseExpression` in Python makes sense and would be good improvement.

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -2273,7 +2346,42 @@ def java_case_to_python_case(ctx, operands, input_plan):
...
+    case_expr = CaseExpression(
+        unified_empty_data, when_expr, unified_then_expr, unified_else_expr
+    )
+    # Restore the original return type if the types of then_expr and else_expr were different and our unification ended up casting one away from the intended result type.
+    # We do this with a CastExpression instead of via CaseExpression empty_data (which attempts to cast safely in _arrow_array_to_pd) so that integer overflow is allowed.
+    # At the moment we choose then_expr to be the result type - is there a better way to decide?
+    if then_needs_cast:
+        return CastExpression(then_expr.empty_data, case_expr)
+    else:
+        return case_expr
```
dmsa-or, line 2531: I noticed we had been setting empty_data for the `CaseExpression` to `then_expr.empty_data`. This was probably under the assumption that `then_expr` and `else_expr` had the same type. If they are not the same type, this means that the result type would change if the condition is simply inverted and `then_expr` and `else_expr` switch positions.

dmsa-or:
@ehsantn Is this good to merge?

ehsantn (approved):

PR #1246, opened 2026-06-22
Support numeric bool operations in C++ backend

dmsa-or:

Numeric boolean operations are now supported in the C++ backend and corresponding tests in `test_bool.py` have been enabled.
Also added a bit of extra logic to `ensure_type_of_expr` and `ensure_arg_is_const_expr_of_type` to allow for multiple possible accepted types.

## Testing strategy

CI

## Checklist


BodoSQL/bodosql/plan_conversion.py
```diff
@@ -1072,6 +1072,86 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+            left_expr = op_exprs[0]
+            right_expr = op_exprs[1]
+
+            ensure_type_of_expr(left_expr, "left_expr", (int, float))
+            ensure_type_of_expr(right_expr, "right_expr", (int, float))
+
+            # Round float inputs
+            int_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.int64()))
+            left_expr_rounded = ArrowScalarFuncExpression(
+                int_empty_data, [left_expr], "round", ()
```
DrTodd13: Sometimes "round" rounds to the nearest integer but keeps the data type the same and then if you really need an int you have to cast.  Do you know for certain this round is returning an int?

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -1072,6 +1072,86 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+            )
+            right_expr_rounded = ArrowScalarFuncExpression(
+                int_empty_data, [right_expr], "round", ()
+            )
+
+            # Get nonzero values as True, zero values as False
+            zero_expr = ConstantExpression(int_empty_data, input_plan, 0)
+            bool_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.bool_()))
+            left_expr_bool = ComparisonOpExpression(
+                bool_empty_data, left_expr_rounded, zero_expr, operator.ne
```
DrTodd13, line 1122: There may be a mismatch here that doesn't matter.  You create a 0 datum as int that compares the same as the float datum from the round operation.  Maybe just some renaming of things if you determine it is a float you get out of round and create a 0 float const as well.

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -1072,6 +1072,86 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+                )
+
+        if func_name == "BOOLNOT" and len(op_exprs) == 1:
+            expr = op_exprs[0]
+            ensure_type_of_expr(expr, "expr", (int, float))
+
+            # Round float inputs
+            int_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.int64()))
+            expr_rounded = ArrowScalarFuncExpression(
+                int_empty_data, [expr], "round", ()
```
DrTodd13, line 1148: Above expr can be int but then here you can call round on it?  Only do this section if datatype is float?

DrTodd13 (approved):

BodoSQL/bodosql/plan_conversion.py
dmsa-or: Hmm, it probably returns a float but there might be a cast somewhere or the types are considered compatible. Further investigation needed.

BodoSQL/bodosql/plan_conversion.py
dmsa-or: I thought it wouldn't hurt, but I guess checking is worth it to avoid trying to round all the elements for no reason.

BodoSQL/bodosql/plan_conversion.py
dmsa-or: Exactly comparing floats seems risky, I would rather just cast to int if necessary. But given that the round expression currently works with int_empty_data, I'm not sure that there is an issue. I wouldn't want to rely on undefined behavior though.

BodoSQL/bodosql/plan_conversion.py
DrTodd13: 0 is an exact representation.

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -1072,6 +1072,86 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
             return UnaryOpExpression(out_empty, inp, "round")
 
+        if func_name in ("BOOLAND", "BOOLOR", "BOOLXOR") and len(op_exprs) == 2:
+            left_expr = op_exprs[0]
+            right_expr = op_exprs[1]
+
+            ensure_type_of_expr(left_expr, "left_expr", (int, float))
+            ensure_type_of_expr(right_expr, "right_expr", (int, float))
+
+            # Round float inputs
```
ehsantn: Why does it need to round float input? The float comparison to zero is limited in Arrow (works for me)? Is it for matching the data type to zero (you can have different zero types)? The Python implementation doesn't round: https://github.com/bodo-ai/Bodo/blob/2c5a21d02e0830c570d3355f08a1d9c721b4b544/BodoSQL/bodosql/kernels/other_array_kernels.py#L218

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -1072,6 +1072,86 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+                int_empty_data, [right_expr], "round", ()
+            )
+
+            # Get nonzero values as True, zero values as False
+            zero_expr = ConstantExpression(int_empty_data, input_plan, 0)
+            bool_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.bool_()))
+            left_expr_bool = ComparisonOpExpression(
+                bool_empty_data, left_expr_rounded, zero_expr, operator.ne
+            )
+            right_expr_bool = ComparisonOpExpression(
```
ehsantn, line 1124: Do the null semantics work here? `BOOLAND` returns `false` if one side is zero and the other is null. But Arrow returns null if either side is null.

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -1072,6 +1072,86 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+                    bool_empty_data, left_expr_bool, right_expr_bool, operator.ne
+                )
+
+        if func_name == "BOOLNOT" and len(op_exprs) == 1:
+            expr = op_exprs[0]
+            ensure_type_of_expr(expr, "expr", (int, float))
+
+            # Round float inputs
+            int_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.int64()))
+            expr_rounded = ArrowScalarFuncExpression(
```
ehsantn, line 1147: Same about rounding.

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -1072,6 +1072,86 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+            return ComparisonOpExpression(
+                bool_empty_data, expr_rounded, zero_expr, operator.eq
+            )
+
+        if func_name == "EQUAL_NULL" and len(op_exprs) == 2:
+            left_expr = op_exprs[0]
+            right_expr = op_exprs[1]
+
+            bool_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.bool_()))
+            values_equal = ComparisonOpExpression(
```
ehsantn, line 1163: Implementations like this may be simpler and faster with a case statement. Not a big deal here.

ehsantn (commented):
See comments.

BodoSQL/bodosql/plan_conversion.py
dmsa-or: You need to round according to Snowflake docs. For example, floats like 0.3 that round to zero should be interpreted as 0.

ehsantn: So the JIT backend can be wrong (haven't tested it).

ehsantn: Let's add a comment to the code.

BodoSQL/bodosql/plan_conversion.py
dmsa-or: Yes, it does work because we use the Kleene logic versions of AND and OR. Take the BOOLAND case. If `left_expr_rounded` is 0, then `left_expr_bool` will be `false`. If `right_expr_rounded` is null, then `right_expr_bool` will be null. Then `false` AND `null` is `false` under Kleene logic.

ehsantn (approved):
Thanks, Daniel.

PR #1247, opened 2026-06-22
Support more cast cases in C++ backend

dmsa-or:

Fixed integer to timestamp casting.
Made a couple tweaks to cast tests so they pass in the new backend.
Enabled cast tests on C++ backend aside from those using unsupported literal types (e.g. DECIMAL, OTHER)
Refactored and improved the expression type checking to be more robust in handling string inputs and Numpy/Pandas/PyArrow datatype classes.

TRY_CAST seems difficult to support because Arrow has no option to emit null instead of throwing an error for an invalid cast.

## Testing strategy

CI

## Checklist


BodoSQL/bodosql/tests/test_cast.py
```diff
@@ -600,12 +599,20 @@ def test_tz_aware_datetime_to_char_cast(
...
 
-    out_arr = tz_aware_df["TABLE1"]["A"].astype(str)
     if bodosql.use_cpp_backend:
+        # US/Pacific is not in Arrow's timezone database, so we swap with the equivalent America/Los_Angeles.
+        # In the future we may want a way to intercept the cast input and switch the timezones with
+        # ones recognized by Arrow before passing to Arrow's cast function.
+        if tz_aware_df["TABLE1"]["A"].dt.tz.key == "US/Pacific":
+            tz_aware_df["TABLE1"]["A"] = tz_aware_df["TABLE1"]["A"].dt.tz_convert(
+                "America/Los_Angeles"
+            )
```
dmsa-or, line 609: Date/time related issue to maybe discuss with @IsaacWarren about.

IsaacWarren: Let's create an issue and reference it here

IsaacWarren (approved):
Thanks @dmsa-or

BodoSQL/bodosql/tests/test_cast.py
```diff
@@ -97,43 +97,9 @@ def test_cast_str_to_numeric(basic_df, use_sf_cast_syntax, memory_leak_check):
     )
 
 
-def test_numeric_to_str(basic_df, use_sf_cast_syntax, memory_leak_check):
```
scott-routledge2, line 100: Why are we removing this test?

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -1885,15 +1899,72 @@ def process_one_search_option(so):
...
+    if expected_type is bool:
+        return pd.api.types.is_bool_dtype(obj_type)
+
+    if pd.api.types.is_dtype_equal(obj_type, expected_type):
+        return True
+
+    if isinstance(obj_type, pa.DataType) and isinstance(expected_type, pa.DataType):
+        if obj_type.equals(expected_type):
+            return True
+    if isinstance(obj_type, pa.DataType):
```
scott-routledge2: Is this saying the result of compare_types on a string type and some arbitrary object would return True? Have we tested this?

scott-routledge2 (approved):
LGTM, Thanks @dmsa-or

BodoSQL/bodosql/tests/test_cast.py
dmsa-or: This test wasn't even being run because there was another test below with the same name. The tests are identical except for the target string type. The one I deleted casted to CHAR, so I just added that as a test case to `test_numeric_to_str` below.

BodoSQL/bodosql/plan_conversion.py
dmsa-or: Yeah, I changed it to the suggestion in the code comment and it appears to be working a lot better. Let me know if you have other ideas for improvements.

BodoSQL/bodosql/tests/test_cast.py
dmsa-or: https://bodo.atlassian.net/browse/BSE-5471

BodoSQL/bodosql/plan_conversion.py
scott-routledge2: I didn't see anything else, it might be nice to have some unit tests for this function. But not a blocker for this PR.

scott-routledge2 (approved):

PR #1248, opened 2026-06-23
[BSE-5175] Iceberg Join FIlter IO Pushdown

scott-routledge2:

Add support for Iceberg Runtime Join filters in BodoSQL C++ backend and allow generating min/max filter values for the I/O nodes, reusing the existing DF infrastructure. 

Also extends `LogicalGetIcebergRead` to pass selected columns, limit along with join filter information, to create a logical iceberg read node that is ready to be converted to a physical plan.

## Testing strategy

Ran tests found in `BodoSQL/bodosql/tests/test_types/test_filesystem_catalog_runtime_join.py` using DF LIB mode and checked that filters were created manually. 

Ran Pydough Iceberg Filesystem end to end tests for additional smoke testing. 

I also ran some manual tests on GPU, but there were some other (seemingly unrelated) issues with iceberg read preventing me from running tests in BodoSQL/bodosql/tests/test_types/test_filesystem_catalog_runtime_join.py

## Checklist


bodo/pandas/iceberg_utils.py
```diff
@@ -0,0 +1,124 @@
+"""
```
scott-routledge2, line 1: This file is mostly just refactoring our Iceberg read into utilities to reuse for converting SQL plans.

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -2187,28 +2193,56 @@ def java_rtjf_to_python_rtjf(ctx, java_plan):
...
         filter_cols = [-1] * len(eq_cols)
         is_first = [False] * len(is_first_cols)
+        build_cols = [-1] * len(eq_cols)
 
         for loc_ind, key in enumerate(orig_key_locs):
             filter_cols[key] = eq_cols[loc_ind]
             is_first[key] = is_first_cols[loc_ind]
+            build_cols[key] = orig_build_keys[loc_ind]
+
+        # orig_build_key_cols for JoinFilterInfo requires a list of indexes
```
scott-routledge2: We don't have good tests for this yet, but my understand is that, for any given join_id/fid, each of the elements in `eq_cols` corresponds to an item in `JoinInfo.leftKeys`, which corresponds to an equality condition with the item from `JoinInfo.rightKeys` at the same position.

The order of the rightKeys/leftKeys lists might change during column pruning, so we track the original filter column location in the Join object so we can reorder join filter columns back to the order they appear in the final join. So after applying the reordering from `orig_key_locs`, each key in `filter_cols` will correspond to a filter derived from values from the key/column in `JoinInfo.rightKeys` at that same position (filter_col can be -1 if the filter column has not emerged yet). 

At the same time, we create a vector of optional min/max values (it's optional since some column types do not support min/max filters.) in the C++ JoinState. 

We then collect these min/max values at `join_state->min_max_values`, where each element into this vector (length is <num_build_keys>) corresponds to the min/max value for that build key. Here, the index implies the order that key appears in the join condition. 

Finally, we use the values at orig_build_key_cols to lookup the min/max values for the corresponding filter column.

https://github.com/bodo-ai/Bodo/blob/f8cbfd4705e346a860fc4121c6735d9e8960d2c0/bodo/pandas/optimizer/runtime_join_filter.cpp#L282
https://github.com/bodo-ai/Bodo/blob/f8cbfd4705e346a860fc4121c6735d9e8960d2c0/bodo/pandas/_util.cpp#L1182
https://github.com/bodo-ai/Bodo/blob/0edd4715fdbb302f505962e3dcdf484f7e971c4a/bodo/libs/streaming/_join.cpp#L1360

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -2377,28 +2383,54 @@ def java_rtjf_to_python_rtjf(ctx, java_plan):
...
+        equality_filter_columns=new_equality_filter_columns,
+        orig_build_key_cols=new_orig_build_keys,
+        equality_is_first_locations=new_equality_is_first_locations,
+    )
 
+
+def generate_runtime_join_filter(
+    join_info: JoinFilterInfo, input_plan: LazyPlan
+) -> LogicalJoinFilter:
+    """Contruct an instance of a runtime join filter plan node from the given join filter info and input plan."""
```
```suggestion
    """Construct an instance of a runtime join filter plan node from the given join filter info and input plan."""
```

bodo/pandas/_physical_conv.cpp
```diff
@@ -832,13 +832,14 @@ void PhysicalPlanBuilder::Visit(bodo::LogicalJoinFilter& op) {
         physical_op = std::make_shared<PhysicalJoinFilter>(
             op, in_table_schema, this->join_filter_states);
     }
+    bool found_join_on_same_device = false;
 #else   // USE_CUDF
     std::shared_ptr<PhysicalJoinFilter> physical_op =
         std::make_shared<PhysicalJoinFilter>(op, in_table_schema,
                                              this->join_filter_states);
+    bool found_join_on_same_device = true;
```
IsaacWarren, line 840: Is there ever case where this could be false? I can't think of one since we support all the joins that can be generated but I'm not sure

bodo/pandas/_plan.cpp
```diff
@@ -1596,20 +1595,43 @@ duckdb::unique_ptr<duckdb::LogicalGet> make_dataframe_get_parallel_node(
...
+    std::optional<JoinFilterProgramState> rtjf_state_map_opt) {
     duckdb::shared_ptr<duckdb::Binder> binder = get_duckdb_binder();
 
     // Convert Arrow schema to DuckDB
     std::shared_ptr<arrow::Schema> arrow_schema = unwrap_schema(pyarrow_schema);
     auto [return_names, return_types] = arrow_schema_to_duckdb(arrow_schema);
 
+    // Apply selected columns if provided by Calcite planner
+    // i.e. we are not running the duckdb optimizer.
+    if (selected_columns_opt.has_value()) {
```
IsaacWarren, line 1610: Do we need to do this for parquet too?

bodo/tests/utils.py
```diff
@@ -235,7 +237,7 @@ def check_func(
     # If dataframe_library_enabled then run compiler tests as df library tests
     # (replaces import pandas as pd with import bodo.pandas as pd)
     # NOTE: This variable takes precedence over other variables
-    only_df_lib = bodo.test_dataframe_library_enabled
+    only_df_lib = bodo.test_dataframe_library_enabled or only_python
```
IsaacWarren: Why can't we just rely on  bodo.test_dataframe_library_enabled?

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -3045,16 +3083,22 @@ def generate_iceberg_read(read_info):
...
         row_filter=row_filter,
+        join_filter_info=read_info.join_filter_info,
         selected_fields=read_fields,
         limit=read_info.limit,
     )
-    return df._plan
+
+    # Insert Runtime Join Filters on top of the read if needed
+    if read_info.join_filter_info is not None:
+        plan = generate_runtime_join_filter(read_info.join_filter_info, plan)
```
IsaacWarren, line 3114: Join filters can be anywhere in the plan, not just directly on top of IO nodes, does this support that case?

IsaacWarren (approved):
Thanks Scott, have we manually confirmed files being pruned during the read?

bodo/pandas/_physical_conv.cpp
scott-routledge2: This is part of the "#else" when building without CUDA, all joins/filters should be on the CPU.

bodo/tests/utils.py
scott-routledge2: I think we can if we set/unset the flag on a per-test basis. I was thinking if we set the flag globally then it would override all tests that use check_func and only run them in DF lib mode.

BodoSQL/bodosql/plan_conversion.py
scott-routledge2: This branch will only insert join filters when the filter can be pushed below an Iceberg conversion node, for example:
```
 IcebergToBodoPhysicalConverter
                    IcebergRuntimeJoinFilter(joinIDs=[[0, 1, 2]], equalityColumnsList=[[[1], [2], [3]]], allEqualityKeysReady=[[true, true, true]], nonEqualityColumnsList=[[[], [], []]])
                      IcebergFilter(condition=[AND(IS NOT NULL($1), IS NOT NULL($2), IS NOT NULL($3))])
```

If the join filter occurs higher up in the plan, that gets handled by `java_plan_to_python_plan`

scott-routledge2:
> have we manually confirmed files being pruned during the read?

I did some manual checking to make sure the filters are being generated before file pruning occurs. I also added a test that checks for this log message (see `BodoSQL/bodosql/tests/test_types/test_filesystem_catalog_runtime_join.py::test_simple_join\[min_max\]`

bodo/tests/utils.py
IsaacWarren: Maybe we can use the cpp_backend marker?

scott-routledge2: I like that idea.

bodo/pandas/_plan.cpp
scott-routledge2: Just tested this, looks like we will need to add this for Parquet as well in addition to filter pushdown. I'll open a followup

BodoSQL/bodosql/plan_conversion.py
ehsantn: Add this as TODO and comment in the code itself? Is there anywhere else in the code that needs documented to clarify this further?

ehsantn (approved):
Thanks @scott-routledge2 .

BodoSQL/bodosql/plan_conversion.py
scott-routledge2: Add todo for more tests for Join keys where the order changes after column pruning? We could also probably add more comments to the C++ code that deals with build key min-max lookup. I was also thinking we could refactor the variable: `orig_build_key_cols` to something like `build_key_ids`, to differentiate the index into the min-max value array from the column index into the build table?

ehsantn: I'm not familiar with this part of the code enough, but seems like some code documentation is missing to make the code more understandable.

PR #1255, opened 2026-06-25
Complete SEARCH handling and enable test_in.py on BodoSQL C++ backend

dmsa-or:

Finish the unhandled cases for SEARCH plan conversion. This is mainly for completeness because, as far as I've found, there aren't any tests that actually rely on SEARCH for ranges wider than a single value.
Enable `test_in.py` on C++ backend. (IN was the one case where SEARCH is sometimes needed, but our previous SEARCH code already dealt with that).

## Testing strategy

CI

## Checklist

### DIFF
--- a/BodoSQL/bodosql/plan_conversion.py
+++ b/BodoSQL/bodosql/plan_conversion.py
@@ -1999,32 +1999,68 @@ def java_call_to_python_call(ctx, java_call, input_plan):
                 # T        T        F              T
                 # T        T        F              F
 
-                # Also, the case of having lower and upper bound and both being
-                # inclusive can be divided up into two cases, where lower and
-                # upper are equal and where they are not equal.
+                # The lower and upper bounds being equal is a special case that does not
+                # involve less-than or greater-than comparison. It can be subdivided
+                # based on whether the lower and upper bounds are inclusive or not.
+                # The typical case is that they will be inclusive, which we can simplify
+                # to an equality check.
                 if (
                     lower_lit is not None
                     and upper_lit is not None
-                    and lower_inclusive
-                    and upper_inclusive
                     and lower_lit == upper_lit
                 ):
-                    # The exception case where we have lower and upper bounds and they are both
-                    # inclusive and the same we can simplify as equality check.
+                    if lower_inclusive and upper_inclusive:
+                        # Range of the form [a..a] - reduce to equality check
+                        const_empty_data = arrow_to_empty_df(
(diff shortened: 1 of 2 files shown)


BodoSQL/bodosql/plan_conversion.py
```diff
@@ -1999,32 +1999,68 @@ def process_one_search_option(so):
...
+                            bool_empty_data,
+                            src,
+                            ConstantExpression(const_empty_data, input_plan, lower_lit),
+                            operator.eq,
+                        )
+                    elif lower_inclusive or upper_inclusive:
+                        # Range of the form [a..a) or (a..a] - interpret as empty
+                        return ConstantExpression(bool_empty_data, input_plan, False)
+                    else:
+                        raise ValueError("SEARCH option range form (a..a) is invalid.")
```
DrTodd13, line 2028: why is (a..a) invalid and (a..a] isn't?  can't you just have a single else and return the constant expression?

DrTodd13 (approved):
thanks

BodoSQL/bodosql/plan_conversion.py
dmsa-or: I'm just following what it says here: https://www.javadoc.io/doc/com.google.guava/guava/23.0/com/google/common/collect/Range.html

> When both endpoints exist, the upper endpoint may not be less than the lower. The endpoints may be equal only if at least one of the bounds is closed:
> [a..a] : a singleton range
> [a..a); (a..a] : empty ranges; also valid
> (a..a) : invalid; an exception will be thrown

BodoSQL/bodosql/tests/test_in.py
```diff
@@ -75,10 +78,11 @@ def test_in_scalar_literals(basic_df, memory_leak_check):
         None,
         check_names=False,
         check_dtype=False,
-        return_codegen=True,
+        return_codegen=not bodosql.use_cpp_backend,
```
scott-routledge2, line 81: Does this query evaluate to a less than + greater than plan when converting to Python?

scott-routledge2 (approved):
Thanks @dmsa-or

BodoSQL/bodosql/tests/test_in.py
dmsa-or: For IN cases like this one that are translated to SEARCH, the possible values in the list are always turned into individual singleton ranges. So we generate three equality comparisons in this test.   The conversion to a less-than and greater-than plan is only for BETWEEN, theoretically.

PR #1262, opened 2026-06-30
Support test_some_any_all.py in C++ backend

ehsantn:

Fill some gaps to support most of test_some_any_all.py in C++ backend.

## Testing strategy

Enabled unit tests.

## User facing changes

None.

## Checklist


DrTodd13 (approved):
I'd prefer that we don't let the CPU and GPU sides get out of sync.  I've already ran into problems with that and do we believe that the GPU side is less compelling for bodosql compared to dflib?

dmsa-or (approved):
Thanks Ehsan.

ehsantn:
> I'd prefer that we don't let the CPU and GPU sides get out of sync. I've already ran into problems with that and do we believe that the GPU side is less compelling for bodosql compared to dflib?

I don't have a local GPU so implementing the GPU part simultaneously isn't easy for me. My assumption was that focusing on the GPU side separately is easier.

PR #1264, opened 2026-07-01
Fix bool type comparison

ehsantn:
As title. Fixes some nightly errors.

### DIFF
--- a/BodoSQL/bodosql/plan_conversion.py
+++ b/BodoSQL/bodosql/plan_conversion.py
@@ -2476,7 +2476,11 @@ def compare_types(obj_type, expected_type):
         expected_type = pd.api.types.pandas_dtype(expected_type)
 
     if isinstance(obj_type, type) and isinstance(expected_type, type):
-        if issubclass(obj_type, expected_type):
+        if issubclass(obj_type, expected_type) and not (
+            obj_type is bool and expected_type is int
+        ):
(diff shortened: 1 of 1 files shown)


dmsa-or (approved):
Thanks. Scott had suggested maybe writing some unit tests for this function to catch edge cases like this.

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -2476,7 +2476,11 @@ def compare_types(obj_type, expected_type):
         expected_type = pd.api.types.pandas_dtype(expected_type)
 
     if isinstance(obj_type, type) and isinstance(expected_type, type):
-        if issubclass(obj_type, expected_type):
+        if issubclass(obj_type, expected_type) and not (
+            obj_type is bool and expected_type is int
+        ):
+            # bool is a subclass of int in Python, but we don't want to consider them
```
DrTodd13: This comment is confusing.  The docstring says return True if subclass or equivalent but here you say we don't want to consider them equivalent yet return true.

DrTodd13 (approved):

BodoSQL/bodosql/plan_conversion.py
ehsantn: Moved the comment up to be maybe more clear. We don't consider bool and int equivalent in the Arrow case so has to be consistent and clearly documented. This function needs more work in general after testing these cases.

PR #1266, opened 2026-07-01
Fix bug in Iceberg to Arrow String Filter Conversion

scott-routledge2:

There was a bug in make unique where it would replace the name of a variable multiple times.

For example if your string was "f0 f1" and your new names are {f0: f1, f1: f2}, it would first replace f0 with f1, making the string "f1 f1" then it would replace f1 with f2 making it "f2 f2". 

This bug was exposed because previously our iceberg filters were left-heavy but after switching to the dataframe library they became right-heavy. For some reason.

So for example 

AND(AND(x,y),z) is now AND(x, AND(y,z)).

## Testing strategy

New test added for this case

## Checklist

### DIFF
--- a/bodo/io/iceberg/filter_conversion.py
+++ b/bodo/io/iceberg/filter_conversion.py
@@ -46,8 +46,9 @@ class _ConvertToArrowExpressionStringAndScalar(
             (right_rename_map[name], value) for name, value in right_child[1]
         ]
         # Rename the right child names in the expression string
+        # Matches whole words that are not inside {} and are in the right_rename_map
         pattern = re.compile(
-            r"\b(" + "|".join(map(re.escape, right_rename_map)) + r")\b"
+            r"(?<!\{)\b(" + "|".join(map(re.escape, right_rename_map)) + r")\b(?!\})"
         )
         right_child_expr = pattern.sub(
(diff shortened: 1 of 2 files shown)


bodo/io/iceberg/filter_conversion.py
```diff
@@ -45,9 +46,13 @@ def make_names_unique(
             (right_rename_map[name], value) for name, value in right_child[1]
         ]
         # Rename the right child names in the expression string
-        right_child_expr = right_child[0]
-        for old_name, new_name in right_rename_map.items():
-            right_child_expr = right_child_expr.replace(old_name, new_name)
+        pattern = re.compile(
```
scott-routledge2, line 54: Using regex to replace everything at once instead of one at a time.

Also if you have a lot of variables on the right side, replace("f1", "fX") would replace f1 and also the f1 in f10.

IsaacWarren (approved):
Thanks Scott

bodo/io/iceberg/filter_conversion.py
ehsantn: Let's add this comment in the code to clarify the reason for using regex.

ehsantn (approved):
Thanks Scott.

PR #1267, opened 2026-07-01
Validate arguments of Series _cmp_method with scalar input.

scott-routledge2:

Pandas raises a type error for certain comparisons between Series and scalar values that we currently do not have. We either throw a runtime error or return wrong results in those cases.

Wrong result examples include string series < date scalar
While runtime (i.e. plan execution) error examples include struct series < int scalars

The issue is that Pandas allows pretty much any scalar type to be compared to an empty Series. This PR adds an extra validation check that calls _cmp_method with a representative value to handle those cases. 

## Checklist

### DIFF
--- a/bodo/pandas/series.py
+++ b/bodo/pandas/series.py
@@ -225,6 +225,7 @@ class BodoSeries(pd.Series, BodoLazyWrapper):
         assert isinstance(empty_data, pd.Series), "_cmp_method: Series expected"
 
         lhs_plan, lhs, rhs = _handle_series_binop_args(self._plan, other)
+
         # Match the source plans of lhs and rhs, if they don't match return None, None
         lhs, rhs = match_binop_expr_source_plans(lhs, rhs)
         if lhs is None and rhs is None:
--- a/bodo/pandas/series.py
+++ b/bodo/pandas/series.py
@@ -10,7 +10,6 @@ import typing as pt
 import warnings
 from collections.abc import Callable, Hashable
 from concurrent.futures import ThreadPoolExecutor
-from decimal import Decimal
 
 import numpy
 import numpy as np
@@ -2509,9 +2508,6 @@ def representative_value_for_arrow_type(dtype: pa.DataType):
     if pa.types.is_floating(dtype):
         return 0.0
 
-    if pa.types.is_decimal(dtype):
-        return Decimal("0")
-
     if pa.types.is_string(dtype) or pa.types.is_large_string(dtype):
         return ""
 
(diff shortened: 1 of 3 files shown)


bodo/tests/test_df_lib/test_frontend.py
```diff
@@ -417,3 +418,33 @@ def test_timestamp_now():
...
+    "data, value",
+    [
+        pytest.param(["1994-12-31"], date(1995, 1, 1), id="string_series"),
+        pytest.param(
+            [{"A": 1.1, "B": {"A": date(1995, 1, 1)}}],
+            date(1995, 1, 1),
+            id="struct_series",
+        ),
+        pytest.param([[False, True, True]], date(1995, 1, 1), id="list_series"),
+        pytest.param([datetime(1995, 1, 1)], date(1995, 1, 1), id="datetime_series"),
```
scott-routledge2, line 435: Pandas raises an error when comparing date/datetime and datetime/date
Also timestamp with timezone vs timestamp without

ehsantn: Let's add this comment to the code.

bodo/pandas/series.py
```diff
@@ -2491,6 +2494,78 @@ def is_numeric(other):
...
+        return datetime.date(2000, 1, 1)
+
+    if pa.types.is_timestamp(dtype):
+        tz = dtype.tz
+        if tz is not None:
+            return pd.Timestamp("2000-01-01 00:00:00", tz=tz)
+        return pd.Timestamp("2000-01-01 00:00:00")
+
+    if pa.types.is_time32(dtype) or pa.types.is_time64(dtype):
+        return datetime.time(0, 0, 0)
```
ehsantn, line 2528: Does Pandas support time value comparison? Does this need to be an Arrow scalar instead of datetime.time?

ehsantn (approved):
Thanks, Scott.

DrTodd13 (approved):

bodo/pandas/series.py
scott-routledge2: Comparing a time series with a datetime.time works, but comparing with a pyarrow scalar gives an error.

scott-routledge2: Actually I was mistaken, a pa.scalar(time(0,0,0)) raises an error in both Bodo and Pandas because we rely on:
`empty_data = zero_size_self._cmp_method(zero_size_other, op)`

Which gives an error:
`NotImplementedError: lt not implemented for <class 'pyarrow.lib.Time64Scalar'>`

But while the following code works in Pandas:

``` python
df = pd.Series([time(10, 0, 0)], dtype=pd.ArrowDtype(pa.time64("ns")))

print(df <= time(10, 0, 0))
```

Bodo throws an error:
`NotImplementedError: Unknown expr type in make_const_expr <class 'datetime.time'>`

scott-routledge2: Opening a followup issue to support datetime.time values in plans

PR #1268, opened 2026-07-02
Support more date/time functions in BodoSQL C++ backend

dmsa-or:

More date/time functions are supported in the C++ backend, including `NEXT_DAY`/`PREVIOUS_DAY`, `DAYOFWEEKISO`, `YEAROFWEEK`, `YEAROFWEEKISO`, `YEARWEEK`, `STR_TO_DATE`.
This also fixes the 2-argument form of DATEADD when the input is an interval type.

## Testing strategy

PR CI and Nightly

## Checklist

### DIFF
--- a/BodoSQL/bodosql/plan_conversion.py
+++ b/BodoSQL/bodosql/plan_conversion.py
@@ -577,26 +577,25 @@ def java_call_to_python_call(ctx, java_call, input_plan):
                     nano_scale_expr = ConstantExpression(
                         int_empty, input_plan, 86_400_000_000_000
                     )
+
+                    return ArrowScalarFuncExpression(
+                        out_empty,
+                        [
+                            date_expr,
+                            amount_expr,
+                            ConstantExpression(int_empty, input_plan, 0),
+                            nano_scale_expr,
+                        ],
+                        "bodo_dateadd",
+                        (),
+                    )
                 else:
                     # Assume we have an interval type (e.g. a MonthDayNanoInterval tuple or duration type).
-                    # Get nanoseconds from interval literal by casting to int64.
-                    amount_expr = CastExpression(
-                        pd.Series(dtype=pd.ArrowDtype(pa.int64())), amount_expr
+                    # Add the interval via ArithOpExpression, which should use DuckDB's calendar-aware
+                    # interval arithmetic as long as amount_expr is a ConstantExpression.
+                    return ArithOpExpression(
+                        out_empty, date_expr, amount_expr, "__add__"
                     )
-                    # nano_scale is 1 since we are already in units of nanoseconds
-                    nano_scale_expr = ConstantExpression(int_empty, input_plan, 1)
-
-                return ArrowScalarFuncExpression(
-                    out_empty,
-                    [
-                        date_expr,
-                        amount_expr,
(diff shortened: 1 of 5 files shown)


BodoSQL/bodosql/tests/test_datetime_fns.py
```diff
@@ -3317,23 +3301,7 @@ def test_date_trunc_unquoted_timeunit(dt_fn_dataframe, memory_leak_check):
     )
 
 
-@pytest.mark.tz_aware
```
IsaacWarren, line 3320: Why remove this?

IsaacWarren (approved):

BodoSQL/bodosql/tests/test_datetime_fns.py
dmsa-or: There's an identical test with the same name below.

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -460,13 +565,30 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+                        int_empty, input_plan, 86_400_000_000_000
+                    )
+                else:
+                    # Assume we have an interval type (e.g. a MonthDayNanoInterval tuple or duration type).
+                    # Get nanoseconds from interval literal by casting to int64.
+                    amount_expr = CastExpression(
+                        pd.Series(dtype=pd.ArrowDtype(pa.int64())), amount_expr
+                    )
+                    # nano_scale is 1 since we are already in units of nanoseconds
+                    nano_scale_expr = ConstantExpression(int_empty, input_plan, 1)
```
dmsa-or: Does this work for intervals that include months? I see we convert MonthDayNanoInterval to duration('ns') elsewhere, but could this lead to problems? What's the best way to add months to a date in Arrow?

dmsa-or: @IsaacWarren

IsaacWarren: I think we were using duckdbs interval so it's calendar aware, I don't think arrow intervals are calendar aware iirc

dmsa-or: Don't you lose calendar awareness by converting to nanoseconds like in the snippet above? Maybe I should instead use `ArithOpExpression` `__add__` instead of `bodo_dateadd` if the argument is already an interval type. Or does that go straight to Arrow?

ehsantn (approved):
Thanks @dmsa-or.

BodoSQL/bodosql/plan_conversion.py
IsaacWarren: I'm not sure off the top of my head, I think bodo_dateadd supports intervals. Yes, converting to nanoseconds loses calendar awareness

dmsa-or: To me it looks like `ArithOpExpression` `__add__` for interval types turns into `PhysicalCalendarIntervalExpression` which uses DuckDB for the calculation, so I've switched to that. `bodo_dateadd` does not appear to support interval types, at least not in a calendar-aware way.

IsaacWarren (approved):

PR #1272, opened 2026-07-07
Support more date/time functions in BodoSQL C++ backend (part 2)

dmsa-or:

More date/time functions are supported in the C++ backend, including `TIMESTAMP_FROM_PARTS` (and variations), `DATE_FROM_PARTS`, `TIME_FROM_PARTS`, `MAKEDATE` with array input, `MONTHS_BETWEEN`, `TIMEZONE_HOUR`/`TIMEZONE_MINUTE`, and `EPOCH` part functions.
This PR also improves the range of date/time strings SQL cast functions can parse.

## Testing strategy

PR CI and Nightly

## Checklist

### DIFF
--- a/BodoSQL/bodosql/plan_conversion.py
+++ b/BodoSQL/bodosql/plan_conversion.py
@@ -577,26 +577,25 @@ def java_call_to_python_call(ctx, java_call, input_plan):
                     nano_scale_expr = ConstantExpression(
                         int_empty, input_plan, 86_400_000_000_000
                     )
+
+                    return ArrowScalarFuncExpression(
+                        out_empty,
+                        [
+                            date_expr,
+                            amount_expr,
+                            ConstantExpression(int_empty, input_plan, 0),
+                            nano_scale_expr,
+                        ],
+                        "bodo_dateadd",
+                        (),
+                    )
                 else:
                     # Assume we have an interval type (e.g. a MonthDayNanoInterval tuple or duration type).
-                    # Get nanoseconds from interval literal by casting to int64.
-                    amount_expr = CastExpression(
-                        pd.Series(dtype=pd.ArrowDtype(pa.int64())), amount_expr
+                    # Add the interval via ArithOpExpression, which should use DuckDB's calendar-aware
+                    # interval arithmetic as long as amount_expr is a ConstantExpression.
+                    return ArithOpExpression(
+                        out_empty, date_expr, amount_expr, "__add__"
                     )
-                    # nano_scale is 1 since we are already in units of nanoseconds
-                    nano_scale_expr = ConstantExpression(int_empty, input_plan, 1)
-
-                return ArrowScalarFuncExpression(
-                    out_empty,
-                    [
-                        date_expr,
-                        amount_expr,
-                        ConstantExpression(int_empty, input_plan, 0),
-                        nano_scale_expr,
-                    ],
-                    "bodo_dateadd",
-                    (),
-                )
             elif num_operands == 3:
                 # 3-operand form: DATEADD(unit, amount, date) → date + (unit * amount)
                 # First operand is a FLAG(unit) interval qualifier from the Java
--- a/BodoSQL/bodosql/plan_conversion.py
+++ b/BodoSQL/bodosql/plan_conversion.py
@@ -335,6 +335,429 @@ def java_call_to_python_call(ctx, java_call, input_plan):
             date_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.date32()))
             return CastExpression(date_empty_data, timestamp_expr)
 
+        def timestamp_from_parts(
+            year_expr,
+            month_expr,
+            day_expr,
+            hour_expr=None,
+            minute_expr=None,
+            second_expr=None,
+            nanosecond_expr=None,
+        ):
+            # Arrow currently lacks a function to make a date or timestamp from parts:
+            # https://github.com/apache/arrow/issues/49514.
+            # There is some possibility of it being added in the future,
+            # so we can revisit this strptime approach then.
+
+            int_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.int64()))
+            float_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.float64()))
+            bool_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.bool_()))
+            string_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.string()))
+
+            max_int64 = ConstantExpression(
+                int_empty_data, input_plan, np.iinfo(np.int64).max
+            )
+            min_int64 = ConstantExpression(
+                int_empty_data, input_plan, np.iinfo(np.int64).min
+            )
+
+            # If any of the inputs are floats, we will round to the nearest integer.
+            # This is what the JIT implementation seems to do, even though float input isn't explicitly allowed in Snowflake.
+            def int_part_expr(part_expr, part_expr_name):
+                ensure_type_of_expr(part_expr, part_expr_name, (int, float))
+                part_expr_dtype = get_expr_dtype(part_expr, part_expr_name)
+                if compare_types(part_expr_dtype, float):
+                    part_expr = ArrowScalarFuncExpression(
+                        float_empty_data,
+                        [part_expr],
+                        "round",
+                        (0, "half_towards_infinity"),
+                    )
+
+                    # The BodoSQL docs say we should return NULL when any input cannot be converted to int64
+                    part_too_large = ComparisonOpExpression(
+                        bool_empty_data, part_expr, max_int64, operator.gt
+                    )
+                    part_too_small = ComparisonOpExpression(
+                        bool_empty_data, part_expr, min_int64, operator.lt
+                    )
+                    part_out_of_bounds = ConjunctionOpExpression(
+                        bool_empty_data, part_too_large, part_too_small, "__or__"
+                    )
+                    part_expr = CaseExpression(
+                        float_empty_data,
+                        part_out_of_bounds,
+                        NullExpression(float_empty_data, input_plan, 0),
+                        part_expr,
+                    )
+
+                    part_expr = CastExpression(int_empty_data, part_expr)
+                return part_expr
+
+            year_expr = int_part_expr(year_expr, "year_expr")
+            month_expr = int_part_expr(month_expr, "month_expr")
+            day_expr = int_part_expr(day_expr, "day_expr")
+
+            if hour_expr:
+                hour_expr = int_part_expr(hour_expr, "hour_expr")
+            if minute_expr:
+                minute_expr = int_part_expr(minute_expr, "minute_expr")
+            if second_expr:
+                second_expr = int_part_expr(second_expr, "second_expr")
+                # Ensure seconds are int64 so the result of the later multiplication doesn't
+                # end up being int32 internally, potentially leading to overflow and the wrong answer.
+                second_expr = CastExpression(int_empty_data, second_expr)
+            if nanosecond_expr:
+                nanosecond_expr = int_part_expr(nanosecond_expr, "nanosecond_expr")
+
+            one_expr = ConstantExpression(int_empty_data, input_plan, 1)
+            zero_expr = ConstantExpression(int_empty_data, input_plan, 0)
+            twelve_expr = ConstantExpression(int_empty_data, input_plan, 12)
+
+            # Timestamp components (aside from the year) can be outside the usual ranges
+            # or even negative, so we must account for this and adjust the constructed timestamp.
+
+            # Calculate the year and month manually to ensure calendar awareness.
+
+            # Calculate month component.
+            # If (month-1) % 12 is 0 or positive, month component is (month-1) % 12 + 1.
+            # If (month-1) % 12 is negative, month component is 13 + (month-1) % 12.
+            month_minus_one = ArithOpExpression(
+                int_empty_data, month_expr, one_expr, "__sub__"
+            )
+            # We rely on bodo_mod returning a negative remainder when (month-1) is negative
+            month_remainder = ArithOpExpression(
+                int_empty_data, month_minus_one, twelve_expr, "__mod__"
+            )
+            month_remainder_negative = ComparisonOpExpression(
+                bool_empty_data, month_remainder, zero_expr, operator.lt
+            )
+            month_num_to_add = CaseExpression(
+                int_empty_data,
+                month_remainder_negative,
+                ConstantExpression(int_empty_data, input_plan, 13),
+                one_expr,
+            )
+            month_component = ArithOpExpression(
+                int_empty_data, month_remainder, month_num_to_add, "__add__"
+            )
+
+            # Calculate year component.
+            # If month is positive, year component is year + (month-1) // 12.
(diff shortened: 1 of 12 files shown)


BodoSQL/bodosql/plan_conversion.py
```diff
@@ -335,6 +335,429 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+            # Only use nanosecond precision if necessary (nanoseconds provided).
+            # This helps avoid overflow in some cases, and can allow us to represent more years.
+            use_nanosecond_precision = nanosecond_expr is not None
+            precision_str = "ns" if use_nanosecond_precision else "s"
+            timestamp_empty_data = pd.Series(
+                dtype=pd.ArrowDtype(pa.timestamp(precision_str))
+            )
+            duration_empty_data = pd.Series(
+                dtype=pd.ArrowDtype(pa.duration(precision_str))
+            )
```
```suggestion
```

bodo/pandas/physical/expression.cpp
```diff
@@ -629,11 +629,45 @@ arrow::Datum do_arrow_compute_cast(arrow::Datum left_res,
 
     // No need to cast if type is already the target type
     if (left_res.type()->Equals(arrow_ret_type)) {
-        return left_res;
+        if (arrow_ret_type->id() == arrow::Type::TIMESTAMP) {
```
IsaacWarren: I don't understand this change, when does it not return left_res?

bodo/pandas/plan_optimizer.pyx
```diff
@@ -743,6 +744,7 @@ cdef class PythonScalarFuncExpression(Expression):
 cdef class ArrowScalarFuncExpression(Expression):
     """Wrapper around DuckDB's BoundFunctionExpression for running Python/Arrow functions.
     """
+    cdef readonly function_name
```
IsaacWarren, line 747: Why do we need to store this now?

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -335,6 +335,429 @@ def java_call_to_python_call(ctx, java_call, input_plan):
             date_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.date32()))
             return CastExpression(date_empty_data, timestamp_expr)
 
+        def timestamp_from_parts(
```
IsaacWarren: This looks pretty complicated, how extensive are our existing tests for this?

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -849,6 +1286,138 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+            # It is okay if the difference is negative.
+            # In this case, month_interval_between would have counted a month that was not a full elapsed month between the dates.
+            # Therefore, adding a negative fraction to the integer month interval would just take away that extra whole month,
+            # which is what we want.
+
+            # Divide by 31 days to get the fraction of a month
+            # 31 days = 2_678_000_000_000_000 nanoseconds
+            month_nanos = ConstantExpression(
+                int_empty_data, input_plan, 2_678_000_000_000_000
+            )
```
IsaacWarren, line 1405: Is always using 31 days OK if one of the months is February? Do we have a test for this case?

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -335,6 +335,429 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+
+            # Calculate year component.
+            # If month is positive, year component is year + (month-1) // 12.
+            # If month is 0 or negative, year component is year + (month) // 12 - 1.
+            month_positive = ComparisonOpExpression(
+                bool_empty_data, month_expr, zero_expr, operator.gt
+            )
+            month_quotient = ArithOpExpression(
+                int_empty_data, month_expr, twelve_expr, "__floordiv__"
```
IsaacWarren: Lets add a comment that we're relying on duckdb's int division to be C style trunc towards 0, not true floor division

IsaacWarren (commented):
Thanks @dmsa-or!

bodo/pandas/physical/expression.cpp
dmsa-or: We don't return `left_res` if `left_res.type()` and `arrow_ret_type` are the same temporal type but have different units. In this case casting is still necessary to convert between the units.

bodo/pandas/plan_optimizer.pyx
dmsa-or: `__str__` of `ArrowScalarFuncExpression` uses `self.function_name` but it wasn't a member of the class. I encountered that while doing some debugging so I assume that code path wasn't tested until now.

BodoSQL/bodosql/plan_conversion.py
dmsa-or: They are quite extensive I think. I know this function looks complicated but I'm pretty confident it is correct. I tried to add lots of comments but let me know if there is anything that is still confusing.

BodoSQL/bodosql/plan_conversion.py
dmsa-or: Yes, it should always be 31 days according to Snowflake.

bodo/pandas/physical/expression.cpp
```diff
@@ -629,11 +629,48 @@ arrow::Datum do_arrow_compute_cast(arrow::Datum left_res,
...
+            auto left_timestamp_type =
+                std::static_pointer_cast<arrow::TimestampType>(left_res.type());
+            auto ret_timestamp_type =
+                std::static_pointer_cast<arrow::TimestampType>(arrow_ret_type);
+            if (left_timestamp_type->unit() == ret_timestamp_type->unit()) {
+                if (left_timestamp_type->timezone() ==
+                    ret_timestamp_type->timezone()) {
```
scott-routledge2:
```suggestion
            if (left_timestamp_type->unit() == ret_timestamp_type->unit() &&   
                left_timestamp_type->timezone() == ret_timestamp_type->timezone()) {
```

bodo/pandas/_util.cpp
```diff
@@ -1476,15 +1476,52 @@ const char *get_py_single_arg_as_cstr(PyObject *args, const char *func_name) {
...
-int64_t get_py_round_arg(PyObject *args) {
+std::tuple<int64_t, arrow::compute::RoundMode> get_py_round_args(
+    PyObject *args) {
     int64_t digits = 0;  // default value if no argument is provided
-    if (PyTuple_Check(args) && PyTuple_Size(args) == 1) {
-        // Get the first element (borrowed reference)
-        PyObject *py_digits = PyTuple_GetItem(args, 0);
-
-        digits = get_py_object_as_int64(py_digits, "round args element");
+    // half_to_even is the default round mode in Arrow; we could change it if we
```
scott-routledge2: This is also the behavior of builtin `round()` function, so this default makes sense to me

BodoSQL/bodosql/plan_conversion.py
scott-routledge2: We could also pull this function out and add a unti test if we still feel like it's necessary (pulling this function out to the toplevel is also better for readability for me). 

Also, adding a docstring with the general approach of this function, i.e. what the output plan will look like, what arrow functions are used as workaround would be helpful here.

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -335,6 +335,427 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+
+            max_int64 = ConstantExpression(
+                int_empty_data, input_plan, np.iinfo(np.int64).max
+            )
+            min_int64 = ConstantExpression(
+                int_empty_data, input_plan, np.iinfo(np.int64).min
+            )
+
+            # If any of the inputs are floats, we will round to the nearest integer.
+            # This is what the JIT implementation seems to do, even though float input isn't explicitly allowed in Snowflake.
```
scott-routledge2: Link to JIT implementation here?

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -335,6 +335,427 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+                        [part_expr],
+                        "round",
+                        (0, "half_towards_infinity"),
+                    )
+
+                    # The BodoSQL docs say we should return NULL when any input cannot be converted to int64
+                    part_too_large = ComparisonOpExpression(
+                        bool_empty_data, part_expr, max_int64, operator.gt
+                    )
+                    part_too_small = ComparisonOpExpression(
```
scott-routledge2: Do we need to upcast int32 before comparing to int64 max here? 

I think if we do not, the int64 max constant will be cast to int32.

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -335,6 +335,427 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+                int_empty_data, year_expr, year_offset, "__add__"
+            )
+
+            # Concatenate date components to get a string that can be parsed by strptime
+            year_string = CastExpression(string_empty_data, year_component)
+            year_string = ArrowScalarFuncExpression(
+                string_empty_data, [year_string], "utf8_lpad", (4, "0")
+            )
+            month_string = CastExpression(string_empty_data, month_component)
+            month_string = ArrowScalarFuncExpression(
```
scott-routledge2: Do we have to handle negative months here?

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -335,6 +335,427 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+            month_expr = int_part_expr(month_expr, "month_expr")
+            day_expr = int_part_expr(day_expr, "day_expr")
+
+            if hour_expr:
+                hour_expr = int_part_expr(hour_expr, "hour_expr")
+            if minute_expr:
+                minute_expr = int_part_expr(minute_expr, "minute_expr")
+            if second_expr:
+                second_expr = int_part_expr(second_expr, "second_expr")
+                # Ensure seconds are int64 so the result of the later multiplication doesn't
```
scott-routledge2: Do we also need to upcast minutes/hours expressions here too?

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -335,6 +335,427 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+            )
+            return CastExpression(
+                pd.Series(dtype=pd.ArrowDtype(pa.date32())), constructed_timestamp
+            )
+
+        if func_name == "TIME_FROM_PARTS" and num_operands in (3, 4):
+            int_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.int64()))
+
+            # If any of the inputs are floats, we will round to the nearest integer.
+            # This is what the JIT implementation seems to do, even though float input isn't explicitly allowed in Snowflake.
```
scott-routledge2, line 750: linking to JIT implementation for reference would be helpful

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -335,6 +335,427 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+                        "round",
+                        (0, "half_towards_infinity"),
+                    )
+                    part_expr = CastExpression(int_empty_data, part_expr)
+                return part_expr
+
+            hour_expr = int_part_expr(0, "hour_expr")
+            minute_expr = int_part_expr(1, "minute_expr")
+            second_expr = int_part_expr(2, "second_expr")
+            # Ensure seconds are int64 so the result of the later multiplication doesn't
```
scott-routledge2, line 771: Same as above, not sure why we only need to do the casting for seconds expr?

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -848,6 +1283,138 @@ def java_call_to_python_call(ctx, java_call, input_plan):
...
+                bool_empty_data, date1_day, date1_last_day, operator.eq
+            )
+            is_date2_on_last_day = ComparisonOpExpression(
+                bool_empty_data, date2_day, date2_last_day, operator.eq
+            )
+            both_dates_on_last_day = ConjunctionOpExpression(
+                bool_empty_data, is_date1_on_last_day, is_date2_on_last_day, "__and__"
+            )
+
+            # Is it a special case where we are supposed to use the raw month interval?
```
scott-routledge2, line 1463:
```suggestion
```

scott-routledge2 (approved):
Thanks @dmsa-or!

BodoSQL/bodosql/plan_conversion.py
dmsa-or: This was a doc/explanation comment but maybe I could rephrase it.

scott-routledge2: Oh I misread it originally. This is just explaining what the case expression is doing so that makes sense.

BodoSQL/bodosql/plan_conversion.py
dmsa-or: Perhaps if you wanted to be 100% safe. I found that it wasn't a problem for hours or minutes because the factors 3_600_000_000_000 and 60_000_000_000 respectively are already greater than the max value of int32. Due to that it doesn't try to cast the factor to int32 and instead upcasts the hours and minutes. That's just what I've found experimentally - I'm not sure exactly where this casting is happening.

BodoSQL/bodosql/plan_conversion.py
dmsa-or: That's handled when we calculate `year_component` and `month_component` above.

BodoSQL/bodosql/plan_conversion.py
dmsa-or: I don't know if it's required from what I've seen. As I said it seems to attempt a safe cast to int32 and will fallback to upcasting to int64. Isaac has said that Arrow casting is slow so I was trying to avoid unnecessary cast expressions, but I can add one if you think I should.

scott-routledge2: I meant that when we perform a comparison, we cast the constant to the type that's on the left (this seems weird to me but that's a separate issue). 

You're right that the constant will be an int64, but it will be cast to whatever the float type is in the comparison op, which should always succeed. I missed that this check is only happening for floats, so you can ignore my above comment. 

 Also, should the operator be ge and le instead of lt and gt?

dmsa-or: Thanks for the correction, I assumed it was using the same logic as for arith operations. Yeah, I also forgot we were dealing with a float there.

Why would the operator need to be ge or le? Wouldn't you be able to cast to int64 if it was equal to `max_int64`? Or is there some weirdness due to `max_int64` being casted to a float?

scott-routledge2: Actually lt, gt is right, I had the logic backwards.

bodo/pandas/_util.cpp
dmsa-or: It seems like the default rounding mode in SQL is usually `half_towards_infinity`/`half_away_from_zero`, including in Snowflake (and BodoSQL, according to [this](https://docs.bodo.ai/latest/api_docs/sql/functions/numeric/round/)) so that's the only reason we might want to change it.

scott-routledge2 (approved):
Thanks @dmsa-or

PR #1273, opened 2026-07-07
BSE-5484: Deterministic Groupby Combination

IsaacWarren:
We saw an incorrect result caused by floats being summed in different orders across ranks so they didn't match join keys. This PR makes groupby aggregate per rank (MPI guarantees order within ranks) and then across ranks from 0 .. n at finalization.

## Testing strategy
PyDough TPCH q15 and PR CI

## Checklist


bodo/libs/streaming/_join.cpp
IsaacWarren, line 1: The only change to this file is just looping over a vector of tables instead of being guaranteed a single table based on the new shuffle return signature

PR #1278, opened 2026-07-09
Fix all-NA sum for SQL

ehsantn:
Makes sure SQL rules are used for sum and product reductions. Fixes `test_multi_col_null_pivot` on nightly. Will open follow ups for other reductions and GPU case.


DrTodd13 (approved):
thanks

scott-routledge2 (approved):
Thanks @ehsantn !

PR #1279, opened 2026-07-09
Support create table.

DrTodd13:

Implement calcite create table conversion.

## Testing strategy

ran pydough test  tests/test_pipeline_bodosql.py::test_bodosql_e2e_defog[broker_adv1] locally

## User facing changes

None

## Checklist

### DIFF
--- a/BodoSQL/bodosql/context.py
+++ b/BodoSQL/bodosql/context.py
@@ -732,7 +732,7 @@ class BodoSQLContext:
                 out = bodo.pandas.plan.execute_plan(
                     plan.internal_plan, optimize=False, use_sql_rules=True
                 )
-                location = os.path.join(*plan.table_create_node.getPath())
+                location = self.catalog.connection_string
                 out.to_iceberg(plan.table_create_node.getTableName(), location=location)
                 out = None
             else:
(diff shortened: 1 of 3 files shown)


IsaacWarren (approved):

BodoSQL/bodosql/context.py
```diff
@@ -723,9 +728,16 @@ def execute_cpp_backend(
             # filter translation during conversion to Python plan.
             self.join_filter_info = {}
             plan = java_plan_to_python_plan(self, java_plan)
-            out = bodo.pandas.plan.execute_plan(
-                plan, optimize=False, use_sql_rules=True
-            )
+            if isinstance(plan, self.NewTable):
+                out = bodo.pandas.plan.execute_plan(
+                    plan.internal_plan, optimize=False, use_sql_rules=True
```
scott-routledge2, line 733: Why are we executing the internal plan here and converting to an in-memory DataFrame as opposed to writing the table to storage?

DrTodd13: Fixed.

BodoSQL/bodosql/context.py
```diff
@@ -723,9 +728,17 @@ def execute_cpp_backend(
...
+                )
+                location = self.catalog.connection_string
+                out.to_iceberg(plan.table_create_node.getTableName(), location=location)
```
scott-routledge2, line 736: This will work for FileSystemCatalog, but I think it will throw an error (and then fallback) for other catalogs. 

This is ok for now since it matches what we have on the read side for the C++ backend, but let's open a followup to support other catalogs for write.

scott-routledge2 (approved):
Thanks @DrTodd13 !

PR #1281, opened 2026-07-09
Support convert_timezone.

DrTodd13:

Support convert_timezone calcite conversion to bodo dataframe.

## Testing strategy

tested locally

## User facing changes

None

## Checklist

### DIFF
--- a/BodoSQL/bodosql/plan_conversion.py
+++ b/BodoSQL/bodosql/plan_conversion.py
@@ -3491,12 +3491,43 @@ def java_call_to_python_call(ctx, java_call, input_plan):
                 "src",
                 (
                     pd._libs.tslibs.timestamps.Timestamp,
-                    pd.ArrowDtype(pa.timestamp("ns", tz=str_timezone.value)),
+                    pd.ArrowDtype(pa.timestamp("ns")),
                 ),
             )
+            timestamp_pa_type = src.empty_data.iloc[:, 0].dtype.pyarrow_dtype
+            target_res = "ns"
+            # We figure out the right resolution to use although right now in Bodo
+            # there are multiple paths that force ns resolution.
+            if pa.types.is_date(timestamp_pa_type):
+                target_res = "s"
+            elif pa.types.is_timestamp(timestamp_pa_type):
+                target_res = timestamp_pa_type.unit
+
+            if pa.types.is_date(timestamp_pa_type) or (
+                pa.types.is_timestamp(timestamp_pa_type)
+                and timestamp_pa_type.tz is None
+            ):
+                tz = ctx.default_tz if ctx.default_tz is not None else "UTC"
+                local_timestamp_empty_data = pd.Series(
+                    dtype=pd.ArrowDtype(pa.timestamp(target_res, tz=tz))
+                )
+                timestamp_empty_data_no_tz = pd.Series(
+                    dtype=pd.ArrowDtype(pa.timestamp(target_res))
+                )
+                # We first cast to a data type that assume_timezone can work with.
+                src = CastExpression(timestamp_empty_data_no_tz, src)
+                # We use the context default_tz to make the timezone explicit.
+                src = ArrowScalarFuncExpression(
+                    local_timestamp_empty_data,
+                    [src],
+                    "assume_timezone",
+                    (tz,),
+                )
+
             target_timestamp_empty_data = pd.Series(
-                dtype=pd.ArrowDtype(pa.timestamp("ns", tz=str_timezone.value))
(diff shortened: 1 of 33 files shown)


BodoSQL/bodosql/plan_conversion.py
```diff
@@ -3209,6 +3217,19 @@ def clean_regex_params(regex_params_expr):
...
                 ),
             )
+        elif func_name == "CONVERT_TIMEZONE" and len(op_exprs) == 2:
+            str_timezone = op_exprs[0]
+            src = op_exprs[1]
+
+            ensure_arg_is_const_expr_of_type(
+                str_timezone, "str_timezone", (str, pa.binary())
+            )
+            ensure_type_of_expr(src, "src", pd._libs.tslibs.timestamps.Timestamp)
```
dmsa-or: This won't work when the expression type is another kind of timestamp. We either need better timestamp comparison in `compare_types` or change `get_expr_dtype` to always get the empty_data type instead of the value type for `ConstantExpression`s.

DrTodd13: added one additional type.  we'll add more as we encounter them.

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -3479,6 +3479,26 @@ def clean_regex_params(regex_params_expr):
...
+
+            ensure_arg_is_const_expr_of_type(
+                str_timezone, "str_timezone", (str, pa.binary())
+            )
+            ensure_type_of_expr(
+                src,
+                "src",
+                (
+                    pd._libs.tslibs.timestamps.Timestamp,
+                    pd.ArrowDtype(pa.timestamp("ns", tz=str_timezone.value)),
```
dmsa-or: `tz=str_timezone.value` is checking that timezone of the input timestamp is the timezone you want to convert to, which isn't correct. Although this may technically still work given that `compare_types` converts this to a Numpy dtype and loses TZ information.

dmsa-or (commented):
Can you run `BodoSQL/bodosql/tests/test_conversions/test_snowflake_date_conversion_fns.py::test_convert_timezone` to make sure it passes?

BodoSQL/bodosql/plan_conversion.py
DrTodd13: fixed

ehsantn (commented):
Are you sure simply casting is enough for this function? The JIT code is not this trivial:
https://github.com/bodo-ai/Bodo/blob/aa946e9424119b0062d3465e83b81b479b3f8867/BodoSQL/bodosql/kernels/snowflake_conversion_array_kernels.py#L2433

Tried `test_convert_timezone` on this branch and it fails.

DrTodd13:
> Are you sure simply casting is enough for this function? The JIT code is not this trivial:
> 
> https://github.com/bodo-ai/Bodo/blob/aa946e9424119b0062d3465e83b81b479b3f8867/BodoSQL/bodosql/kernels/snowflake_conversion_array_kernels.py#L2433
> 
> Tried `test_convert_timezone` on this branch and it fails.

We determined we'd hold off on the 3 argument version for now.  There are 2 2-argument tests.  The first one uses an unsupported type for conversion to pyarrow and seems like too much effort.  The second one should work and did illustrate a bug.  The code has been improved and Daniel and I verified it gets the right answer but it still won't pass the test since we have auto conversion to nanosecond all over the place which prints extra fractional second data when converted to a string and so it doesn't match the test's hardcoded string expected output which doesn't include fractional seconds.

dmsa-or (approved):
Looks good to me. Supporting this function has shown that we need better handling of different timestamp types throughout the code and better time unit preservation.

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -3479,6 +3479,57 @@ def clean_regex_params(regex_params_expr):
...
+            timestamp_pa_type = src.empty_data.iloc[:, 0].dtype.pyarrow_dtype
+            target_res = "ns"
+            # We figure out the right resolution to use although right now in Bodo
+            # there are multiple paths that force ns resolution.
+            if pa.types.is_date(timestamp_pa_type):
+                target_res = "s"
+            elif pa.types.is_timestamp(timestamp_pa_type):
+                target_res = timestamp_pa_type.unit
+
+            if pa.types.is_date(timestamp_pa_type) or (
```
ehsantn: Let's add a comment on why adding a timezone is necessary here.

ehsantn (approved):
Thanks for the fixes.

PR #1282, opened 2026-07-10
Support epoch date/time functions in BodoSQL C++ backend

dmsa-or:

All the MySQL epoch functions in `test_epoch_datetime_fns.py` are now supported and enabled on the C++ backend.

## Testing strategy

CI

## Checklist

### DIFF
--- a/BodoSQL/bodosql/tests/test_epoch_datetime_fns.py
+++ b/BodoSQL/bodosql/tests/test_epoch_datetime_fns.py
@@ -4,12 +4,13 @@ Test correctness of SQL queries that are dependent on time since year 0, or the 
 
 import numpy as np
 import pandas as pd
+import pytest
 
 from bodo.tests.utils import pytest_slow_unless_codegen
 from bodosql.tests.utils import check_query
 
 # Skip unless any codegen files were changed
-pytestmark = pytest_slow_unless_codegen
+pytestmark = pytest_slow_unless_codegen + [pytest.mark.bodosql_cpp]
 
 
 # the difference in days between the unix epoch and the start of year 0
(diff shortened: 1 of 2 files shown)


IsaacWarren (approved):
Thanks @dmsa-or!

BodoSQL/bodosql/tests/test_epoch_datetime_fns.py
```diff
@@ -168,10 +169,13 @@ def test_unix_timestamp(basic_df, memory_leak_check):
...
     # right as the ten thousandths place changes values
     query = "SELECT A, Floor(UNIX_TIMESTAMP() / 10000) as output from table1"
+    tz = "America/Los_Angeles"
     expected_output = pd.DataFrame(
         {
             "A": basic_df["TABLE1"]["A"],
-            "OUTPUT": np.float64(pd.Timestamp.now().value // (10000 * 1000000000)),
+            # pd.Timestamp.now() needs a timezone passed in, else it will be
+            # timezone-naive in the local timezone.
+            "OUTPUT": np.float64(pd.Timestamp.now(tz=tz).value // (10000 * 1000000000)),
```
ehsantn, line 178: What was the problem with this test? Was it the default timezone in BodoSQL being different from this expected output?

ehsantn (approved):
Thanks @dmsa-or.

BodoSQL/bodosql/tests/test_epoch_datetime_fns.py
dmsa-or: The actual timezone doesn't matter as long as it is timezone-aware so the timestamp value is in UTC. The issue is that pd.Timetamp.now() without the `tz` argument returns a local TZ-naive timestamp in my local timezone. I'm guessing there is a configuration difference that allows this to pass in CI (local timezone is UTC, probably).

PR #1291, opened 2026-07-15
Support more numeric functions in C++ backend [GPU]

dmsa-or:

Support all numeric functions in `test_numeric_fns.py` except for `HAVERSINE`, `FACTORIAL`, `UNIFORM`, `CONV`, and the `TO_NUMBER`-type functions. Several are supported on GPU as well.

## Testing strategy

PR CI and Nightly

## Checklist

### DIFF
--- a/BodoSQL/bodosql/plan_conversion.py
+++ b/BodoSQL/bodosql/plan_conversion.py
@@ -3212,12 +3212,10 @@ def java_call_to_python_call(ctx, java_call, input_plan):
             out_empty = left.empty_data.iloc[:, 0] ** right.empty_data.iloc[:, 0]
             return ArithOpExpression(out_empty, left, right, "__pow__")
         elif func_name == "SQUARE" and len(op_exprs) == 1:
-            left = op_exprs[0]
-            out_empty = left.empty_data.iloc[:, 0] ** 2
-            exponent_expr = ConstantExpression(
-                pd.Series(dtype=pd.ArrowDtype(pa.int64())), input_plan, 2
-            )
-            return ArithOpExpression(out_empty, left, exponent_expr, "__pow__")
+            inp = op_exprs[0]
+            ensure_type_of_expr(inp, "SQUARE input", (int, float))
+            out_empty = inp.empty_data.iloc[:, 0] * inp.empty_data.iloc[:, 0]
+            return ArithOpExpression(out_empty, inp, inp, "__mul__")
         elif func_name == "LOG2" and len(op_exprs) == 1:
             inp = op_exprs[0]
             ensure_type_of_expr(inp, "LOG2 input", (int, float))
@@ -4200,6 +4198,9 @@ def java_call_to_python_call(ctx, java_call, input_plan):
         op_exprs = [java_expr_to_python_expr(ctx, o, input_plan) for o in operands]
         func_name = op.getName().upper()
 
+        # NOTE: Calcite maps SQL RANDOM() to RANDOM() which means a random number
+        # between [0.0, 1.0] in Calcite and does not accept a seed argument. For
+        # completeness, to match Snowflake, we support a seed parameter anyway.
         if func_name == "RANDOM" and len(op_exprs) in (0, 1):
             """Generates random 64-bit integers"""
             int_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.int64()))
--- a/BodoSQL/bodosql/plan_conversion.py
+++ b/BodoSQL/bodosql/plan_conversion.py
@@ -4545,22 +4545,6 @@ def get_common_int_type_list(
         common_type = eval(
             f"pa.{'' if common_type_signed else 'u'}int{common_type_width}()"
         )
-        """
-        if common_is_signed == expr_is_signed:
-            # Use wider type if inputs have same signedness
-            if expr_width > common_width:
-                common_type = dtype
-        else:
-            # For mixed signedness, use the wider bit width of the two.
-            # The common type is signed only if the max value of the unsigned input can fit in the signed int
-            common_type_width = max(common_width, expr_width)
-            common_type_signed = (common_is_signed and common_width > expr_width) or (
-                expr_is_signed and expr_width > common_width
-            )
-            common_type = eval(
-                f"pa.{'' if common_type_signed else 'u'}int{common_type_width}()"
-            )
-        """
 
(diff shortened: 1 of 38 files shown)


bodo/pandas/physical/expression.h
```diff
@@ -1632,6 +1659,106 @@ class PhysicalArrowExpression : public PhysicalExpression {
...
                 result = ConvertDatumToArrayInfo(if_else_res.ValueOrDie());
             }
+        } else if (scalar_func_data.arrow_func_name == "rand") {
+            // Get dummy input as an Arrow array.
+            // We need the result to match the length of this array.
+            std::shared_ptr<arrow::Array> res_array =
+                ConvertExprResultToDatum(res, "rand input").make_array();
+
+            // Retrieve the function object for random()
+            auto *registry = arrow::compute::GetFunctionRegistry();
```
DrTodd13: Consider making the lookups here call_once so we don't pay the price every batch.  Is there a reason you'd want totally separate seeds per batch?

bodo/pandas/physical/expression.h
```diff
@@ -1632,6 +1659,106 @@ class PhysicalArrowExpression : public PhysicalExpression {
...
+
+            std::mt19937_64 gen;
+            if (num_args == 1) {
+                // Seed was explicitly provided, so use it
+                auto [seed] = get_py_args_as_types(
+                    scalar_func_data.args,
+                    scalar_func_data.arrow_func_name.c_str(),
+                    get_py_object_as_int64);
+                // Create psuedo-random number generator
+                gen = std::mt19937_64(seed);
```
DrTodd13: There may be a big problem here.  If they give a seed and you start from that seed for every batch then you are going to get duplicate sequences of values which aren't random.  Seems like the generator somehow has to persist across batches and we can't use the same seed on all workers as that will also duplicate sequences.  This is a very complicated area to get right. We'll need to look up the suggested formula for how to modify the seed based on rank.

bodo/pandas/physical/expression.h
```diff
@@ -1632,6 +1659,106 @@ class PhysicalArrowExpression : public PhysicalExpression {
...
+            } else if (num_args == 0) {
+                // Generate a 96-bit seed from system's random_device and the
+                // curent system time. time(NULL) is mainly a backup in case
+                // std::random_device falls back to a deterministic
+                // implementation.
+                std::random_device rd;
+                std::seed_seq seed{rd(), static_cast<unsigned int>(time(NULL)),
+                                   rd()};
+                // Create psuedo-random number generator
+                gen = std::mt19937_64(seed);
```
DrTodd13: Suppose some chance multiple workers get same seed here as well so should also do the same modification as above with rank.

bodo/pandas/physical/gpu_expression.h
```diff
@@ -754,6 +761,94 @@ class PhysicalGPUUnaryExpression : public PhysicalGPUExpression {
...
+                cudf::data_type left_type = left_as_array->result->type();
+                cudf::data_type return_type;
+                if (cudf::is_floating_point(left_type)) {
+                    return_type = left_type;
+                } else {
+                    return_type = cudf::data_type{cudf::type_id::INT8};
+                }
+
+                // Computes is_positive - is_negative.
+                // True is treated as 1 and False as 0. Hence:
```
DrTodd13, line 841: Mention -1?

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -2408,67 +2408,135 @@ def cur_local_timestamp():
             dummy_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.duration("ns")))
             return ConstantExpression(dummy_empty_data, input_plan, combined_val)
 
+        if func_name == "SIGN" and len(op_exprs) == 1:
+            inp = op_exprs[0]
+            ensure_type_of_expr(inp, func_name + " input", (int, float))
+
+            # To match Arrow's sign(), we return int8 if input is an integer,
```
DrTodd13: Don't understand this comment.  Why are we matching arrow's sign instead of sql's?

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -3003,19 +3087,345 @@ def clean_regex_params(regex_params_expr):
...
             out_empty = left.empty_data.iloc[:, 0] ** right.empty_data.iloc[:, 0]
             return ArithOpExpression(out_empty, left, right, "__pow__")
-        elif func_name == "FLOOR" and len(op_exprs) == 1:
+        elif func_name == "SQUARE" and len(op_exprs) == 1:
+            left = op_exprs[0]
+            out_empty = left.empty_data.iloc[:, 0] ** 2
+            exponent_expr = ConstantExpression(
+                pd.Series(dtype=pd.ArrowDtype(pa.int64())), input_plan, 2
+            )
+            return ArithOpExpression(out_empty, left, exponent_expr, "__pow__")
```
DrTodd13: Do x*x instead of x**2.

DrTodd13 (changes requested):
Thanks.  Try to make a bit smaller PRs.  Also, think we should probably start to make separate functions for the various kinds of function classes because that function is getting huge at this point.  That last point is just an observation.  Don't think such refactoring should be included in this PR.

bodo/pandas/physical/expression.h
dmsa-or: Yeah, I thought about this. I don't know if the JIT implementation handles it either because the docs say this: `Note that running with multiple processors may affect the randomization results.`

BodoSQL/bodosql/plan_conversion.py
dmsa-or: I don't think SQL specifies what the return type should be, so I just decided to match what Arrow returns since that's what we're calling.

bodo/pandas/physical/gpu_expression.h
dmsa-or: What do you mean?

DrTodd13 (approved):
Thanks for the updates.

BodoSQL/bodosql/plan_conversion.py
ehsantn: What about matching the return type in the SQL plan here? Another operator may potentially expect a different type from this operator as input, leading to errors.

ehsantn (approved):
Thanks @dmsa-or !

PR #1293, opened 2026-07-15
Fix ANY_VALUE with nulls test on nightly

scott-routledge2:

We recently updated ANY_VALUE to run on C++ backend, which returns the first non-null value (or null if all values are null). 

## Checklist

### DIFF
--- a/BodoSQL/bodosql/tests/test_agg_nogroupby.py
+++ b/BodoSQL/bodosql/tests/test_agg_nogroupby.py
@@ -13,13 +13,12 @@ from bodo.tests.utils import (
     check_func,
     count_array_REPs,
     dist_IR_contains,
-    pytest_slow_unless_groupby,
 )
 from bodo.tests.utils_jit import DistTestPipeline
 from bodosql.tests.utils import check_query
 
 # Skip unless any groupby-related files were changed
-pytestmark = pytest_slow_unless_groupby
+# pytestmark = pytest_slow_unless_groupby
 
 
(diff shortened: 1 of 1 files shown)


BodoSQL/bodosql/tests/test_agg_nogroupby.py
```diff
@@ -587,7 +587,7 @@ def test_any_value_nulls(memory_leak_check):
         check_names=False,
         is_out_distributed=False,
         expected_output=pd.DataFrame(
-            {0: pd.array([None], dtype=pd.ArrowDtype(pa.int64()))}
+            {0: pd.array([1], dtype=pd.ArrowDtype(pa.int64()))}
```
DrTodd13, line 590: Whole point of function is to tests nulls so can we really change to 1 here?

scott-routledge2: According to the issue mentioned in this test https://bodo.atlassian.net/browse/BSE-2934?search_id=566a0b89-7d40-4ac1-8361-54b85b2a2c7b 1 should also be a valid output of ANY_VALUE. 

The test was designed to make sure the presence of null in the first value does not lead to an output that wasn't present in the input column.

DrTodd13: Oops.  Sorry..  Misread it.

DrTodd13 (approved):

ehsantn (approved):

PR #1294, opened 2026-07-15
Use JIT kernels for next_day/previous_day with tz-aware data

ehsantn:

As title. This is an example to hopefully follow for other kernels.

## Testing strategy

Enabled unit tests.

## User facing changes

None.

## Checklist

### DIFF
--- a/BodoSQL/bodosql/kernels/datetime_array_kernels.py
+++ b/BodoSQL/bodosql/kernels/datetime_array_kernels.py
@@ -2540,6 +2540,36 @@ def next_day_util(arr0, arr1):
     )
 
 
+def next_day_wrapper(df):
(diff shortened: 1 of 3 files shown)


DrTodd13 (commented):
Thanks!

DrTodd13 (approved):

dmsa-or (approved):
Any ballpark for what percentage of kernels should work with this wrapper? If it is most of them, we could have a general one and only make dedicated wrappers for the special cases.

ehsantn:
> Any ballpark for what percentage of kernels should work with this wrapper? If it is most of them, we could have a general one and only make dedicated wrappers for the special cases.

Creating a general wrapper needs generating a function from text which can potentially cause issues in caching. See comments in the code.

PR #1296, opened 2026-07-15
Named and dynamic params.

DrTodd13:

Mostly stuff for named and dynamic params but a fix for data types and casting for integer subtraction plus marking some other tests as passing.

## Testing strategy

run_ci

## User facing changes

None

## Checklist

### DIFF
--- a/BodoSQL/bodosql/plan_conversion.py
+++ b/BodoSQL/bodosql/plan_conversion.py
@@ -289,53 +289,43 @@ type_str_to_pa = {
 
 def java_dynamic_param_to_python_literal(ctx, dyn_param, input_plan):
     """Convert a BodoSQL Java dynamic param expression to a DataFrame library constant."""
-    dyn_param = dyn_param.toString()
-    if dyn_param.startswith("?"):
-        dyn_param = dyn_param[1:]
-        if not dyn_param.isnumeric():
-            raise ValueError("Dynamic parameter is not numeric.")
-        dyn_param = int(dyn_param)
+    dyn_param = dyn_param.getIndex()
+    if dyn_param >= len(ctx.dynamic_params_list[0]) or dyn_param >= len(
+        ctx.dynamic_params_list[1]
+    ):
+        raise ValueError(f"Dynamic parameter {dyn_param} not found.")
 
-        if dyn_param >= len(ctx.dynamic_params_list[0]) or dyn_param >= len(
-            ctx.dynamic_params_list[1]
-        ):
-            raise ValueError(f"Dynamic parameter {dyn_param} not found.")
-
-        column = ctx.dynamic_params_list[0][dyn_param]
-        value = ctx.dynamic_params_list[1][dyn_param]
-        if isinstance(value, np.generic):
-            value = value.item()
-        cdtype = column.getDataType().toString()
-        if cdtype in type_str_to_pa:
-            dummy_empty_data = pd.Series(dtype=pd.ArrowDtype(type_str_to_pa[cdtype]))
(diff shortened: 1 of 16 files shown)


BodoSQL/bodosql/plan_conversion.py
```diff
@@ -277,9 +261,83 @@ def java_expr_to_python_expr(ctx, java_expr, input_plan):
...
+    "FLOAT64": pa.float64(),
+    "STRING": pa.string(),
+    "BOOL8": pa.bool_(),
+    "TIMESTAMP_NTZ": pa.timestamp("ns"),
+}
+
+
+def java_dynamic_param_to_python_literal(ctx, dyn_param, input_plan):
+    """Convert a BodoSQL Java dynamic param expression to a DataFrame library constant."""
+    dyn_param = dyn_param.toString()
```
ehsantn: This function converts `dyn_param` and `column.getDataType()` to string before processing. Understanding and using Java types is more robust.

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -277,9 +261,83 @@ def java_expr_to_python_expr(ctx, java_expr, input_plan):
...
+        if cdtype in type_str_to_pa:
+            dummy_empty_data = pd.Series(dtype=pd.ArrowDtype(type_str_to_pa[cdtype]))
+            return ConstantExpression(dummy_empty_data, input_plan, value)
+
+    raise NotImplementedError("DynamicParam format or type not supported yet")
+
+
+def java_named_param_to_python_literal(ctx, named_param, input_plan):
+    """Convert a BodoSQL Java named param expression to a DataFrame library constant."""
+    named_param = named_param.toString()
```
ehsantn: Same.

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -277,9 +261,83 @@ def java_expr_to_python_expr(ctx, java_expr, input_plan):
...
+        column = ctx.named_params_dict[0][named_param]
+        value = ctx.named_params_dict[1][named_param]
+        if isinstance(value, np.generic):
+            value = value.item()
+        cdtype = column.getDataType().toString()
+        if cdtype in type_str_to_pa:
+            dummy_empty_data = pd.Series(dtype=pd.ArrowDtype(type_str_to_pa[cdtype]))
+            return ConstantExpression(dummy_empty_data, input_plan, value)
+        else:
+            print("cdtype not found:", cdtype)
```
ehsantn: Raise error instead of print.

ehsantn (commented):
See comments.

BodoSQL/bodosql/plan_conversion.py
DrTodd13: Actually, I did it this way in an attempt to be more robust.  There's been several cases where there is a critical flag that needs to be checked but doesn't seem to be accessible via normal methods but is visible in  some text field.  You don't know what you don't know and sure you could dig into code for a hour to see if there is any path to this object type other than @var but the benefit of the string approach is if there is another case you don't know about then it won't match and you won't silently do the wrong thing.  In any case, I switched it to use the param name and index functions and will operate on the assumption that is the only path to these objects.

BodoSQL/bodosql/plan_conversion.py
DrTodd13: done

ehsantn (approved):

scott-routledge2 (approved):
LGTM, thanks @DrTodd13 !

PR #1297, opened 2026-07-15
Support TZ-aware DATEDIFF in C++ backend

dmsa-or:

Support TZ-aware DATEDIFF in C++ backend and enable `test_timestampdiff_cols` in `test_datetime_difference_fns.py`.
Support the 2-argument form of DATEDIFF as described in BodoSQL docs. Not sure if there is a test for it.

The edge cases here are tricky because Snowflake's documentation is not clear about the rules for TIMESTAMP_TZ input. BodoSQL docs claim that the timestamps should always be converted to UTC, but that is evidently not happening in many cases. There appear to be bugs in the JIT implementation too that make DATEDIFF behave differently for different timestamp types (such as Pandas TZ-aware timestamps vs. Bodo TimestampTZ) and thus it is harder to tell what the intention was. On top of that, some of the hard-coded test expected outputs seem to be or incorrect or at least written around the buggy JIT implementation.

To lay out the rules decided in this PR:
- If both inputs are TZ-naive, no transformations happen.
- If exactly one input is TZ-aware, we assume that the other input is a local timestamp in the same timezone.
- If both inputs are TZ-aware, they must have the same timezone, else we throw an error. If we need to calculate the difference in units of time (e.g. hours), we first convert both timezones to UTC so that we can compensate for DST. If instead the requested unit is a date unit (e.g. days), we use the local timestamp, thus preserving the local date boundaries. 

## Testing strategy

CI and running tests locally

## Checklist

### DIFF
--- a/BodoSQL/bodosql/plan_conversion.py
+++ b/BodoSQL/bodosql/plan_conversion.py
@@ -1257,39 +1257,64 @@ def java_call_to_python_call(ctx, java_call, input_plan):
                         ),
                     ],
                 )
-        if func_name == "DATEDIFF" and num_operands == 3:
-            # TODO: May need TZ-aware support
+        if func_name == "DATEDIFF" and num_operands in (2, 3):
+            if num_operands == 3:
+                # First operand is a FLAG(unit) interval qualifier from the Java
+                # planner (e.g. FLAG(DAY), FLAG(MONTH)).
+                unit_str = get_java_symbol(java_call.getOperands()[0])
+                unit_str = standardize_java_time_unit(func_name, unit_str).upper()
+                assert unit_str in INTERVAL_UNIT_MAP, (
+                    f"Unsupported DATEDIFF unit: {unit_str}"
+                )
 
-            # First operand is a FLAG(unit) interval qualifier from the Java
-            # planner (e.g. FLAG(DAY), FLAG(MONTH)).
-            unit_str = get_java_symbol(java_call.getOperands()[0])
-            unit_str = standardize_java_time_unit(func_name, unit_str).upper()
-            assert unit_str in INTERVAL_UNIT_MAP, (
-                f"Unsupported DATEDIFF unit: {unit_str}"
-            )
-            # date_expr1 will be subtracted from date_expr2
-            date_expr1 = java_expr_to_python_expr(
-                ctx, java_call.getOperands()[1], input_plan
-            )
-            date_expr2 = java_expr_to_python_expr(
-                ctx, java_call.getOperands()[2], input_plan
+                # date_expr1 will be subtracted from date_expr2
+                date_expr1 = java_expr_to_python_expr(
+                    ctx, java_call.getOperands()[1], input_plan
+                )
+                date_expr2 = java_expr_to_python_expr(
+                    ctx, java_call.getOperands()[2], input_plan
+                )
+            else:
+                # According to BodoSQL docs:
+                # If only two arguments are provided, the unit should be days.
+                # Also, the order of the dates is reversed compared to the
+                # three argument version (second date is subtracted from first).
+                unit_str = "DAY"
+                date_expr2 = java_expr_to_python_expr(
+                    ctx, java_call.getOperands()[0], input_plan
+                )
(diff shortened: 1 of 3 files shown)


BodoSQL/bodosql/plan_conversion.py
```diff
@@ -1299,33 +1324,89 @@ def int_part_expr(op_index, part_expr_name):
...
+                        ),
+                        date_expr1,
+                    )
+                    date_expr2 = CastExpression(
+                        pd.Series(
+                            dtype=pd.ArrowDtype(pa.timestamp(date2_precision, tz="UTC"))
+                        ),
+                        date_expr2,
+                    )
+                # Don't convert timezone if we shouldn't get the answer in units of time.
```
ehsantn, line 1403: This comment isn't clear to me. Does Arrow work in this case correctly? You think JIT implementation is wrong? You think the test is wrong?

ehsantn (approved):
Thanks @dmsa-or. See question.

BodoSQL/bodosql/plan_conversion.py
dmsa-or: This is the main scenario where it's not clear what the intended behavior by looking at the JIT implementation and tests. The expected outputs of `test_timestampdiff_cols` (which is the test I enabled in this PR) are consistent with comparing the date parts of the local timestamps rather than the UTC timestamps, so that is what I chose to do. However I discovered another test (`test_timestamp_tz.py::test_timestamp_tz_datediff[day]`) that says we should always do the comparison in UTC, which is further complicated by a mistake in the test that ends up checking that the timestamps are NOT compared in UTC.

So to summarize:

- Not Arrow's fault - Arrow will predictably take the local timestamps of the inputs, so it's on us whether to convert timezones or not before passing to Arrow.
- As far as I can tell, for date units, the JIT implementation always effectively uses the local timestamps when they have timezones.
- There is a test that tries to verify that JIT implementation does compare the UTC-converted timestamps even for date units, and seems to magically pass only because of a lucky combination of mistakes in the JIT implementation and the expected output of the test.

BodoSQL/bodosql/tests/test_timestamp_tz.py
```diff
@@ -334,6 +334,15 @@ def test_timestamp_tz_dateadd(
             # is a different day from the local date - this tests that datediff
             # is comparing using the UTC date
             "TIMESTAMPDIFF(day, T, TO_TIMESTAMP_TZ('2024-01-01 00:00:00 +1234'))",
+            # FIXME: Are the expected outputs here correct?
+            # '2024-01-01 00:00:00 +1234' should correspond to 2024-01-01 00:00:00
+            # in local time, or 2023-12-31 11:26:00 UTC. So we would expect to be
+            # subtracting from datetime.date(2023, 12, 31).
+            # Also, the JIT implementation may not behave well when the timestamps
+            # have different UTC offsets. When the input is a bodo.types.TimestampTZ,
```
scott-routledge2, line 342: IIRC Timestamp TZ is a snowflake specific type where each row in a column can have a different offset from UTC-- it is basically a struct of {utc_time, offset}. I'm wondering if these tests were created using snowflake as a reference solution? Have we checked what they do?

BodoSQL/bodosql/plan_conversion.py
scott-routledge2: Looking at the docs for datediff, it looks like it just compares the units being subtracted, so I'd be curious to see what snowflake does in these cases. Maybe if the unit is days or higher it ignores the time/offset part?

scott-routledge2 (approved):
Thanks @dmsa-or !

BodoSQL/bodosql/tests/test_timestamp_tz.py
dmsa-or: We could settle this if we could try it out in Snowflake, but is it practical for me to do that?

I still think there is a bug with how the JIT implementation handles `bodo.types.TimestampTZ`, since it ignores the UTC offset part even if the unit is hours or smaller. That's very unintuitive so I would be surprised if Snowflake did the same thing.

PR #1300, opened 2026-07-16
Find the all flag and add distinct to set ops.

DrTodd13:

Look at all flag and add distinct if necessary.

## Testing strategy

run_ci

## User facing changes

None

## Checklist

### DIFF
--- a/BodoSQL/bodosql/plan_conversion.py
+++ b/BodoSQL/bodosql/plan_conversion.py
@@ -253,6 +253,25 @@ def java_union_to_python_union(ctx, java_plan):
         plan = LogicalSetOperation(
             input_plans[i].empty_data, plan, input_plans[i], "union all"
         )
+    # Can't see how to get the value of "all" out of the java_plan directly
+    # so use regex on the string conversion.
+    re_res = re.search(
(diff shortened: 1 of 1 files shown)


BodoSQL/bodosql/plan_conversion.py
```diff
@@ -254,6 +253,25 @@ def java_union_to_python_union(ctx, java_plan):
         plan = LogicalSetOperation(
             input_plans[i].empty_data, plan, input_plans[i], "union all"
         )
+    # Can't see how to get the value of "all" out of the java_plan directly
+    # so use regex on the string conversion.
+    re_res = re.search(
+        r"\ball\b\s*[:=]\s*(true|false)\b", java_plan.toString(), flags=re.IGNORECASE
```
dmsa-or, line 259: How confident are you in this regex pattern?

dmsa-or (approved):

BodoSQL/bodosql/plan_conversion.py
DrTodd13: I manually tested both cases so I know it works as is.  Tried to make it precise so that other things that might appear in the string don't get misinterpreted.  Guess I could check that there's only one such match.

scott-routledge2 (approved):
Thanks @DrTodd13 !

PR #1315, opened 2026-07-21
Support REGEXP_INSTR in BodoSQL C++ backend

dmsa-or:

REGEXP_INSTR is now supported in the C++ backend. We don't support case-insensitive matching, so there are a couple tests I couldn't enable because of that.

## Testing strategy

CI

## Checklist

### DIFF
--- a/BodoSQL/bodosql/plan_conversion.py
+++ b/BodoSQL/bodosql/plan_conversion.py
@@ -680,7 +680,7 @@ def java_call_to_python_call(ctx, java_call, input_plan):
         )
         return timestamp_expr
 
-    def clean_regex_params(regex_params_expr):
+    def clean_regex_params(regex_params_expr, func_name):
         """Verify the raw regex parameters passed into functions such as REGEXP_SUBSTR
         or REGEXP_INSTR and turn them into a standard form."""
         if "c" in regex_params_expr.value and "i" in regex_params_expr.value:
@@ -3107,8 +3107,10 @@ def java_call_to_python_call(ctx, java_call, input_plan):
 
                 if start_expr.value > 0:
                     start = start_expr.value - 1
+                elif start_expr.value == 0:
+                    start = 0
                 else:
-                    start = start_expr.value
+                    raise ValueError("Start index must be positive for REGEXP_SUBSTR.")
             else:
                 start = 0
 
@@ -3131,7 +3133,7 @@ def java_call_to_python_call(ctx, java_call, input_plan):
                 ensure_arg_is_const_expr_of_type(
                     regex_params_expr, "regex_params_expr", str
                 )
-                regex_params = clean_regex_params(regex_params_expr)
+                regex_params = clean_regex_params(regex_params_expr, func_name)
             else:
                 regex_params = "c"
 
@@ -3962,8 +3964,10 @@ def java_call_to_python_call(ctx, java_call, input_plan):
 
                 if start_expr.value > 0:
                     start = start_expr.value - 1
+                elif start_expr.value == 0:
+                    start = 0
                 else:
-                    start = start_expr.value
+                    raise ValueError("Start index must be positive for REGEXP_INSTR.")
             else:
                 start = 0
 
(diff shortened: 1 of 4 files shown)


BodoSQL/bodosql/plan_conversion.py
```diff
@@ -680,6 +680,38 @@ def int_part_expr(part_expr, part_expr_name):
         )
         return timestamp_expr
 
+    def clean_regex_params(regex_params_expr):
+        """Verify the raw regex parameters passed into functions such as REGEXP_SUBSTR
+        or REGEXP_INSTR and turn them into a standard form."""
+        if "c" in regex_params_expr.value and "i" in regex_params_expr.value:
```
IsaacWarren, line 686: Not sure I understand this, isn't "i" unsupported?
Why can we strip c if i is last and vice versa? Does the c++ backend use the other value by default if there's only one in the expr string?

IsaacWarren (approved):
One question but otherwise LGTM, thanks @dmsa-or

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -3949,6 +3949,165 @@ def get_common_signed(
...
+            ensure_arg_is_const_expr_of_type(regexp, "regexp", (str, pa.binary()))
+
+            if len(op_exprs) >= 3:
+                start_expr = op_exprs[2]
+                ensure_arg_is_const_expr_of_type(start_expr, "start_expr", int)
+
+                if start_expr.value > 0:
+                    start = start_expr.value - 1
+                else:
+                    start = start_expr.value
```
IsaacWarren: Are we expected to handle negative indices? If not we should throw an error I think

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -680,6 +680,38 @@ def int_part_expr(part_expr, part_expr_name):
...
+                "c" not in regex_params_expr.value
+                and "i" not in regex_params_expr.value
+            ):
+                regex_params = regex_params_expr.value + "c"
+            else:
+                regex_params = regex_params_expr.value
+        for character in regex_params:
+            if character not in ("c", "i", "m", "e", "s"):
+                raise ValueError(
+                    f"{func_name} regex parameter {character} does not exist"
```
IsaacWarren, line 707: I don't think func_name is in this scope?

BodoSQL/bodosql/plan_conversion.py
dmsa-or: The rule according to Snowflake is that if the regex params contain both "c" and "i", that one that applies is the one that appears latest in the regex params. So this code just strips out the redundant "c"s or "i"s so it's easier to identify if we are in case-sensitive mode or not. If "i" is remaining (case-insensitive mode), we throw an error because case-insensitive matching is unsupported.

BodoSQL/bodosql/plan_conversion.py
dmsa-or: Good point, looks like REGEXP_INSTR is not supposed to handle negative start positions.

bodo/pandas/physical/expression.h
```diff
@@ -1547,6 +1548,26 @@ class PhysicalArrowExpression : public PhysicalExpression {
...
+
+    template <typename T>
+    arrow::Datum do_arrow_compute_regexp_substr(T res, std::string pattern_str,
+                                                std::string regex_params_str,
+                                                int64_t group_to_extract) {
+        bool extract_submatches =
+            regex_params_str.find('e') != std::string::npos;
+
+        // Ensure all groups are named so that we can pass to extract_regex.
+        // Do this once per batch.
```
ehsantn: This should be once per operator and not batch right?

ehsantn (approved):
Thanks, Daniel.

bodo/pandas/physical/expression.h
dmsa-or: Yep, that's a typo.

PR #1316, opened 2026-07-22
More decimal work in bodosql and df lib.

DrTodd13:

1) Improve duckdb function signature infrastructure for cases like decimal where you need more advanced type checking and returning.  You can't register just one decimal type because you get everything case to that type and you can't register multiple types because it can't differentiate.  So, you have to register generic "ANY" kernels and use these binding functions to check for decimal and in general return the same kind of decimal.
2) Add decimal support for duckdb floor, ceil, and sign functions signatures.
3) Add decimal support in duckdbTypeToArrow, arrowtypeToDuckDB, and duckdb_logicaltype_to_cudf.
4) Add test for decimal arithmetic.
5) Mark bodosql tests with working decimal.

## Testing strategy

run_ci

## User facing changes

None

## Checklist

### DIFF
--- a/BodoSQL/bodosql/plan_conversion.py
+++ b/BodoSQL/bodosql/plan_conversion.py
@@ -2623,23 +2623,16 @@ def java_call_to_python_call(ctx, java_call, input_plan):
             )
 
             inp_dtype = get_expr_dtype(inp, func_name + " input")
+            sign_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.int64()))
             if compare_types(inp_dtype, int):
                 # If input is an int, first use int8 empty data since
                 # we have to match the return type of Arrow's sign().
                 int8_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.int8()))
                 int8_sign = UnaryOpExpression(int8_empty_data, inp, "sign")
-
-                # Calcite retains the size of the input integer type.
-                # Using inp.empty_data directly could be problematic
-                # if it is unsigned, so cast to the equivalent pyarrow
-                # type of what is in the plan.
-                inp_sql_type = java_call.getOperands()[0].getType()
-                pa_int_type = sql_type_to_pa_type(ctx, inp_sql_type)
-                sql_pa_empty = pd.Series(dtype=pd.ArrowDtype(pa_int_type))
-                return CastExpression(sql_pa_empty, int8_sign)
+                return CastExpression(sign_empty_data, int8_sign)
             else:
                 # If input is a float, return the original float type
-                return UnaryOpExpression(inp.empty_data, inp, "sign")
+                return UnaryOpExpression(sign_empty_data, inp, "sign")
 
         # Binary power: POWER(x, y) -> use __pow__ via ArithOpExpression
         if func_name == "POWER" and len(op_exprs) == 2:
@@ -3403,7 +3396,7 @@ def java_call_to_python_call(ctx, java_call, input_plan):
                         inp.empty_data, scale_negative, negative_scale_result, inp
                     )
                 else:
-                    # If input is a float, we do:
+                    # If input is a float or decimal, we do:
                     # result = [FLOOR/CEIL](inp * 10^scale) / 10^scale
                     power_of_10 = ArithOpExpression(
                         float_empty_data, ten_expr, scale_expr, "__pow__"
@@ -3414,8 +3407,36 @@ def java_call_to_python_call(ctx, java_call, input_plan):
                     scaled_inp_rounded = UnaryOpExpression(
                         float_empty_data, scaled_inp, func_name.lower()
                     )
+                    output_empty_data = inp.empty_data
+                    if compare_types(inp_dtype, pa.Decimal128Type):
+                        if isinstance(scale_expr, ConstantExpression) and isinstance(
+                            scale_expr.value, int
+                        ):
+                            new_scale = scale_expr.value
+
+                            # get the pyarrow decimal type for the column
+                            pa_dtype = output_empty_data.dtypes[0].pyarrow_dtype
+
+                            if not pa.types.is_decimal(pa_dtype):
+                                raise TypeError(
+                                    f"column {output_empty_data.columns[0]} is not a decimal Arrow dtype: {pa_dtype!r}"
+                                )
+
+                            precision = pa_dtype.precision
+                            current_scale = pa_dtype.scale
+
+                            if new_scale < current_scale:
+                                output_empty_data = pd.Series(
+                                    dtype=pd.ArrowDtype(
+                                        pa.decimal128(precision, new_scale)
+                                    )
+                                ).to_frame(output_empty_data.columns[0])
+
                     return ArithOpExpression(
-                        inp.empty_data, scaled_inp_rounded, power_of_10, "__truediv__"
+                        output_empty_data,
+                        scaled_inp_rounded,
+                        power_of_10,
+                        "__truediv__",
                     )
         elif func_name == "POW" and len(op_exprs) == 2:
             left = op_exprs[0]
--- a/BodoSQL/bodosql/plan_conversion.py
+++ b/BodoSQL/bodosql/plan_conversion.py
@@ -375,7 +375,7 @@ def adjustScale(inp_dtype, scale_expr, output_empty_data):
         elif isinstance(scale_expr, int):
             new_scale = scale_expr
 
-        if new_scale:
+        if new_scale is not None:
             # get the pyarrow decimal type for the column
             pa_dtype = output_empty_data.dtypes.iloc[0].pyarrow_dtype
 
@@ -387,7 +387,7 @@ def adjustScale(inp_dtype, scale_expr, output_empty_data):
             precision = pa_dtype.precision
             current_scale = pa_dtype.scale
 
-            if new_scale < current_scale:
+            if new_scale != current_scale:
                 output_empty_data = pd.Series(
                     dtype=pd.ArrowDtype(pa.decimal128(precision, new_scale))
                 ).to_frame(output_empty_data.columns[0])
@@ -2653,16 +2653,8 @@ def java_call_to_python_call(ctx, java_call, input_plan):
             )
 
             inp_dtype = get_expr_dtype(inp, func_name + " input")
-            sign_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.int64()))
-            if compare_types(inp_dtype, int):
-                # If input is an int, first use int8 empty data since
-                # we have to match the return type of Arrow's sign().
-                int8_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.int8()))
-                int8_sign = UnaryOpExpression(int8_empty_data, inp, "sign")
-                return CastExpression(sign_empty_data, int8_sign)
-            else:
-                # If input is a float, return the original float type
-                return UnaryOpExpression(sign_empty_data, inp, "sign")
+            sign_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.int8()))
+            return UnaryOpExpression(sign_empty_data, inp, "sign")
 
         # Binary power: POWER(x, y) -> use __pow__ via ArithOpExpression
         if func_name == "POWER" and len(op_exprs) == 2:
@@ -5795,7 +5787,7 @@ def sql_type_to_pa_type(ctx, sql_type):
     if sql_type_name.equals(SqlTypeName.DECIMAL):
         precision = sql_type.getPrecision()
         scale = sql_type.getScale()
-        if scale > 38:
+        if precision > 38:
             raise ValueError("BodoSQL cpp backend does not support decimal256.")
         else:
             return pa.decimal128(precision, scale)
--- a/BodoSQL/bodosql/plan_conversion.py
+++ b/BodoSQL/bodosql/plan_conversion.py
@@ -3415,7 +3415,7 @@ def java_call_to_python_call(ctx, java_call, input_plan):
                             new_scale = scale_expr.value
 
                             # get the pyarrow decimal type for the column
-                            pa_dtype = output_empty_data.dtypes[0].pyarrow_dtype
+                            pa_dtype = output_empty_data.dtypes.iloc[0].pyarrow_dtype
 
                             if not pa.types.is_decimal(pa_dtype):
                                 raise TypeError(
--- a/BodoSQL/bodosql/plan_conversion.py
+++ b/BodoSQL/bodosql/plan_conversion.py
@@ -1975,8 +1975,8 @@ def java_call_to_python_call(ctx, java_call, input_plan):
                 cleaned_input,
             )
 
-        if operand_type.getSqlTypeName().equals(
-            SqlTypeName.DECIMAL
+        if pa.types.is_decimal(
+            in_expr.empty_data.dtypes.iloc[0].pyarrow_dtype
         ) and target_type.getSqlTypeName().equals(SqlTypeName.VARCHAR):
             dscale = in_expr.empty_data.dtypes.iloc[0].pyarrow_dtype.scale
             if dscale < 0:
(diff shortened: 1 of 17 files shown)


bodo/tests/utils.py
```diff
@@ -1350,6 +1350,15 @@ def _test_equal(
     except ImportError:
         csr_matrix = type(None)
 
+    if int(os.environ.get("BODO_TESTING_SHOW_OUTPUT", 0)) > 0:
```
IsaacWarren, line 1353: Doesn't pytest show the mismatched values if you run with pytest -s -v?

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -365,6 +365,36 @@ def java_named_param_to_python_literal(ctx, named_param, input_plan):
...
+                    f"column {output_empty_data.columns[0]} is not a decimal Arrow dtype: {pa_dtype!r}"
+                )
+
+            precision = pa_dtype.precision
+            current_scale = pa_dtype.scale
+
+            if new_scale < current_scale:
+                output_empty_data = pd.Series(
+                    dtype=pd.ArrowDtype(pa.decimal128(precision, new_scale))
+                ).to_frame(output_empty_data.columns[0])
```
IsaacWarren: Why only decrease the scale?

bodo/pandas/_plan.cpp
```diff
@@ -1735,6 +1737,112 @@ const duckdb::vector<ScalarFunctionSignature> UNARY_INT_SIGNATURES = {
...
+    if (arg_type.id() != duckdb::LogicalTypeId::DECIMAL) {
+        throw std::runtime_error("IntDecimalBind: expected DECIMAL argument");
+    }
+
+    return duckdb::make_uniq<IntDecimalBindData>(duckdb::LogicalType::TINYINT);
+}
+
+const duckdb::vector<ScalarFunctionSignature> UNARY_DECIMAL_SIGNATURES = {
+    ScalarFunctionSignature({duckdb::LogicalType::ANY},
+                            duckdb::LogicalType::ANY, SameReturnDecimalBind)};
```
dmsa-or: If there's only one decimal signature then maybe this shouldn't be a vector. But your choice.

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -5754,6 +5792,13 @@ def sql_type_to_pa_type(ctx, sql_type_name):
         return pa.time64("ns")
     if sql_type_name.equals(SqlTypeName.NULL):
         return pa.null()
+    if sql_type_name.equals(SqlTypeName.DECIMAL):
+        precision = sql_type.getPrecision()
+        scale = sql_type.getScale()
+        if scale > 38:
+            raise ValueError("BodoSQL cpp backend does not support decimal256.")
```
IsaacWarren, line 5810: This should be precision > 38, right?

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -2620,26 +2648,21 @@ def cur_local_timestamp():
 
         if func_name == "SIGN" and len(op_exprs) == 1:
             inp = op_exprs[0]
-            ensure_type_of_expr(inp, func_name + " input", (int, float))
+            ensure_type_of_expr(
+                inp, func_name + " input", (int, float, pa.Decimal128Type)
+            )
 
             inp_dtype = get_expr_dtype(inp, func_name + " input")
+            sign_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.int64()))
```
IsaacWarren: Do we always want to use int64? Previously we were using the input type if it was an int

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -365,6 +365,36 @@ def java_named_param_to_python_literal(ctx, named_param, input_plan):
...
+    if compare_types(inp_dtype, pa.Decimal128Type):
+        new_scale = None
+        if isinstance(scale_expr, ConstantExpression) and isinstance(
+            scale_expr.value, int
+        ):
+            new_scale = scale_expr.value
+        elif isinstance(scale_expr, int):
+            new_scale = scale_expr
+
+        if new_scale:
```
```suggestion
        if new_scale is not None:
```

bodo/pandas/_plan.cpp
```diff
@@ -1750,12 +1858,20 @@ duckdb::vector<ScalarFunctionSignature> copy_signatures(
...
+    register_duckdb_scalar_func(
+        db, "ceil",
+        append_signatures(copy_signatures(UNARY_FLOAT_SIGNATURES),
+                          UNARY_DECIMAL_SIGNATURES));
+    register_duckdb_scalar_func(
+        db, "abs",
+        append_signatures(
+            copy_signatures(UNARY_FLOAT_SIGNATURES),
+            append_signatures(copy_signatures(UNARY_DECIMAL_SIGNATURES),
+                              UNARY_INT_SIGNATURES)));
```
dmsa-or: Perhaps this is slightly more readable if you do:

```suggestion
    register_duckdb_scalar_func(
        db, "abs",
        append_signatures(
            append_signatures(
                copy_signatures(UNARY_FLOAT_SIGNATURES),
                UNARY_DECIMAL_SIGNATURES),
            UNARY_INT_SIGNATURES));
```

IsaacWarren (approved):
Thanks @DrTodd13

bodo/pandas/_plan.cpp
```diff
@@ -1779,7 +1895,11 @@ void register_duckdb_scalar_funcs(duckdb::shared_ptr<duckdb::DuckDB> db) {
         append_signatures(
             {ScalarFunctionSignature({duckdb::LogicalType::BIGINT},
                                      duckdb::LogicalType::TINYINT)},
-            UNARY_FLOAT_SIGNATURES));
+            append_signatures(
+                {ScalarFunctionSignature({duckdb::LogicalType::ANY},
+                                         duckdb::LogicalType::TINYINT,
+                                         IntDecimalBind)},
+                UNARY_FLOAT_SIGNATURES)));
```
dmsa-or: As above, it's probably cleanest to pass nested `append_signatures` as the first argument (and will also result in the fewest total appends), although it's an optimization that doesn't matter.

bodo/pandas/_util.cpp
```diff
@@ -222,12 +222,27 @@ getDefaultValueForDuckdbValueType(const duckdb::Value &value) {
...
         } break;
+        case duckdb::LogicalTypeId::DECIMAL: {
+            uint8_t precision = 0;
+            uint8_t scale = 0;
+
+            value.type().GetDecimalProperties(precision, scale);
+
+            if (precision <= 0 || precision > 38) {
+                throw std::runtime_error(
+                    "duckdbTypeToArrow invalid DECIMAL precision " +
```
dmsa-or:
```suggestion
                    "getDefaultValueForDuckdbValueType invalid DECIMAL precision " +
```

BodoSQL/bodosql/tests/test_decimal.py
```diff
@@ -1820,7 +1834,7 @@ def test_decimal_floor(df, expr, answer, memory_leak_check):
                         [
                             Decimal("648.2935"),
                             Decimal("-152.5826"),
-                            Decimal("-0.15122"),
+                            Decimal("-1.15122"),
```
dmsa-or, line 1831: Why was this changed?

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -3432,22 +3471,16 @@ def clean_regex_params(regex_params_expr):
...
-        elif func_name == "LOG" and len(op_exprs) == 1:
-            # The SQL function LOG(x) is mapped to Calcite LOG(x),
-            # which in Calcite and most systems means LN(x), but Bodo
-            # defines SQL LOG(x) as LOG10(x). For now, we raise an
-            # informative error while we decide on the best solution.
-            raise ValueError(
-                "LOG with 1 argument is not currently supported in the C++ backend. Consider using the unambiguous LN or LOG10 instead."
-            )
-        elif func_name == "LOG" and len(op_exprs) == 2:
+        elif func_name == "LOG" and len(op_exprs) in (1, 2):
```
dmsa-or, line 3485: If we've decided to use LOG10 in the 1-argument case, there are some test cases we can enable, mostly in `test_numeric_fns.py`.

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -365,6 +365,36 @@ def java_named_param_to_python_literal(ctx, named_param, input_plan):
...
+
+            if not pa.types.is_decimal(pa_dtype):
+                raise TypeError(
+                    f"column {output_empty_data.columns[0]} is not a decimal Arrow dtype: {pa_dtype!r}"
+                )
+
+            precision = pa_dtype.precision
+            current_scale = pa_dtype.scale
+
+            if new_scale < current_scale:
```
dmsa-or: I think this code works properly when `scale_expr` is negative but maybe add a comment about it. And is there an issue with increasing the scale beyond the scale of the original decimal type?

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -365,6 +365,36 @@ def java_named_param_to_python_literal(ctx, named_param, input_plan):
         raise NotImplementedError(f"NamedParam type {cdtype} not supported yet")
 
 
+def adjustScale(inp_dtype, scale_expr, output_empty_data):
```
dmsa-or:
```suggestion
def adjust_scale(inp_dtype, scale_expr, output_empty_data):
```

bodo/tests/utils.py
DrTodd13: I was running with -v flag and it didn't seem to do it for me.

BodoSQL/bodosql/plan_conversion.py
DrTodd13: yep.  good catch.  got it right the other places but brainfart on this one.

bodo/pandas/_plan.cpp
```diff
@@ -1735,6 +1737,112 @@ const duckdb::vector<ScalarFunctionSignature> UNARY_INT_SIGNATURES = {
     ScalarFunctionSignature({duckdb::LogicalType::BIGINT},
                             duckdb::LogicalType::BIGINT)};
 
+// helper FunctionData to carry resolved return type
+struct DecimalBindData : public duckdb::FunctionData {
```
dmsa-or, line 1741: Isn't this identical to `IntBindData` aside from the name? Can you use the same struct for both cases?

BodoSQL/bodosql/plan_conversion.py
DrTodd13: Seems that systems vary.  Some do that but it seems like even int8 is fine.  That can simplify the code too.  Will try always int8 for all types.

bodo/pandas/_plan.cpp
```diff
@@ -1735,6 +1737,112 @@ const duckdb::vector<ScalarFunctionSignature> UNARY_INT_SIGNATURES = {
...
+        if (other_decimal) {
+            return return_type == other_decimal->return_type;
+        } else {
+            return false;
+        }
+    }
+};
+
+// Bind callback: inspect concrete argument types and set return type
+static duckdb::unique_ptr<duckdb::FunctionData> IntDecimalBind(
```
dmsa-or: Can you parameterize the return type of this so it's not limited to TINYINT? Any maybe also parameterize the input type so it's not restricted to DECIMAL? (We could reuse it for other DuckDB types with parameters like DECIMAL.)

BodoSQL/bodosql/plan_conversion.py
DrTodd13: good point.  fixed and testing.

bodo/pandas/_plan.cpp
```diff
@@ -1735,6 +1737,112 @@ const duckdb::vector<ScalarFunctionSignature> UNARY_INT_SIGNATURES = {
...
+        if (other_decimal) {
+            return return_type == other_decimal->return_type;
+        } else {
+            return false;
+        }
+    }
+};
+
+// Bind callback: inspect concrete argument types and set return type
+static duckdb::unique_ptr<duckdb::FunctionData> SameReturnDecimalBind(
```
dmsa-or: This function is the same as `IntDecimalBind` but it uses the argument type as the return type. These functions can definitely be refactored for less code duplication.

bodo/pandas/_plan.cpp
```diff
@@ -1735,6 +1737,112 @@ const duckdb::vector<ScalarFunctionSignature> UNARY_INT_SIGNATURES = {
...
+    duckdb::ClientContext &context, duckdb::ScalarFunction &bound_function,
+    duckdb::vector<bododuckdb::unique_ptr<bododuckdb::Expression>> &arguments) {
+    // Expect exactly one argument for unary abs
+    if (arguments.size() != 1) {
+        throw std::runtime_error(
+            "SameReturnDecimalBind: expected one argument");
+    }
```
dmsa-or: Not specific to abs. I would probably put "unary" somewhere in the titles of these functions.

dmsa-or (changes requested):

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -2620,26 +2648,13 @@ def cur_local_timestamp():
...
-                # type of what is in the plan.
-                inp_sql_type = java_call.getOperands()[0].getType()
-                pa_int_type = sql_type_to_pa_type(ctx, inp_sql_type.getSqlTypeName())
-                sql_pa_empty = pd.Series(dtype=pd.ArrowDtype(pa_int_type))
-                return CastExpression(sql_pa_empty, int8_sign)
-            else:
-                # If input is a float, return the original float type
-                return UnaryOpExpression(inp.empty_data, inp, "sign")
+            sign_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.int8()))
+            return UnaryOpExpression(sign_empty_data, inp, "sign")
```
dmsa-or: Please add a comment about why we always return int8 now. I think Ehsan was concerned before that Calcite might choose something else. Also, did you make sure there is no error due to a type mismatch using int8 empty data when the input is float? Since Arrow returns float for float input.

BodoSQL/bodosql/tests/test_decimal.py
DrTodd13: Arrow makes that -0 and then convert to string doesn't match.

BodoSQL/bodosql/plan_conversion.py
DrTodd13: Already changed to new_scale != current_scale.

BodoSQL/bodosql/plan_conversion.py
DrTodd13: done

bodo/pandas/_plan.cpp
DrTodd13: done

bodo/pandas/_plan.cpp
DrTodd13: Judged not worth it.

bodo/pandas/_plan.cpp
DrTodd13: If need be then next person can templatize this function.

bodo/pandas/_plan.cpp
dmsa-or: Really? This one would be an easy fix.

bodo/pandas/_plan.cpp
dmsa-or: Perhaps allowing other input types than DECIMAL is too much right now, but adding a return type parameter would be very easy.

dmsa-or (approved):

bodo/pandas/_plan.cpp
```diff
@@ -1789,6 +1789,57 @@ static duckdb::unique_ptr<duckdb::FunctionData> SameReturnDecimalBindUnary(
...
+        if (other_decimal) {
+            return return_type == other_decimal->return_type;
+        } else {
+            return false;
+        }
+    }
+};
+
+// Bind callback: inspect concrete argument types and set return type
+static duckdb::unique_ptr<duckdb::FunctionData> SameReturnRoundDecimalBindUnary(
```
dmsa-or: If we need more variants of this function we should consider a refactor because the code duplication will quickly get out of hand. Also "Unary" should not be in the name of this one.

PR #1319, opened 2026-07-23
Support REGEXP_REPLACE in BodoSQL C++ backend

dmsa-or:

REGEXP_REPLACE is now supported in the C++ backend. The case-insensitive option is unsupported, meaning most of the test cases could not be marked `bodosql_cpp`. I added a test case that goes through the `replace_substring_regex_single` path.

## Testing strategy

CI

## Checklist


bodo/pandas/physical/expression.h
```diff
@@ -1742,6 +1742,90 @@ class PhysicalArrowExpression : public PhysicalExpression {
...
+    arrow::Datum do_arrow_compute_replace_substring_regex_single(
+        T res, std::string pattern_str, std::string replacement_str,
+        int occurrence_num) {
+        if (occurrence_num < 1) {
+            throw std::invalid_argument(
+                "occurrences_num argument to replace_substring_regex_single "
+                "must be one or greater.");
+        }
+
+        // Intermediate results must be Datums only because Bodo doesn't support
```
IsaacWarren: We do support list types, maybe some support is missing somewhere though

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -4172,6 +4172,114 @@ def get_common_signed(
...
+            if len(op_exprs) >= 5:
+                # Need to replace the substring that is the op_exprs[4]-th occurrence / regex match
+                occurrence_expr = op_exprs[4]
+                ensure_arg_is_const_expr_of_type(
+                    occurrence_expr, "occurrence_expr", int
+                )
+                occurrence_num = occurrence_expr.value
+                if occurrence_num < 0:
+                    raise ValueError(
+                        f"{func_name} occurences argument must be 0 or greater"
```
```suggestion
                        f"{func_name} occurrences argument must be 0 or greater"
```

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -4172,6 +4172,114 @@ def get_common_signed(
...
+                elif occurrence_num == 0:
+                    # If occurrence_num is 0, all occurrences should be replaced.
+                    # For replace_substring_regex, this is signified by passing max_replacements = -1.
+                    occurrence_num = -1
+            else:
+                # Default is to replace all occurrences
+                occurrence_num = -1
+
+            if len(op_exprs) >= 6:
+                regex_params_expr = op_exprs[5]
```
IsaacWarren, line 4307: This isn't used anywhere?

IsaacWarren: Also should we throw an error if there's an 'i'?

IsaacWarren (approved):
Thanks @dmsa-or

BodoSQL/bodosql/plan_conversion.py
dmsa-or: Yeah, at the moment this is not used. `clean_regex_params` should throw an error if there is an 'i' though. However my next PR should add support for all the regex parameters.

bodo/pandas/physical/expression.h
dmsa-or: Oh, I guess it's possible, I don't think I actually tried it without converting to Datum first. But I have to convert to Datum at some point anyway and it's probably cleanest to do it at the start of the function. Will delete the comment.

bodo/pandas/physical/expression.cpp
```diff
@@ -482,11 +475,16 @@ arrow::Datum do_arrow_compute_multi_input_datum(
             // At least one of the second and third arguments are arrays: use
             // bodo_substr_three
 
+            // Copy so we can modify datum args (pre-sized to 3 elements)
+            std::vector<arrow::Datum> array_arg_datums(3);
+            std::copy(arg_datums.begin(), arg_datums.end(),
+                      array_arg_datums.begin());
```
ehsantn, line 481: Does this copy the data as well? If so, can it be avoided?

bodo/pandas/physical/expression.h
```diff
@@ -1742,6 +1742,88 @@ class PhysicalArrowExpression : public PhysicalExpression {
         return captured_field_index_datum;
     }
 
+    template <typename T>
+    arrow::Datum do_arrow_compute_replace_substring_regex_single(
```
ehsantn: Why does this need to be a method of this class instead of a standalone function? The class is getting big and harder to read now.

ehsantn: Also, this function needs a docstring to explain what it is.

ehsantn (approved):
Looks good to me overall.

bodo/pandas/physical/expression.h
dmsa-or: Okay, I'll move all of these `do_arrow_compute ...` functions outside of the class.

bodo/pandas/physical/expression.cpp
dmsa-or: I don't know that there's an easy way to avoid this. But it shouldn't copy the data. Datums are usually wrappers for shared pointers to Arrays.

bodo/pandas/physical/expression.h
dmsa-or: How do you feel about moving them to `expression.cpp` instead? `do_arrow_compute_replace_substring_regex_single` should be more-or-less independent, but the others modify fields in `PhysicalArrowExpression`.

ehsantn: Sure.

ehsantn: It's more clear if independent ones are not class methods.

PR #1320, opened 2026-07-23
bodosql trunc 2-arg

DrTodd13:

Implement truncate with 2-arguments.

## Testing strategy

run_ci

## User facing changes

None

## Checklist

### DIFF
--- a/BodoSQL/bodosql/plan_conversion.py
+++ b/BodoSQL/bodosql/plan_conversion.py
@@ -2623,23 +2623,16 @@ def java_call_to_python_call(ctx, java_call, input_plan):
             )
 
             inp_dtype = get_expr_dtype(inp, func_name + " input")
+            sign_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.int64()))
             if compare_types(inp_dtype, int):
                 # If input is an int, first use int8 empty data since
                 # we have to match the return type of Arrow's sign().
                 int8_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.int8()))
                 int8_sign = UnaryOpExpression(int8_empty_data, inp, "sign")
-
-                # Calcite retains the size of the input integer type.
-                # Using inp.empty_data directly could be problematic
-                # if it is unsigned, so cast to the equivalent pyarrow
-                # type of what is in the plan.
-                inp_sql_type = java_call.getOperands()[0].getType()
-                pa_int_type = sql_type_to_pa_type(ctx, inp_sql_type)
-                sql_pa_empty = pd.Series(dtype=pd.ArrowDtype(pa_int_type))
-                return CastExpression(sql_pa_empty, int8_sign)
+                return CastExpression(sign_empty_data, int8_sign)
             else:
                 # If input is a float, return the original float type
-                return UnaryOpExpression(inp.empty_data, inp, "sign")
+                return UnaryOpExpression(sign_empty_data, inp, "sign")
 
         # Binary power: POWER(x, y) -> use __pow__ via ArithOpExpression
         if func_name == "POWER" and len(op_exprs) == 2:
@@ -3403,7 +3396,7 @@ def java_call_to_python_call(ctx, java_call, input_plan):
                         inp.empty_data, scale_negative, negative_scale_result, inp
                     )
                 else:
-                    # If input is a float, we do:
+                    # If input is a float or decimal, we do:
                     # result = [FLOOR/CEIL](inp * 10^scale) / 10^scale
                     power_of_10 = ArithOpExpression(
                         float_empty_data, ten_expr, scale_expr, "__pow__"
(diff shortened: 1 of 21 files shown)


bodo/pandas/physical/expression.cpp
```diff
@@ -520,6 +521,25 @@ std::shared_ptr<array_info> do_arrow_compute_cast(
...
 }
 
+void do_result_type_cast(arrow::Result<arrow::Datum>& out_res,
+                         const std::shared_ptr<arrow::DataType> result_type) {
+    arrow::Datum out_datum = out_res.ValueOrDie();
+    std::shared_ptr<arrow::DataType> out_dtype = out_datum.type();
+    if (result_type && out_dtype != result_type) {
+        // Cast to result type if available and different from current type.
+        arrow::compute::CastOptions cast_opts;
+        cast_opts.allow_int_overflow = true;
```
dmsa-or, line 531: Any reason why we might want or not want to have `cast_opts.allow_float_truncate = true` here?

IsaacWarren (approved):

bodo/pandas/physical/expression.cpp
```diff
@@ -620,6 +627,7 @@ arrow::Datum do_arrow_compute_unary(
             "): " + cmp_res.status().message());
     }
 
+    do_result_type_cast(cmp_res, result_type);
```
dmsa-or, line 632: I guess this means we don't have to be as careful about the empty data type matching the Arrow return type, since we always cast to the empty data type.

bodo/pandas/physical/expression.cpp
```diff
@@ -520,6 +521,25 @@ std::shared_ptr<array_info> do_arrow_compute_cast(
     return ConvertDatumToArrayInfo(casted);
 }
 
+void do_result_type_cast(arrow::Result<arrow::Datum>& out_res,
+                         const std::shared_ptr<arrow::DataType> result_type) {
+    arrow::Datum out_datum = out_res.ValueOrDie();
+    std::shared_ptr<arrow::DataType> out_dtype = out_datum.type();
+    if (result_type && out_dtype != result_type) {
```
dmsa-or: You should use Arrow's dedicated `DataType.Equals` function for this comparison to properly handle parameterized types. Is `out_dtype != result_type` comparing the pointers here? If so, it would always be true and thus a cast would always take place as long as `result_type` is given. I bet if you did `!=` on the actual `DataType`s, this would return False for decimals with different scales, so your trunc fix wouldn't work. So make sure to compare with `DataType.Equals`.

bodo/pandas/physical/expression.cpp
```diff
@@ -533,29 +553,14 @@ arrow::Datum do_arrow_compute_binary(
             "): " + cmp_res.status().message());
     }
 
-    arrow::Datum cmp_datum = cmp_res.ValueOrDie();
```
dmsa-or, line 536: I think there was a similar cast happening in `do_arrow_compute_case`, can you replace it with `do_result_type_cast` there too?

dmsa-or (approved):

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -2815,19 +2815,52 @@ def cur_local_timestamp():
...
+                    return UnaryOpExpression(output_empty_data, inp, "trunc")
             else:
-                # If input is a float, return trunc(inp) as normal
-                return UnaryOpExpression(inp.empty_data, inp, "trunc")
+                scale_expr = op_exprs[1]
+                ensure_type_of_expr(scale_expr, "scale_expr", int)
+                if compare_types(inp_dtype, int):
+                    raise TypeError(
+                        "TRUNCATE with scale not currently supported for integer types."
+                    )
```
dmsa-or, line 2839: I think you could reuse a lot of the code that was written for `FLOOR` and `CEIL` for integers. TRUNC should be the equivalent of FLOOR when input is positive and CEIL when input is negative, but there's probably a simpler formula for TRUNC.

bodo/pandas/physical/expression.cpp
DrTodd13: done

bodo/pandas/physical/expression.cpp
DrTodd13: didn't write or modify this code in this PR.  only moved it but good catch and I changed it.

bodo/pandas/physical/expression.cpp
DrTodd13: done.

bodo/pandas/physical/expression.cpp
DrTodd13: added that and decimal overflow as well.

bodo/pandas/physical/expression.cpp
DrTodd13: done.

PR #1324, opened 2026-07-24
Support STRCMP and FORMAT in BodoSQL C++ backend

dmsa-or:

STRCMP and FORMAT are now supported in the C++ backend.

## Testing strategy

CI

## Checklist

### DIFF
--- a/BodoSQL/bodosql/plan_conversion.py
+++ b/BodoSQL/bodosql/plan_conversion.py
@@ -3879,6 +3879,201 @@ def java_call_to_python_call(ctx, java_call, input_plan):
             return ArrowScalarFuncExpression(
                 uint64_empty_data, [is_bit_set, one_expr, zero_expr], "if_else", ()
             )
+        elif func_name == "FORMAT" and len(op_exprs) == 2:
+            """
+            Returns a formatted string representation of a number.
+            The number is rounded to the specified precision and printed
+            with enough zeros to demonstrate that precision.
+            Commas are also inserted as thousands separators.
+            """
+
+            src = op_exprs[0]
+            ensure_type_of_expr(src, "src", (int, float))
+
+            precision_digits = op_exprs[1]
+            ensure_type_of_expr(precision_digits, "precision_digits", int)
+
+            # MySQL treats any precision digits < 0 as 0
+            int_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.int64()))
+            zero_expr = ConstantExpression(int_empty_data, input_plan, 0)
+            precision_digits = ArrowScalarFuncExpression(
+                int_empty_data,
+                [precision_digits, zero_expr],
+                "max_element_wise",
+                (False,),
+            )
+
+            # Round the number to the specified precision
+            rounded = ArithOpExpression(src.empty_data, src, precision_digits, "round")
+
+            # Get the string representation of the number
+            str_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.string()))
+            rounded_str = CastExpression(str_empty_data, rounded)
+
+            # Get the character index of the decimal point in the string. Returns -1 if not found.
+            decimal_point_pos = ArrowScalarFuncExpression(
+                int_empty_data, [rounded_str], "find_substring", (".",)
+            )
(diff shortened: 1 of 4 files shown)


bodo/pandas/physical/expression.cpp
```diff
@@ -1194,9 +1196,31 @@ std::shared_ptr<ExprResult> PhysicalArrowExpression::ProcessBatch(
...
-                                              scalar_func_data.arrow_func_name);
+
+        // Special handling for multi-input Arrow functions with options
+        if (scalar_func_data.arrow_func_name == "max_element_wise" ||
+            scalar_func_data.arrow_func_name == "min_element_wise") {
+            arrow::compute::ElementWiseAggregateOptions opts;
+
+            assert_py_args_is_tuple(scalar_func_data.args,
+                                    scalar_func_data.arrow_func_name.c_str());
+            if (PyTuple_Size(scalar_func_data.args) > 1) {
```
IsaacWarren:
```suggestion
            if (PyTuple_Size(scalar_func_data.args) > 0) {
```

bodo/pandas/physical/expression.h
```diff
@@ -200,7 +200,8 @@ extern std::function<bool(int)> less_equal_test;
  */
 arrow::Datum do_arrow_compute_multi_input_datum(
     const std::vector<arrow::Datum> &arg_datums,
-    const std::string &arrow_func_name);
+    const std::string &arrow_func_name,
+    const arrow::compute::FunctionOptions *func_options = nullptr);
```
IsaacWarren, line 204: func_options isn't used anywhere?

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -4034,6 +4229,33 @@ def get_common_signed(
...
+            str1_greater = ComparisonOpExpression(
+                bool_empty_data, str1_expr, str2_expr, operator.gt
+            )
+            strings_equal = ComparisonOpExpression(
+                bool_empty_data, str1_expr, str2_expr, operator.eq
+            )
+
+            # Using lexicographical comparison, return 1 if str1 > str2, -1 if str1 < str2, 0 if str1 == str2.
+            int_empty_data = pd.Series(dtype=pd.ArrowDtype(pa.int64()))
+            one_expr = ConstantExpression(int_empty_data, input_plan, 1)
```
IsaacWarren, line 4248: Instead of using case expressions we could do result = (str1 > str2) - (str1 < str2), should be better performance with no branching

IsaacWarren (commented):
Thanks @dmsa-or, a couple comments

BodoSQL/bodosql/plan_conversion.py
dmsa-or: Seems like Arrow doesn't support boolean subtraction, so we would need to cast them to int first. Do you  think this way is still faster with the casting?

IsaacWarren: Not sure to be honest. Fine to leave as is

dmsa-or: Okay, left it alone for now. I was going to benchmark the different approaches, but there's an unrelated bug making that difficult right now. I told Ehsan about it.

IsaacWarren (approved):

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -3892,6 +3892,201 @@ def get_common_signed(
...
+            Commas are also inserted as thousands separators.
+            """
+
+            src = op_exprs[0]
+            ensure_type_of_expr(src, "src", (int, float))
+
+            precision_digits = op_exprs[1]
+            ensure_type_of_expr(precision_digits, "precision_digits", int)
+
+            # MySQL treats any precision digits < 0 as 0
```
ehsantn, line 3909: This "vectorized" implementation loses a lot of memory locality and can be a lot slower than a custom kernel. May not be a big deal for this function right now but let's keep this in mind.

ehsantn (approved):
Looks good to me overall. Thanks @dmsa-or.

PR #1327, opened 2026-07-28
[GPU] Handle empty build table case in broadcast join

scott-routledge2:

As titled, previously an empty build table would result in a null join handle. We can handle empty build table case for broadcast join in the same way we handle it for other joins. 

As fixes minor testing issue with broadcast join in DF library tests

## Testing strategy

New test added, PR CI

## Checklist

### DIFF
--- a/bodo/tests/test_df_lib/test_end_to_end.py
+++ b/bodo/tests/test_df_lib/test_end_to_end.py
@@ -26,7 +26,6 @@ from bodo.pandas.utils import (
 from bodo.tests.utils import (
     _test_equal,
     is_multi_worker_per_gpu_test,
-    set_broadcast_join,
     temp_config_override,
 )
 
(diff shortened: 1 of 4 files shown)


bodo/tests/test_df_lib/test_end_to_end.py
```diff
@@ -1448,12 +1447,11 @@ def test_project_after_filter(datapath):
 
 @pytest.mark.gpu
 @pytest.mark.parametrize("how", ["inner", "left", "right", "outer"])
-@pytest.mark.parametrize("broadcast", [True, False])
```
scott-routledge2, line 1451: Because broadcast join is something that we evaluate at plan execution time (i.e. inside of _test_equal) these tests weren't actually doing broadcast join.

DrTodd13 (approved):

IsaacWarren (approved):

PR #1328, opened 2026-07-29
[GPU][BSE-5520] Fix aggregations containing nulls

scott-routledge2:

Enable SQL ANY_VALUE on GPU and address 2 issues with null/empty data in GPU reduce:
1. When using SQL rules, Null values were being treated as regular values in the MPI reduce
2. If Rank 0 had empty data, it was excluded from the MPI reduce

The first issue is addressed by using an identity value for ranks containing Null (in addition to an extra all reduce to determine if any rank has valid data)

The second issue is addressed by having all ranks participate in the computation. For FIRST/ANY_VALUE, the first rank with non-empty data sends it's result to rank 0. 

## Testing strategy

New tests added, TPCH Q17,Q19 enabled for SQL on GPU. 

## User facing changes

`SELECT ANY_VALUE(x) FROM T` Now runs on GPU

## Checklist

### DIFF
--- a/BodoSQL/bodosql/tests/test_agg_nogroupby.py
+++ b/BodoSQL/bodosql/tests/test_agg_nogroupby.py
@@ -571,7 +571,9 @@ def test_agg_replicated(datapath, memory_leak_check):
 @pytest.mark.parametrize(
     "query",
     [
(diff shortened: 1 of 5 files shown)


DrTodd13 (approved):
thanks!

IsaacWarren (approved):

PR #1330, opened 2026-07-29
Improve support for LIST type in BodoSQL C++ backend

dmsa-or:

Added conversion logic for LIST types and LIST scalars between Arrow and DuckDB.
Added ZIP helper function since Arrow doesn't have compute functions for making ListArrays from regular arrays.

## Testing strategy

Subsequent PRs should test these paths. I've already tested them locally. We might want to write unit tests for `do_arrow_compute_zip` in the future though.

## Checklist

### DIFF
--- a/bodo/pandas/_util.cpp
+++ b/bodo/pandas/_util.cpp
@@ -152,7 +152,61 @@ extractValue(const duckdb::Value &value) {
                 } break;
             }
         } break;
+        case duckdb::LogicalTypeId::LIST: {
+            const auto &duckdb_list_elements =
+                duckdb::ListValue::GetChildren(value);
 
+            duckdb::LogicalType child_duckdb_type =
+                duckdb::ListType::GetChildType(value.type());
+            std::shared_ptr<arrow::DataType> child_arrow_type =
+                duckdbTypeToArrow(child_duckdb_type);
+
+            // Use generic ArrayBuilder to append list elements since we don't
+            // know the child list type at compile time
+            std::unique_ptr<arrow::ArrayBuilder> builder;
+            auto make_builder_status = arrow::MakeBuilder(
+                arrow::default_memory_pool(), child_arrow_type, &builder);
+            if (!make_builder_status.ok()) {
+                throw std::runtime_error(
+                    "Failed to create Arrow ArrayBuilder for list: " +
+                    make_builder_status.ToString());
+            }
+
+            // Iterate over the DuckDB list elements, convert each one to an
+            // Arrow scalar, and append them to the ArrayBuilder
+            for (const auto &list_element : duckdb_list_elements) {
+                if (list_element.IsNull()) {
+                    auto append_status = builder->AppendNull();
+                    if (!append_status.ok()) {
+                        throw std::runtime_error(
+                            "Failed to append null to list builder: " +
+                            append_status.ToString());
+                    }
+                } else {
+                    auto arrow_list_element = convertDuckdbValueToArrowScalar(
+                        list_element, child_arrow_type);
+                    auto append_status =
+                        builder->AppendScalar(*arrow_list_element);
+                    if (!append_status.ok()) {
+                        throw std::runtime_error(
+                            "Failed to append scalar to list builder" +
+                            append_status.ToString());
+                    }
+                }
+            }
+
+            std::shared_ptr<arrow::Array> arrow_list_elements_array;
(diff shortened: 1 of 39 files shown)


bodo/pandas/physical/expression.cpp
```diff
@@ -2668,6 +2670,173 @@ arrow::Datum PhysicalArrowExpression::do_arrow_compute_random_int64(
...
+arrow::Datum do_arrow_compute_zip(const std::vector<arrow::Datum>& datums) {
+    if (datums.empty()) {
+        throw std::invalid_argument(
+            "do_arrow_compute_zip does not accept an empty vector of datums.");
+    }
+
+    // First, loop over the datums to get the number of rows. We stop after
+    // finding the first array datum since we assume all have
+    // the same length. Should we bother supporting the case where
+    // they don't have the same length by padding with nulls?
```
scott-routledge2: I would lean towards checking the array inputs are all the same length and scalar + array inputs are all the same type and throw an error otherwise to avoid unexpected issues, but I guess it depends on the use case for this function. Is there a specific SQL operation we are going to use it for?

bodo/pandas/physical/expression.cpp
```diff
@@ -2668,6 +2670,173 @@ arrow::Datum PhysicalArrowExpression::do_arrow_compute_random_int64(
...
+        // then the second elements, etc.). This is not going to be ideal for
+        // cache locality, but there may not be a better generic solution that
+        // handles variable-width datatypes.
+        for (int64_t i = 0; i < num_rows; i++) {
+            for (const InputView& input : inputs) {
+                // Append element at index i (or 0, for scalars)from the array.
+                // AppendArraySlice should be faster than getting the scalar at
+                // position i and then appending that.
+                arrow::Status append_status =
+                    interleaved_arr_builder->AppendArraySlice(
```
scott-routledge2, line 2800: So for example, in the case where we have a single array with N-1 scalars, the first element of the output would be a list with N elements and the rest of the rows would be lists containing a single element?

bodo/pandas/_util.cpp
```diff
@@ -907,7 +1044,23 @@ std::shared_ptr<arrow::Scalar> convertDuckdbValueToArrowScalar(
         throw std::runtime_error("Failed to convert duckdb value to scalar: " +
                                  scalar_res.status().ToString());
     }
-    return scalar_res.ValueOrDie();
+    std::shared_ptr<arrow::Scalar> scalar = scalar_res.ValueOrDie();
+
+    if (arrow_type && !arrow_type->Equals(scalar->type)) {
+        // If the scalar type doesn't match our expected type, attempt a cast.
+        // Probably a better way would be to avoid creating a scalar of the
```
scott-routledge2, line 1051: But it sounds like we can't always create a scalar of the desired type due to arrow limitations?

scott-routledge2 (approved):
Thanks @dmsa-or! I think generally unit tests for the custom arrow compute functions would be a good idea. (You can look at bodo/tests/test_*.cpp for examples of c++ unit tests)

bodo/pandas/_util.cpp
dmsa-or: I'm not 100% positive, but I think this is some kind of limitation due to the `MakeScalar` templating. I believe something like `std::make_shared<arrow::LargeStringScalar>("text")` works to create a `LargeStringScalar` directly, but that's not a generic approach.

bodo/pandas/physical/expression.cpp
dmsa-or: No, all rows of the output would be lists with N elements. The final (N-1) elements in each list would just be the (N-1) scalars in order. The first element of each list would be different corresponding to the row value of the single input array. So the output would look like:
```
[ [a0_row0,     s1, s2, ..., s(N-1)],
  [a0_row1,     s1, s2, ..., s(N-1)],
   ...
  [a0_row(R-1), s1, s2, ..., s(N-1)] ]
```

bodo/pandas/physical/expression.cpp
dmsa-or: Okay, I will do that validation. In terms of SQL functions, I think right now I'm only using this with a single argument for the `STRTOK_TO_ARRAY` and `SPLIT` plan conversions.

ehsantn (approved):
Thanks @dmsa-or .

PR #1335, opened 2026-08-03
[BSE-5551] Calcite upgrade 1.39

scott-routledge2:

* Applied diff from Calcite 1.38 -> 1.39 upgrade to relevant files. (See the release notes for a complete list of changes: https://calcite.apache.org/docs/history.html#v1-39-0)

* Changed CatalogSchema `getTable()` and `getSubSchema()` APIs to return null when a sub-schema/table isn't found and disabled caching to be compatible with changes introduced in this issue: https://issues.apache.org/jira/browse/CALCITE-6728. I opened [a followup issue](https://bodo.atlassian.net/jira/software/c/projects/BSE/boards/25?assignee=712020%3A75fb3a19-d187-4101-b627-42e0d56f6b97&selectedIssue=BSE-5563) to implement the new Lookup API for our Schemas to get better performance-- though as far as I can tell, performance is currently not worse than it was before the upgrade (see comments below). 

*  Reverted [this change](https://issues.apache.org/jira/browse/CALCITE-6586) due to regressions in planner tests. Opened a [followup issue](https://bodo.atlassian.net/jira/software/c/projects/BSE/boards/25?assignee=712020%3A75fb3a19-d187-4101-b627-42e0d56f6b97&selectedIssue=BSE-5568) to reenable.  

## Testing strategy

Java tests, [Nightly CI](https://github.com/bodo-ai/Bodo/actions/runs/31522458504)

## Checklist


BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/application/logicalRules/BodoSQLReduceExpressionsRule.java
```diff
@@ -115,10 +113,6 @@ protected FilterReduceExpressionsRule(
     @Override
     public void onMatch(RelOptRuleCall call) {
       final Filter filter = call.rel(0);
-      // Bodo Change: Do not reduce filters that contain Case statements.
```
scott-routledge2, line 118: I believe the original intention of the rule was to prevent the planner from simplifying a CASE statement inside a filter to something else and preventing our extract-case-out-to-projection from running. 

In newer versions of calcite, not simplifying case statements here results in the planner getting stuck in an infinite loop for some trivial case statement like:

```
CASE (A)
  WHEN 'x' THEN '1'
  WHEN 'y' THEN '1'
  WHEN 'z' THEN '1'
```

Ideally, I'd like to understand a bit more about why there is an infinite loop in this case. It might also be the case we are generating 1000s of matches which makes the query slow but it might complete after an hour or more.

BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/schema/CatalogSchema.java
```diff
@@ -85,7 +85,7 @@ public Set<String> getTableNames() {
    * schema depth doesn't contain any tables.
    *
    * @param name Name of the table to load.
-   * @return Table object.
```
scott-routledge2, line 88: Calcite requires returning null if a table/subSchema is not found in a given schema. The upgrade exposed this issue by changing the order table lookup happens in. 

These methods are also being deprecated and will be removed in calcite 2.0, I'm leaving implementing the new methods as a followup https://bodo.atlassian.net/browse/BSE-5563

BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/schema/RootSchema.kt
```diff
@@ -107,9 +107,10 @@ class RootSchema {
                     rootMap[catalog] = newSchema
                     newSchema
                 }
-            // Wrap in a CachingCalciteSchema
-            val cachedSchema = CalciteSchema.createRootSchema(false, true, rootName, schema)
-            return cachedSchema.plus()
+            // Wrap in a SimpleCalciteSchema
+            // Disable caching because it can be very expensive to initially populate the cache for large schemas
+            val simpleSchema = CalciteSchema.createRootSchema(false, false, rootName, schema)
```
scott-routledge2, line 113: Calcite's new implementation of CachingSchema seems to be populating a mapping of table name -> actual table entry for every table name in our schema when it is first used, which makes our snowflake tests timeout.

We already do caching (cache names + table entries when we look them up) in the CatalogSchema, so this change shouldn't affect performance.

BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/schema/CatalogSchema.java
```diff
@@ -139,13 +133,14 @@ public CatalogSchema getSubSchema(String schemaName) {
     if (this.subSchemaMap.containsKey(schemaName)) {
       return this.subSchemaMap.get(schemaName);
     }
-    CatalogSchema schema = this.catalog.getSchema(getFullPath(), schemaName);
-    if (schema == null) {
-      throw new RuntimeException(
```
scott-routledge2, line 144: Our implementation of getSchema for all our catalogs returns a valid schema object, even if that schema does not exist. Ideally, they should return null when a schema does not exist. 

Previous versions of calcite would do something like call getSubSchemaNames (or load from cache) -> if name not found in sub-schema names, return null otherwise call getSubSchema. After the upgrade, getSubSchema is called directly.

BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/application/logicalRules/BodoSQLReduceExpressionsRule.java
scott-routledge2: Confirmed this was related to https://issues.apache.org/jira/browse/CALCITE-6586, will revisit in a followup

DrTodd13 (approved):
Thanks!  We don't think the new "checked" versions of arithmetic operators are going to suddenly start appearing as a result of our existing queries?  I guess if it passes CI the answer is no but would be nice to know if our inputs could ever possibly generate the checked versions and since we don't support them we become more fragile with this possibility.  Anyway, minor stuff just wondering.

scott-routledge2:
> Thanks! We don't think the new "checked" versions of arithmetic operators are going to suddenly start appearing as a result of our existing queries? I guess if it passes CI the answer is no but would be nice to know if our inputs could ever possibly generate the checked versions and since we don't support them we become more fragile with this possibility. Anyway, minor stuff just wondering.

My understanding is that the checked arithmetic are only for the BigQuery and SQLServer2008 dialects, so we shouldn't generate them.

BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/schema/RootSchema.kt
ehsantn: Do we need need a follow up issue and more comments about this?

ehsantn (approved):
Thanks @scott-routledge2 !

BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/schema/RootSchema.kt
scott-routledge2: There is a followup to implement the new lookup API (https://bodo.atlassian.net/jira/software/c/projects/BSE/boards/25?assignee=712020%3A75fb3a19-d187-4101-b627-42e0d56f6b97&selectedIssue=BSE-5563), which depending on how it's implemented might fix this issue for us. 

I'll link to it in the comments and include some more of the details I mention here.

IsaacWarren (approved):
Thanks @scott-routledge2!

PR #1336, opened 2026-08-05
Add cast options infrastructure for CastExpression in C++ backend

dmsa-or:

We now pass a cast options object (`CastOptions` or `BodoStringCastOptions`) through `CastExpression`, which controls which cast function should be called ("cast" or "bodo_string_cast"). This also exposes cast options like `allow_int_overflow` and `allow_float_truncate` to Python that we were applying unilaterally because we had no way to selectively enable them for the few cases where we actually needed them. I've kept those as defaults for now so I don't accidentally break things that were relying on them, but ideally they would be off unless specifically enabled to avoid silent correctness errors where int overflow or float truncation are not expected.

## Testing strategy

Tested locally for try-casting. The plan conversions for that aren't in this PR. Since all `CastExpression`s now go through this code path, I'll run Nightly to make sure nothing accidentally broke.

## User facing changes

The new `CastExpression` API should be backwards compatible with the existing syntax where cast options are not explicitly passed. If cast options are passed, the class symbolizes which cast function should be called, and also contains options for that cast function.

## Checklist

### DIFF
--- a/bodo/pandas/plan_optimizer.pyx
+++ b/bodo/pandas/plan_optimizer.pyx
@@ -3,7 +3,8 @@
 # distutils: language = c++
 # cython: c_string_type=unicode, c_string_encoding=utf8
 
-from libcpp.memory cimport unique_ptr, make_unique, dynamic_pointer_cast
+from cython.operator cimport dereference as deref
+from libcpp.memory cimport shared_ptr, make_shared, unique_ptr, make_unique, static_pointer_cast, dynamic_pointer_cast
 from libcpp.utility cimport move, pair
 from libcpp.string cimport string as c_string
 from libcpp.vector cimport vector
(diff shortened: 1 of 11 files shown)


bodo/pandas/vendor/duckdb/src/include/duckdb/function/cast/default_casts.hpp
```diff
@@ -114,14 +114,15 @@ struct BoundCastInfo {
 	DUCKDB_API
 	BoundCastInfo( // NOLINT: allow explicit cast from cast_function_t
 	    cast_function_t function, unique_ptr<BoundCastData> cast_data = nullptr,
-		// Bodo change: added arrow_type
-	    init_cast_local_state_t init_local_state = nullptr, std::shared_ptr<arrow::DataType> arrow_type = nullptr);
+		// Bodo change: added arrow_cast_opts
```
DrTodd13, line 117: Is there no extraneous data field we can use with some anonymous pointer rather than changing duckdb here?

DrTodd13 (approved):

bodo/pandas/plan_optimizer.pyx
```diff
@@ -20,11 +21,190 @@ import numpy as np
 import pyarrow as pa
 import bodo
 
+from pyarrow._compute cimport FunctionOptions
+from pyarrow.lib cimport pyarrow_wrap_data_type, pyarrow_unwrap_data_type
+from pyarrow.includes.libarrow cimport CDataType, CFunctionOptions, CCastOptions
```
ehsantn, line 26: I'm wondering if these imports work in our package builds too. Let's run the release build to make sure.

ehsantn (approved):
Looks good to me overall. Thanks @dmsa-or .

bodo/pandas/vendor/duckdb/src/include/duckdb/function/cast/default_casts.hpp
dmsa-or: I don't think there's an obvious way. Ehsan had already created this field before, so I thought it wasn't a big deal to switch it to FunctionOptions.

DrTodd13: In that case fine.

bodo/pandas/plan_optimizer.pyx
dmsa-or: Release build succeeds: https://github.com/bodo-ai/Bodo/actions/runs/31281778483

ehsantn:
Nightly passed except flaky tests: https://github.com/bodo-ai/Bodo/actions/runs/32862796028

PR #1338, opened 2026-08-05
Second batch of changes for iceberg tpch and decimal support.

DrTodd13:

1. Add use_iceberg option to bodo tpch dataframe_queries.py
2. Add support for precision and scale in arrow_reader, array allocation routines in _bodo_common.cpp.
3. Add support for decimal divide and comparisons in _decimal_ext.cpp.
4. Create two routines that operate on two arrays of decimals and returns another decimal or bool array respectively.
5. Create generic routine that takes two decimals arrays and an operation string and returns a result array.
6. Modified do_arrow_compute_binary in expression.cpp to determine if a decimal operation can potentially overflow and if so to use the potentially slower routines in _decimal_ext.
7. Better support for datatype to decimal conversion in plan_conversion.py.

## Testing strategy

run_ci

## User facing changes

None

## Checklist


bodo/pandas/_util.cpp
ehsantn: Have you verified this? DuckDB's other physical types set different values of the underlying union. Maybe this is correct because the rest of the union is guaranteed to be zeros?

https://github.com/bodo-ai/Bodo/blob/b6831ac9551f6cbb7fd6a4a50bb281021d0265e7/bodo/pandas/vendor/duckdb/src/common/types/value.cpp#L606

bodo/pandas/_util.cpp
```diff
@@ -1934,6 +1938,291 @@ size_t col_ref_map_lookup(
...
+            return std::make_pair<int, int>(5, 0);
+        case arrow::Type::INT32:
+        case arrow::Type::UINT32:
+            return std::make_pair<int, int>(10, 0);
+        case arrow::Type::INT64:
+        case arrow::Type::UINT64:
+            return std::make_pair<int, int>(19, 0);
+        case arrow::Type::FLOAT:
+        case arrow::Type::DOUBLE:
+            return precision_scale_from_double_array(input.make_array());
```
ehsantn, line 2136: Does Snowflake actually look into double values to set precision and scale? Doesn't seem likely to me.

bodo/pandas/_util.cpp
```diff
@@ -1934,6 +1938,291 @@ size_t col_ref_map_lookup(
...
+    arrow::Decimal128Builder builder(arrow::decimal128(precision, scale),
+                                     arrow::default_memory_pool());
+
+    arrow::Status status;
+    status = builder.Reserve(n);
+    if (!status.ok()) {
+        throw std::runtime_error("ConvertIntArrayToDecimal128 Reserve failed.");
+    }
+
+    for (int64_t i = 0; i < n; ++i) {
```
ehsantn: We can't just call an Arrow cast or conversion function? This loop can be quite slow I think.

bodo/libs/_decimal_ext.cpp
```diff
@@ -2928,3 +2955,279 @@ void decimal_to_str(uint64_t in_low, int64_t in_high, NRT_MemInfo** meminfo_ptr,
...
+        if constexpr (rescale_left) {
+            lhs = d1.Rescale(s1, out_scale).ValueOrDie();
+        }
+        if constexpr (rescale_right) {
+            rhs = d2.Rescale(s2, out_scale).ValueOrDie();
+        }
+        *result = Op(lhs, rhs);
+    } else {
+        // Ensure the two decimals have the same scale (but are first upcasted
+        // to 256)
```
ehsantn, line 2979: Do others like Snowflake upcast for these operations instead of ignoring the lower digits? I'm wondering if this is too expensive generally.

bodo/libs/_decimal_ext.cpp
```diff
@@ -2928,3 +2955,279 @@ void decimal_to_str(uint64_t in_low, int64_t in_high, NRT_MemInfo** meminfo_ptr,
...
+    status = builder.Finish(&out_arr);
+    if (!status.ok()) {
+        throw std::runtime_error(
+            "arrow_array_decimal_arithmetic error in builder.Finish");
+    }
+    return out_arr;
+}
+
+template <auto Op>
+std::shared_ptr<arrow::Array> arrow_array_boolean_op(
```
ehsantn, line 3133: There are no Arrow comparison kernels (including in Gandiva) to pick up?

ehsantn (commented):
See comments. Generally, there are a lot of manual kernels here that could have reliability and performance implications. Are there no Arrow/Gandiva kernels to reuse?

bodo/libs/_decimal_ext.cpp
DrTodd13: Arrow comparison kernels generate decimal overflow.  Didn't see any comparison kernels in Gandiva either unless I missed them somehow.

bodo/libs/_decimal_ext.cpp
DrTodd13: This code is styled off of what you do in the same file for arithmetic.  So, yes, it could be you want a different approach for arithmetic and boolean comparisons but I'd think the default position is to be consistent.  It is either expanding the type or giving approximate answers and the latter it seems you'd need significant justification which we don't have because we have no performance numbers.

bodo/pandas/_util.cpp
DrTodd13: No, it doesn't.  From what I can tell, it gives the same output type as the input decimal type.  I wouldn't say I added a bunch of kernels, in fact I'd say I was trying to avoid them like not having mixed-type kernels of decimal and int/float.  Arrow seems to have double output type of decimal*double which we could cast afterwards.  In general, we have too many type systems and they all have different rules.  Having to understand and code particularly the decimal type behavior of c++ code in python is going to be fragile.

bodo/pandas/_util.cpp
DrTodd13: done

bodo/pandas/_util.cpp
DrTodd13: I simplified the plan_conversion side to not look at the float values but keep the same decimal precision and scale as the decimal input.  From what I can tell, that is the snowflake behavior.  I'm not sure we can simplify the execution side much.  Arrow is definitely treating doubles as precision 19 for some purposes because if you have decimal(20,2) times double then you get a decimal precision out of range 39 error even though the output type for decimal(19,2)*double is double.  I can look into whether we can treat double as decimal(19,0) so we don't have to look at the input but if we still want to use our _decimal_ext kernels then they both need to be decimal so we need the convert of double to decimal.  Figuring out the minimal precision/scale would let us use the arrow kernels more versus the potentially slower _decimal_ext ones so that is a tradeoff.

bodo/pandas/_util.cpp
DrTodd13: fixed.

bodo/libs/_decimal_ext.cpp
```diff
@@ -2928,3 +2955,279 @@ void decimal_to_str(uint64_t in_low, int64_t in_high, NRT_MemInfo** meminfo_ptr,
...
+}
+
+// -------------------Decimal comparisons ---------------------------------
+
+template <size_t N>
+struct ct_string {
+    char value[N];
+    constexpr ct_string(const char (&s)[N]) {
+        for (size_t i = 0; i < N; ++i)
+            value[i] = s[i];
```
scott-routledge2:
```suggestion
        for (size_t i = 0; i < N; ++i) {
            value[i] = s[i];
        }
```

scott-routledge2 (approved):
LGTM, thanks @DrTodd13

bodo/pandas/_util.cpp
ehsantn: Ok, thanks for looking into this.

ehsantn (approved):
Thanks for the effort @DrTodd13.

PR #1344, opened 2026-08-08
Support TO_TIMESTAMP for numeric input + enable passing cast function tests

dmsa-or:

Supported two-argument TO_TIMESTAMP and variations for numeric input in BodoSQL C++ backend.
Added `fixture_id_is_in` and `fixture_id_not_in` functions for use with the `mark_bodosql_cpp_if` decorator.
Defined globals `SqlTypeName` and `SqlKind` in plan_conversion.py since it was getting repetitive making an aliases for them everywhere.

## Testing strategy

PR CI and Nightly

## Checklist


DrTodd13 (approved):

bodo/spawn/spawner.py
```diff
@@ -403,14 +403,26 @@ def handler(*args, **kwargs):
                     exceptions.append(excep)
 
                 # Combine all exceptions into a single chain
+                exception_strings = []
                 accumulated_exception = None
                 for excep in exceptions:
+                    # Collect the rank error messages and notes to add
```
scott-routledge2, line 409: Do you have an example of what the new exception message looks like after this change?

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -2543,7 +2550,12 @@ def calculate_scale_factor(current_unit, target_unit):
                 (0, "half_towards_infinity"),
             )
 
-        # Integers are assumed in seconds in BodoSQL (instead of nanoseconds as converted by sql_type_to_pa_type())
+        # Integers are assumed in seconds in BodoSQL (instead of nanoseconds as converted by sql_type_to_pa_type()).
+        # TODO: Are they actually? Calcite converts the single-argument TO_TIMESTAMP to regular cast to timestamp.
```
scott-routledge2, line 2552: Is this comment saying that we treat integers as seconds for to_timestamp but cast to timestamp has a different behavior? Can we open a followup issue with a reproducer?

We may want to just update the docs here if the current behavior doesn't match what's documented.

scott-routledge2 (approved):
LGTM, thanks @dmsa-or !

bodo/spawn/spawner.py
dmsa-or: Before: `C++ backend execution failed with error: Unexpected Internal Error: Some ranks failed`
After:
```
C++ backend execution failed with error: Unexpected Internal Error: Some ranks failed:
do_arrow_compute_cast: Error in Arrow compute: Failed to parse string: '.' as a scalar of type double
  ^ From rank 0
do_arrow_compute_cast: Error in Arrow compute: Failed to parse string: '2e' as a scalar of type double
  ^ From rank 1
do_arrow_compute_cast: Error in Arrow compute: Failed to parse string: '4-7' as a scalar of type double
  ^ From rank 2
do_arrow_compute_cast: Error in Arrow compute: Failed to parse string: ' - 0.73 ' as a scalar of type double
  ^ From rank 3
```
If you printed the full traceback of the exception, you would've already been able to see the rank error messages, but  now they are more visible. Also, now that they are part of the top-level message, we can pass tests like `test_to_double_invalid_cols` that search for a specific error message, even with parallel on.

BodoSQL/bodosql/plan_conversion.py
dmsa-or: This is saying that for single-arg TO_TIMESTAMP, the C++ backend treats integers as seconds. But that is not always the case for the JIT backend, if the BodoSQL docs are to be believed (I haven't checked the actual code).

PR #1347, opened 2026-08-10
Next batch of iceberg tpch sql fixes.

DrTodd13:

1. Fix bug with not filling in array with nulls.
2. Allow overflow.
3. Better decimal scalar support.
4. More common routines for getting datum precision and scale and operation precision and scale.
5. Decimal reduce support.

## Testing strategy

run_ci

## User facing changes

None

## Checklist


bodo/libs/_decimal_ext.cpp
```diff
@@ -3232,3 +3230,79 @@ std::shared_ptr<arrow::Array> arrow_array_decimal_arithmetic_util(
...
+            left_val, left_precision, left_scale, right_val, right_precision,
+            right_scale, result_precision, result_scale, &overflow);
+    } else {
+        throw std::runtime_error(
+            "arrow_array_decimal_arithmetic does not support operation " + op);
+    }
+
+    if (overflow) {
+        return nullptr;
+    }
```
ehsantn: Allow overflow here to match the other case?

ehsantn (approved):
Thanks @DrTodd13 .

scott-routledge2 (commented):
Looks like CI is failing due to some decimal issues https://github.com/bodo-ai/Bodo/actions/runs/31433735402/job/93608565267?pr=1347#step:11:3144

bodo/libs/_decimal_ext.cpp
DrTodd13: there were two other cases actually.  fixed.

BodoSQL/bodosql/plan_conversion.py
```diff
@@ -5724,32 +5742,43 @@ def java_binop_to_python_expr(ctx, kind, op_name, op_exprs):
...
                 lambda l, r: l - r,
-                lambda ll, ls, rl, rs: (max(ll, rl) + 1, max(ls, rs)),
+                decimal_addition_subtraction_output_precision_scale,
             )
             expr = ArithOpExpression(out_empty, left, right, "__sub__")
         return expr
 
     if kind.equals(SqlKind.TIMES):
+        from bodo.libs.decimal_arr_ext import (
+            decimal_multiplication_output_precision_scale,
```
scott-routledge2, line 5760: What is the current status of JIT import in the C++ backend? 

I think this line will import the compiler which is unnecessary. We might want to refactor these utilities to some other file that does not import the compiler. 

Although if we are importing the compiler in other places it's probably fine to leave as is for now.

scott-routledge2 (approved):
Thanks @DrTodd13 !

BodoSQL/bodosql/plan_conversion.py
DrTodd13: Went ahead and bit the bullet and moved them to a new file in bodo.utils.decimal_utils.

PR #1350, opened 2026-08-12
[GPU] Fix BodoSQL GPU env initialization

ehsantn:
Fixes scatterv hang in GPU builds. BODO_GPU is on by default in GPU builds so no other action is required to avoid user errors:
https://github.com/bodo-ai/Bodo/blob/f5dddc6976e4fe27cc261b66aab9c3dd7fad69f7/buildscripts/bodo/conda-recipe/build.sh#L21

### DIFF
--- a/BodoSQL/bodosql/__init__.py
+++ b/BodoSQL/bodosql/__init__.py
@@ -1,5 +1,9 @@
 import os
 
+# Import bodo first to ensure bodo initialization is done before any other imports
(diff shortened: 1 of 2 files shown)


DrTodd13 (approved):
Approved with the addition of what we discussed in stand-up today about setting the variable automatically for GPU builds on Linux.

dmsa-or (approved):
Thanks @ehsantn

PR #1351, opened 2026-08-12
[BSE-5565][GPU] Make cached subplans more robust in C++ backend

scott-routledge2:

* `join_state_map` now stores a shared_ptr to a `join_state` instead of the raw pointer to ensure the join state does not get freed before the filter is used. This is mainly done for robustness; ideally this scenario should never happen but it currently can happen when a join filter occurs inside a cached node that does not later get turned into a CTE. 

* Reset subplan cache after each plan execution in `BodoSQLContext` to avoid getting a cached subplan from previous plan executions. 

## Testing strategy
CI , defog Iceberg queries. 

## Checklist

### DIFF
--- a/BodoSQL/bodosql/context.py
+++ b/BodoSQL/bodosql/context.py
@@ -727,6 +727,7 @@ class BodoSQLContext:
             # Keeps track of join ids and their join filter key locations for join
             # filter translation during conversion to Python plan.
             self.join_filter_info = {}
+            self.subplan_cache = {}
             # Temporarily monkey-patch so java_plan_to_python_plan
             # can see dynamic and named params.
             self.named_params_dict = (java_named_params_map, named_params_dict)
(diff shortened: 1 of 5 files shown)


BodoSQL/bodosql/context.py
```diff
@@ -727,6 +727,7 @@ def execute_cpp_backend(
             # Keeps track of join ids and their join filter key locations for join
             # filter translation during conversion to Python plan.
             self.join_filter_info = {}
+            self.subplan_cache = {}
```
scott-routledge2, line 730: Unrelated bug I found while testing my PR. 

If running back to back queries with cached nodes in the same bodosql context, we need to reset the cache, otherwise plan 2 will use cache node 0 from plan 1.

DrTodd13 (approved):
Thanks!

ehsantn (approved):
Thanks @scott-routledge2.

PR #1354, opened 2026-08-13
DF lib iceberg tpch fixes.

DrTodd13:

1. Checks and conversions in dataframe_queries where we have real constant in an arithmetic op with a decimal column.
2. Fix decimal divide scale.
3. Streaming groupby mean generates float when input is decimal.
4. Be able to generate duckdb constant expressions from arrow scalars.
5. Decimal arrow scalar to duckdb value.
6. Support optional filters in duckdb to pyiceberg.
7. Bodo series numeric operations tries regular pandas op on empty tables to generate output table but if decimal overflow happens then call same get_output_type function that plan_conversion uses.

## Testing strategy

run_ci

## User facing changes

None

## Checklist


bodo/pandas/_util.cpp
```diff
@@ -390,6 +390,23 @@ duckdb::Value ArrowScalarToDuckDBValue(
...
+        case arrow::Type::DECIMAL: {
+            auto dec_scalar =
+                std::static_pointer_cast<arrow::Decimal128Scalar>(scalar);
+            arrow::Decimal128 dec_value = dec_scalar->value;
+            auto dec_type = std::static_pointer_cast<arrow::Decimal128Type>(
+                dec_scalar->type);
+            int32_t precision = dec_type->precision();
+            int32_t scale = dec_type->scale();
+            int64_t high_bits = dec_value.high_bits();
+            if (high_bits != 0) {
```
DrTodd13: Not sure we have a great way to handle 128-bit values here.  It seems like a hugeint constructor from string would be required for that but perhaps in our paring down our duckdb vendored code we removed that constructor.  Anyway, unlikely in practice to need more than 64-bits here.

ehsantn: Fixed this using `duckdb::hugeint_t`.

ehsantn (approved):

ehsantn:
> ## Changes included in this PR
> 1. Checks and conversions in dataframe_queries where we have real constant in an arithmetic op with a decimal column.
> 2. Fix decimal divide scale.
> 3. Streaming groupby mean generates float when input is decimal.

@DrTodd13 I don't see 1,2,3 items in the code changes.

bodo/pandas/utils.py
```diff
@@ -1706,3 +1706,66 @@ def getUnusedColumnName(columns):
...
+    right_atype,
+    non_decimal_func,
+    decimal_func,
+):
+    """Get the output of binary operation considering decimal input cases
+
+    Args:
+        left_empty (DataFrame/Series): Empty dataframe/series with type of left hand side
+        left_atype (ArrowType): Arrow type of left hand side
+        right_empty (DataFrame/Series): Empty dataframe/series with type of right hand side
```
scott-routledge2, line 1748: This function will fail if Series is passed for non-decimal case.

bodo/pandas/series.py
```diff
@@ -371,12 +372,64 @@ def _numeric_binop(self, other, op, reverse):
...
+            except Exception:
+                if type(other) in (BodoSeries, BodoScalar):
+                    left_atype = zero_size_self.dtype.pyarrow_dtype
+                    right_atype = zero_size_other.dtype.pyarrow_dtype
+                    decimal_precision_scale_func = None
+                    if pa.types.is_decimal(left_atype) and pa.types.is_decimal(
+                        right_atype
+                    ):
+                        if op in ("__add__", "__radd__", "__sub__", "__rsub__"):
+                            from bodo.libs.decimal_arr_ext import (
```
```suggestion
                            from bodo.utils.decimal_utils import (
```

Avoid importing JIT here? (And other cases below)

ehsantn:
@IsaacWarren @scott-routledge2 this is ready for review.

scott-routledge2 (approved):
LGTM

bodo/pandas/series.py
```diff
@@ -371,12 +372,64 @@ def _numeric_binop(self, other, op, reverse):
...
             empty_data = zero_size_self
         else:
             # Compute schema of new series.
-            empty_data = getattr(zero_size_self, op)(zero_size_other)
+            try:
+                empty_data = getattr(zero_size_self, op)(zero_size_other)
+            # Pandas decimal binops may fail with Arrow's "Decimal precision out of
+            # range" error. We apply the same logic as BodoSQL to get the output
+            # precision and scale for decimal binops.
+            except Exception:
```
IsaacWarren, line 384: Scope this to only the decimal precision error?

IsaacWarren (approved):

bodo/pandas/utils.py
ehsantn: Yes, but current uses don't need it. It will be an easy error to fix if uses change.

bodo/pandas/series.py
ehsantn: Good catch. Fixed.

bodo/pandas/series.py
ehsantn: The next PR avoids exception catching for this. I'll open it today.

PR #1355, opened 2026-08-13
[BSE-5163][GPU] Support Sum0 initialization in reductions

scott-routledge2:

Adds SUM0 to reduction functions, which is the same as SumReductionFunction except it always initializes to 0. 

From what I can tell, there is no equivalent to this issue for Groupby, since SUM0 and SUM were equivalent in the plans I looked at. 

## Testing strategy

New tests added. I also verified the Pydough-generated defog queries pass now-- which is where this issue originated from. 

## Checklist


BodoSQL/bodosql/tests/test_tpch_first_half.py
```diff
@@ -34,7 +34,7 @@
             "iceberg",
             "s3://duckdb-iceberg-data-427443013497-us-east-2-an/tpch_sf1_iceberg_results/q01_output",
             id="iceberg",
-            marks=pytest.mark.xfail(
```
scott-routledge2, line 37: Changing from xfail to skip since on GPU, some of these tests were crashing, which causes everything after to fail.

bodo/pandas/physical/aggregate.h
```diff
@@ -389,6 +389,7 @@ const std::map<std::string, int32_t> PhysicalAggregate::function_to_ftype = {
     {"kurtosis", Bodo_FTypes::kurtosis},
     {"std", Bodo_FTypes::std},
     {"sum", Bodo_FTypes::sum},
+    {"sum0", Bodo_FTypes::sum},
```
scott-routledge2: From what I could tell, Calcite generates SUM0 (inside an aggregate with grouping keys) by pushing COUNT aggregates below a join. 

So Count() on top of a join would become SUM0 above the join and then below the join there would be a COUNT aggregation on both side. The SUM0 above the joins would sum the product of the lower counts. 

For example this query:
```
SELECT
    c_nationkey,
    COUNT(o_comment)
FROM customer
JOIN orders
    ON c_custkey = o_custkey
GROUP BY c_nationkey
```

Produces a plan:
```
CombineStreamsExchange
  BodoPhysicalAggregate(group=[{0}], EXPR$1=[$SUM0($1)])
    BodoPhysicalProject(C_NATIONKEY=[$3], $f5=[*($1, $4)])
      BodoPhysicalJoin(condition=[=($2, $0)], joinType=[inner], JoinID=[0])
        BodoPhysicalAggregate(group=[{0}], EXPR$1=[COUNT($1)])
          IcebergToBodoPhysicalConverter
            IcebergRuntimeJoinFilter(joinIDs=[[0]], equalityColumnsList=[[[0]]], allEqualityKeysReady=[[true]], nonEqualityColumnsList=[[[]]])
              IcebergProject(O_CUSTKEY=[$1], O_COMMENT=[$2])
                IcebergFilter(IS NOT NULL($1))])
                  IcebergTableScan(table=[[., ., ORDERS]], Columns=[[O_ORDERKEY, O_CUSTKEY, O_COMMENT]], Iceberg=[true])
        BodoPhysicalAggregate(group=[{0, 1}], agg#0=[COUNT()])
          IcebergToBodoPhysicalConverter
            IcebergFilter(condition=[IS NOT NULL($0)])
              IcebergTableScan(table=[[., ., CUSTOMER]], Columns=[[C_CUSTKEY, C_NATIONKEY]], Iceberg=[true])

```

However, since it turns COUNT into a SUM0 of COUNTs and COUNT can never be null, SUM and SUM0 are indistinguishable in this case. 

The difference between SUM0 and SUM matters when there are no aggregation keys, since the input to the SUM0 can be empty, which is not possible when grouping keys are present in the plan.  

------
For completeness, I can add it, however there would be no way to test it end to end and we'd have to rely on some manually constructed Python plan.

BodoSQL/bodosql/tests/test_agg_groupby.py
```diff
@@ -2092,3 +2094,47 @@ def test_groupby_agg_on_key_col(memory_leak_check):
         check_dtype=False,
         use_duckdb=True,
     )
+
+
+@pytest.mark.cpp_backend
+def test_sum0_in_lazy_plan():
```
scott-routledge2, line 2100: As far as I can tell it's currently not possible to write a SQL query that produces a SUM0 inside a groupby AND main and my branch produce different results (see my comment in aggregate.h for more details), but I added it for completeness and used a lazy plan instead of SQL to test the initialization behavior.

bodo/pandas/physical/gpu_aggregate.cpp
```diff
@@ -12,6 +12,8 @@ const std::map<std::string, int32_t> PhysicalGPUAggregate::function_to_ftype = {
     {"skew", Bodo_FTypes::skew},
     {"std", Bodo_FTypes::std},
     {"sum", Bodo_FTypes::sum},
+    // TODO: Support Pandas-style/sum0 initialization for GPU.
```
scott-routledge2: Currently we don't support Pandas-style semantics for GPU (everything is null initialized), there is no equivalent to "use sql style" in libcudf, as far as I can tell, and cudf returns NA when the input for a group is all NA for sum, which differs from Pandas.

DrTodd13 (approved):
LGTM.

ehsantn (approved):
Looks good to me.

PR #1381, opened 2026-09-03
Fix to_number decimal issues

ehsantn:
Fixes BodoSQL JIT backend issues for TPC-H Q17 with decimal input.


DrTodd13 (approved):
thanks

scott-routledge2 (approved):
LGTM

PR #1382, opened 2026-09-03
Avoid writing all data points in profile aggregation by default

ehsantn:
As title. Data points for every single rank are rarely necessary and make reading the output difficult.


DrTodd13 (approved):

scott-routledge2 (approved):
Thanks!

PR #1383, opened 2026-09-03
[BSE-5568] Enable Predicate Metadata Pullup for Rel Subets

scott-routledge2:

Remove change that prevented predicated metadata from being propagated from Rel Subsets in the Volcano planner, enabling additional optimizer rules such as `ProjectReduceExpressionRule`.

Also fixes some existing bugs exposed by predicate metadata changes and removes some changes to predicates that were merged upstream. 
  
## Testing strategy

PR CI; Java plan tests (See: https://github.com/bodo-ai/customer-sample-code/pull/161)

## Checklist

### DIFF
--- a/BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/application/logicalRules/BodoSQLReduceExpressionsRule.java
+++ b/BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/application/logicalRules/BodoSQLReduceExpressionsRule.java
@@ -17,8 +17,6 @@ package com.bodosql.calcite.application.logicalRules;
  * limitations under the License.
  */
 
-import static com.bodosql.calcite.application.logicalRules.FilterRulesCommon.rexNodeContainsCase;
-
 import com.bodosql.calcite.application.utils.BodoSQLStyleImmutable;
 import com.bodosql.calcite.application.utils.RexNormalizer;
 import com.bodosql.calcite.rel.logical.BodoLogicalProject;
@@ -115,10 +113,6 @@ public abstract class BodoSQLReduceExpressionsRule<C extends BodoSQLReduceExpres
     @Override
     public void onMatch(RelOptRuleCall call) {
       final Filter filter = call.rel(0);
-      // Bodo Change: Do not reduce filters that contain Case statements.
-      if (rexNodeContainsCase(filter.getCondition())) {
-        return;
-      }
       final List<RexNode> expList = Lists.newArrayList(filter.getCondition());
       RexNode newConditionExp;
       boolean reduced;
--- a/BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/application/logicalRules/BodoSQLReduceExpressionsRule.java
+++ b/BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/application/logicalRules/BodoSQLReduceExpressionsRule.java
@@ -294,24 +294,12 @@ public abstract class BodoSQLReduceExpressionsRule<C extends BodoSQLReduceExpres
                   .collect(Collectors.toList());
           boolean changed = !project.getProjects().equals(finalExpList);
           if (changed) {
-            RelNode newProject =
+            call.transformTo(
                 call.builder()
                     .push(project.getInput())
                     .project(finalExpList, project.getRowType().getFieldNames())
-                    .build();
-            call.transformTo(newProject);
-            //            for (int i = 0; i < project.getProjects().size(); i++) {
-            //              RexNode before = project.getProjects().get(i);
-            //              RexNode after = finalExpList.get(i);
-            //
-            //              if (before instanceof RexInputRef
-            //                      && !(after instanceof RexInputRef)
-            //                      && !before.equals(after)) {
-            //                System.err.println("INPUT REF EXPANDED:");
-            //                System.err.println("  before: " + before);
-            //                System.err.println("  after:  " + after);
-            //              }
-            //            }
+                    .build());
+
             // New plan is absolutely better than old plan.
             call.getPlanner().prune(project);
           }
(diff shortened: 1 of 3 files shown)


BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/application/logicalRules/BodoSQLReduceExpressionsRule.java
```diff
@@ -115,10 +113,6 @@ protected FilterReduceExpressionsRule(
     @Override
     public void onMatch(RelOptRuleCall call) {
       final Filter filter = call.rel(0);
-      // Bodo Change: Do not reduce filters that contain Case statements.
```
scott-routledge2: This seems wrong to me but I still don't know why. 

It causes an explosion of rules which either causes a hang or is just really slow in some planner tests involving case statements.

BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/rel/metadata/BodoRelMdPredicates.java
```diff
@@ -314,14 +272,12 @@ public RelOptPredicateList inferPredicates(boolean includeEqualityInference) {
       Mappings.TargetMapping rightMapping =
           Mappings.createShiftMapping(
               nSysFields + nFieldsLeft + nFieldsRight, 0, nSysFields + nFieldsLeft, nFieldsRight);
-      // BODO CHANGE: uses TypeChangingRexPermuteInputsShuttle instead of the default
```
scott-routledge2, line 317: This change is no longer necessary since it was fixed in the upstream

BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/rel/metadata/BodoRelMdPredicates.java
```diff
@@ -53,12 +51,7 @@
 import org.apache.calcite.util.mapping.Mappings;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
-/**
- * Utility to infer Predicates that are applicable above a RelNode.
- *
- * <p>The only change that Bodo makes is to Join predicate inference. This change can be undone
```
scott-routledge2, line 59: This comment is out of date. The PR mentioned was closed and an equivalent one was merged. We have since made some other changes in this file but they have already been annotated with Bodo Change.

BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/application/logicalRules/BodoSQLReduceExpressionsRule.java
scott-routledge2: Simplifying case statements here is necessary to enable normalization of case statements and prevent "rule explosions" which can look like hangs in the optimizer.  

I am assuming, based on the PR this was originally added for testing purposes and probably not necessary.

BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/application/logicalRules/BodoSQLReduceExpressionsRule.java
```diff
@@ -115,10 +113,6 @@ protected FilterReduceExpressionsRule(
...
-      if (rexNodeContainsCase(filter.getCondition())) {
```
scott-routledge2: I think removing is the right thing to do here to prevent an infinite loop inside the optimizer. 

Consider a simple query like this one (minimal structure to reproduce the hang):

```
SELECT colA
FROM t1
WHERE colA =
    CASE 1
        WHEN 1 THEN 'xxxx'
        WHEN 2 THEN 'yyyy'
    END;
```

Our input plan to volcano will look something like:
```
project col0=$0
   filter $0  = $1
    project col0=$0, col1=CASE...
```
Because of our case-in-filter extracting HEP steps. 

But since the case is constant WRT the rows, our predicate metadata will simplify the filter to become:

```
filter $0  = $1 -> filter $0  = CASE...
```

But since we have this guard, we can't simplify this filter further, leading to our filter extract rule to recreate the input structure, causing a loop/explosion:

```
project col0=$0
   filter $0  = CASE...
    project col0=$0, col1=CASE...

-> 

project col0=$0,
   filter $0  = $1
    project col0=$0, col1=CASE...
     project col0=$0, col1=CASE...
```
Which will continue adding projections and merging them, generating the filter and extracting it, etc...

BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/application/logicalRules/BodoSQLReduceExpressionsRule.java
```diff
@@ -153,6 +153,13 @@ public void onMatch(RelOptRuleCall call) {
         return;
       }
 
+      // Bodo change: Do not rewrite a filter to introduce a CASE expression.
```
scott-routledge2, line 156: Fixes an existing bug that was previously not possible because of empty predicates. But now it is needed to prevent an infinite optimizer cycle/explosion:

Consider a simple query like this one:

```
SELECT colA
FROM t1
WHERE colA =
    CASE 1
        WHEN 1 THEN 'xxxx'
        WHEN 2 THEN 'yyyy'
    END;
```

Our input plan to volcano will look something like:
```
project col0=$0
   filter $0  = $1
    project col0=$0, col1=CASE...
```
because of our case-in-filter extracting HEP steps. 

Since the case is constant WRT the rows, our predicate metadata will simplify the filter to become:

```
filter $0  = $1 -> filter $0  = CASE...
```

but we are basically undoing what our filter extract rule does, and it will eventually extract the case again, creating an explosion/infinite loop:

```
project col0=$0
  filter $0  = CASE...
    project col0=$0, col1=CASE...

-> 
project col0=$0
  project col0=$0,
    filter $0  = $1
      project col0=$0, col1=CASE...
        project col0=$0, col1=CASE...
```
which will continue adding projections and merging them, generating the filter and extracting it, etc...

ehsantn: Let's add this comment to the code for context.

ehsantn (approved):

IsaacWarren (approved):

PR #1390, opened 2026-09-08
Prefetch join hash table bucket

IsaacWarren:
## Summary

- Adds a `prefetch_bucket` method to the vendored ankerl hash map that issues a read-only `__builtin_prefetch` for the bucket cache line `find()` will touch
- Adds a software-prefetch pass to all 4 join probe loops: while probing row `i`, the bucket line for row `i + D` (D=16, tunable via `BODO_JOIN_PROBE_PREFETCH_DIST`) is prefetched, hiding part of the dependent cache-miss chain that dominates `ht_probe_time`
- Hoists three per-row pin-guard resolutions in `produce_probe_output` out of the loop (resolved once per batch, passed as raw pointers)

The join probe hot loop is a pointer chase: hash -> bucket line -> entry/key line. Hardware prefetchers cannot help (non-stride). `ht_probe_time` was ~30% of the probe stage, the largest join-side timer.

## Performance findings

All benchmarks: 8 workers, TPC-H SF10, interleaved A/B rounds, stage metrics from `BODO_TRACING_OUTPUT_DIR` (wall-clock medians; medians of 3 runs for the microbench).

### Pure join microbenchmark (lineitem 60M x orders 15M int keys)

| prefetch distance | `ht_probe_time` / rank-exec | delta |
|---|---|---|
| off | 0.457 / 0.443s | - |
| 8 | 0.319 / 0.299s | -31% |
| **16 (default)** | 0.304 / 0.293s | **-34%** |
| 32 | 0.303 / 0.303s | -33% |

Probe stage total -13%.

### Full TPC-H SF10, all 22 queries

| metric | off | on (D=16) | delta |
|---|---|---|---|
| wall, sum of per-query best | 92.74s | 82.98s | **-10.5%** |
| `ht_probe_time` | 0.347s | 0.218s | **-37%** |
| probe stage total | 1.017s | 0.835s | **-18%** |

### DIFF
--- a/bodo/libs/streaming/_join.cpp
+++ b/bodo/libs/streaming/_join.cpp
@@ -680,6 +680,51 @@ inline size_t handle_probe_input_for_partition(
     return group_id;
 }
 
+// Prefetch distance (in rows) for the join probe hot loop: while looking up
+// row i, the bucket line of row i + JOIN_PROBE_PREFETCH_DIST is prefetched,
+// hiding part of the dependent cache miss chain. 0 disables prefetching.
+static const size_t JOIN_PROBE_PREFETCH_DIST = [] {
+    const char* dist_str = std::getenv("BODO_JOIN_PROBE_PREFETCH_DIST");
+    if (dist_str == nullptr) {
+        return size_t(16);
+    }
+    int dist = std::atoi(dist_str);
+    return dist > 0 ? static_cast<size_t>(dist) : size_t(0);
+}();
+
+/**
+ * @brief Prefetch the hash table bucket cache line of a probe row ahead of
+ * its lookup (see handle_probe_input_for_partition), to partially hide the
+ * dependent cache miss of the bucket load.
+ *
+ * @param ht Partition's pinned hash table to prefetch.
+ * @param i_row Row index in partition->probe_table to prefetch.
+ */
+inline void prefetch_probe_input_for_partition(
+    const bodo::pin_guard<bodo::pinnable<JoinPartition::hash_table_t>>& ht,
+    size_t i_row) {
+    ht->prefetch_bucket(-i_row - 1);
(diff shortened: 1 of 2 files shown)


DrTodd13 (approved):
LGTM.  thanks

bodo/libs/streaming/_join.cpp
```diff
@@ -3593,6 +3631,10 @@ bool join_probe_consume_batch(HashJoinState* join_state,
     const auto& active_partition_ht =
         active_partition.get()->build_hash_table_guard.value();
     for (size_t i_row = 0; i_row < in_table->nrows(); i_row++) {
+        if (i_row + JOIN_PROBE_PREFETCH_DIST < in_table->nrows()) {
```
ehsantn, line 3634: Can this cause a slowdown with a lot of cores and therefore data mostly shuffled or if most of the data is filtered by the bloom filter?

ehsantn (approved):
Thanks @IsaacWarren. Do we know how much `probe_state` refactoring contributed to the performance improvement?

IsaacWarren:
> Thanks @IsaacWarren. Do we know how much `probe_state` refactoring contributed to the performance improvement?

Hoisting the pin was something like a ~3% improvement in probe time but it was never worse so no reason not to keep it IMO.

bodo/libs/streaming/_join.cpp
IsaacWarren: A skipped lookup only wastes one non-blocking fill and since it's non-blocking it doesn't stall the probe. Probe should be latency bounded, not bandwidth bounded since it's a pointer chase. Prefetches are also strictly scoped to the pinned batch's lookup loop, so there's no interaction with shuffle reception.

PR #1394, opened 2026-09-10
Simplify search with null cast in RexSimplifier inside AND

scott-routledge2:

Add functionality to RexSimplifier that looks for equivalent Search terms inside of an AND call, which have equivalent sargs and operands (except for a NOT NULL cast), and removes redundant terms. 

This fixes a plan regression from enabling predicates pullup.

For example:
`AND(SEARCH($0, Sarg['...'; NULL AS FALSE]:VARCHAR"), SEARCH(CAST($0):VARCHAR NOT NULL, Sarg['...']:VARCHAR))` 

Would become `SEARCH($0, Sarg['...'; NULL AS FALSE]:VARCHAR")`

## Testing strategy

Ran problematic test and compared diff to before predicates pullup PR to confirm join order change and redundant AND'ed Search arguments went away; Java CI

## Checklist

### DIFF
--- a/BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/org/apache/calcite/rex/RexSimplify.java
+++ b/BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/org/apache/calcite/rex/RexSimplify.java
@@ -1683,15 +1683,17 @@ public class RexSimplify {
     }
 
     /**
-     * Determine whether two search terms are equivalent
+     * Determine whether two search terms are equivalent.
      * Search terms are equivalent if their arguments are the same
-     * except for operand nullability is represented i.e.:
+     * except for operand nullability i.e.:
      *
      * <p>SEARCH(x, Sarg[..., NULL AS FALSE])</p>
      *
      * <p>versus</p>
      *
      * <p>SEARCH(CAST(x AS T NOT NULL), Sarg[...])</p>
+     *
+     * Assumes UNKNOWN and FALSE are equivalent in the outer expression.
      */
     private boolean equivalentSearchTerms(RexCall searchA, RexCall searchB) {
         assert ((searchA.getKind() == SqlKind.SEARCH) && (searchB.getKind() == SqlKind.SEARCH));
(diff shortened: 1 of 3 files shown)


BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/org/apache/calcite/rex/RexSimplify.java
```diff
@@ -1656,6 +1658,72 @@ RexNode simplifyAnd2ForUnknownAsFalse(List<RexNode> terms,
         return simplifyAnd2ForUnknownAsFalse(terms, notTerms, Comparable.class);
     }
 
+    /**
+     * Bodo Change: Add private methods for finding equivalent search terms
+     * inside conjunctions and removing redundant terms.
+     */
+    private void deduplicateSearchTerms(List<RexNode> terms) {
```
ehsantn, line 1680: This is a very narrow optimization. Let's add the exact example that motivated this in comments. I was wondering about the root cause of this happening if we know it (why one side has the non-null cast and the other doesn't).

ehsantn (approved):

BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/org/apache/calcite/rex/RexSimplify.java
IsaacWarren: I agree, I think there's a version of this that could be more general

IsaacWarren (approved):

BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/org/apache/calcite/rex/RexSimplify.java
scott-routledge2: The not null cast in search is coming from filter pushdown rules.

Before optimization, the plan looks like:
```
Filter(SEARCH(...)
  Join(...)
     Project(CAST($0 NOT NULL,...)
       SnowflakeTableScan(...)
     SnowflakeTableScan(...)
```

So one of the sides of the join has an additional `Project(CAST($0) NOT NULL)`, which when the filter containing search is pushed below this projection, it becomes `Filter(SEARCH(CAST(($0) NOT NULL)`.

The `Project(CAST($0 NOT NULL,...)` is coming from the fact that the table metadata in the snowflake catalog has this column as NOT NULL, but the view definition has that column as nullable. In BodoSQL, we explicitly insert casts when we encounter this kind of type mismatch:
https://github.com/bodo-ai/Bodo/blob/734809ab65903f9f8cae94372b86cf9ea26416c9/BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/prepare/AbstractPlannerImpl.java#L405

scott-routledge2: And also we end up with `AND(SEARCH(...), SEARCH(...)` because of the `JOIN_PUSH_TRANSITIVE_PREDICATES` rule, which moves predicates on one side of a join to the other side. 

An alternative fix might be at this level, avoid adding predicates that are equivalent, or removing the not null cast when adding the predicates initially.

PR #1398, opened 2026-09-14
Generate one-sided predicates from OR-of-AND clauses in Join conditions

ehsantn:

As title. Improves TPC-H Q7 and Q19 plans significantly by generating filters pushed down to I/O.

## Testing strategy

See https://github.com/bodo-ai/customer-sample-code/pull/165

## User facing changes

New optimization.

## Checklist


BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/prepare/BodoRules.kt
```diff
@@ -1321,6 +1329,11 @@ object BodoRules {
             PRUNE_EMPTY_UNION_RULE,
             PROJECT_VALUES_REDUCE_RULE,
             JOIN_REORDER_CONDITION_RULE,
+            // Avoiding repeated application of this rule here since it can lead to cycles. It
+            // requires extracting canonical forms of predicates from query subtrees to know if the derived predicate
+            // is already applied, but our infrastructure cannot provide this reliably. For example, "SEARCH(a, [1,2])"
+            // is not converted to "a=1 or a=2".
+//            JOIN_DERIVE_OR_PREDICATES_RULE,
```
DrTodd13, line 1336: Just a comment? Going to remove?

BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/application/logicalRules/JoinDeriveOrPredicatesRule.java
```diff
@@ -0,0 +1,248 @@
...
+    final RexSimplify simplify =
+        new BodoRexSimplify(rexBuilder, RelOptPredicateList.EMPTY, executor);
+
+    final int nLeft = join.getLeft().getRowType().getFieldCount();
+    final int nRight = join.getRight().getRowType().getFieldCount();
+    final ImmutableBitSet leftBits = ImmutableBitSet.range(0, nLeft);
+    final ImmutableBitSet rightBits = ImmutableBitSet.range(nLeft, nLeft + nRight);
+
+    // Predicates on the outer side of the join cannot be pushed down so don't derive them.
+    // TODO(Ehsan): avoid sides affected by any outer join in the query subtree similar to Postgres:
```
IsaacWarren, line 95: Why can we merge without addressing this?

BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/application/logicalRules/JoinDeriveOrPredicatesRule.java
```diff
@@ -0,0 +1,248 @@
...
+   *
+   * @param join The join node that may be rewritten.
+   * @return If the join is a candidate.
+   */
+  public static boolean containsOrCondition(Join join) {
+    RexNode cond = join.getCondition();
+    return filterContainsOr(cond)
+        && !RexUtil.SubQueryFinder.containsSubQuery(join)
+        && !RexOver.containsOver(cond)
+        && (join.getJoinType() != JoinRelType.FULL);
```
DrTodd13, line 76: When I was looking at constraints, it seemed to indicate only safe for inner joins.  So, could/should we be more restrictive?

IsaacWarren (approved):
Thanks Ehsan, after a quick look shouldn't this rule be responsible for this? JoinExtractFilterRule

DrTodd13 (approved):

BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/prepare/BodoRules.kt
ehsantn: I think it's important to know why this rule isn't present here while similar ones like `JOIN_REORDER_CONDITION_RULE` are here.

BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/application/logicalRules/JoinDeriveOrPredicatesRule.java
ehsantn: The added conjunctions are redundant and always safe. They just won't benefit performance if the join is outer of their side (the downside could be different cost estimates). Full outer join can't benefit so avoid the extra work.

scott-routledge2 (approved):
LGTM, thanks @ehsantn !

BodoSQL/calcite_sql/bodosql-calcite-application/src/main/java/com/bodosql/calcite/application/logicalRules/JoinDeriveOrPredicatesRule.java
ehsantn: Generating predicates is always safe, but there may be no benefit. In the cases specified here, there could be some benefit but there won't be filter pushdown into I/O so less benefit. Not a huge deal here.

ehsantn:
> Thanks Ehsan, after a quick look shouldn't this rule be responsible for this? JoinExtractFilterRule

`JoinExtractFilterRule` converts the join's condition to a filter on top of it with the hope of the filter being combined with other things there. It's unrelated to this case.

IsaacWarren:
> > Thanks Ehsan, after a quick look shouldn't this rule be responsible for this? JoinExtractFilterRule
> 
> `JoinExtractFilterRule` converts the join's condition to a filter on top of it with the hope of the filter being combined with other things there. It's unrelated to this case.

Oh so that filter doesn't get pushed below it? Interesting

ehsantn:
> > > Thanks Ehsan, after a quick look shouldn't this rule be responsible for this? JoinExtractFilterRule
> > 
> > 
> > `JoinExtractFilterRule` converts the join's condition to a filter on top of it with the hope of the filter being combined with other things there. It's unrelated to this case.
> 
> Oh so that filter doesn't get pushed below it? Interesting

Other rules do that.

PR #1408, opened 2026-09-18
Fix arith op expr with decimal.

DrTodd13:

* ArithOpExpr was previous getting the original value but if other side is a decimal then we convert the float side to decimal and now pass the converted value instead of the original. 
* Fixes bug with decimal division scale. 
* Send scale in shuffle. 

## Testing strategy

run_ci

## User facing changes

None

## Checklist


BodoSQL/bodosql/tests/test_decimal.py
```diff
@@ -3719,3 +3719,40 @@ def test_haversine_decimal(df, ans, memory_leak_check):
...
+
+    with temp_config_override("bodo_use_decimal", True):
+        check_query(
+            query,
+            ctx,
+            None,
+            expected_output=pd.DataFrame({"RES": [expected]}),
+            rtol=1e-04,
+            sort_output=False,
+            check_dtype=False,
```
IsaacWarren: Should we check dtype?

IsaacWarren (approved):

BodoSQL/bodosql/tests/test_decimal.py
scott-routledge2: I just copied the flags from some of the other tests. I think we can do check_dtype here though.

scott-routledge2 (approved):

